PR #2, opened 2016-03-16
Update dataloader.q

mortensorensen:
Do optional data processing (adding/dropping columns, etc. through `dataprocessfunc) before enumerating the table.


jonnypress:
Looks good, thanks Morten

PR #3, opened 2016-03-23
Fix .Q.fu issue when used within scaleSampling

WKent-Lee:
.Q.fu is coded wrongly to take `count each x` into the function `{calcsize[x;typesize x 0;$[1000<count x;0;attrsize each x]]}`, causing the typesize to be long type and return incorrect result.
.Q.fu is removed since scaleSampling already does the optimization. Doesn't make sense to have it in there.


WKent-Lee:
Function should work as expected now compare to the old one:
```
q)n:10000000
q)quote:([]time:`timestamp$abs n?0;sym:n?`4;bid:`float$n?0;ask:`float$n?0;bsize:`long$n?0;asize:`long$n?0;exch:n?.Q.a;exch1:n?.Q.a;str:{10?.Q.a}each til n)
q)objsize quote
1293078816
q)objsize2 quote
1293078816
q)\ts objsize quote
71 5506944
q)\ts objsize2 quote
1588 570426544
```

PR #10, opened 2016-10-25
Gateway retrydiscovery

aodhan1:
Feature #1747: Improve connections in TorQ Gateway


jonnypress:
Could you remove .servers.retrydiscovery[] from the sleep logic - that should already be called in .servers.startup[]

PR #11, opened 2016-10-26
Updated torq.q readproc[] to handle malformed port fields or unset en…

AllanMoore20:
…vironment variables in process.csv

### DIFF
--- a/torq.q
+++ b/torq.q
@@ -308,7 +308,7 @@ file:$[`procfile in key params;
 	first `$params `procfile;
  	first getconfigfile["process.csv"]];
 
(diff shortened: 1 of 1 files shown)


picoDoc:
Yeah it seems to work well.  Also seems to let you start a process with no port which is nice.

One request tho, can you break that function up a bit to make it more readable?  Give it some commenting like the functions around it to explain what it does, and get rid of that triply nested @ stuff by properly naming the functions according to what they do and calling them, rather than having everything in one mad line.  Let's be kind to our future selves :-)

picoDoc:
Looks good, going to merge

PR #12, opened 2016-10-27
Port closing fix

Hanso38:
Tried overnight, and it seems to have fixed the monitor bug. At the very least it is a sensible change to avoid the use of deprecated code.

### DIFF
--- a/code/common/html.q
+++ b/code/common/html.q
@@ -23,7 +23,7 @@ init:{
  modifier,::new!(count new)#{-8!.j.j updformat["upd";`tablename`tabledata!(x 1;x 2)]}}
 
 del:{w[x]_:w[x;;0]?y};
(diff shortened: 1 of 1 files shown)


code/common/html.q
```diff
@@ -23,7 +23,7 @@ init:{
  modifier,::new!(count new)#{-8!.j.j updformat["upd";`tablename`tabledata!(x 1;x 2)]}}
 
 del:{w[x]_:w[x;;0]?y};
-.z.pc:{.html.del[;y] each .html.t; x@y}@[value;`.z.pc;{{[x]}}]
+if[.z.K >= 3.3;.z.wc:{.html.del[;y] each .html.t; x@y}@[value;`.z.wc;{{[x]}}];.z.pc:{.html.del[;y] each .html.t; x@y}@[value;`.z.pc;{{[x]}}]]
```
picoDoc: Looks good.  Can you define this function separately (rather than writing out twice), and add a comment explaining what's going on, and then we're good!

Hanso38:
Alright, how's that now?

picoDoc:
Looks good, merging

PR #15, opened 2016-11-03
Adding tickerplant to TorQ, TP improvements

jonathonmcmurray:
Add TP to TorQ
Dictionary of per-table counts in TP, CTP for quicker recovery
Explicit TP parameters for schema file and log directory
Better timezone handling in TP


jonnypress:
Looks good, merging

jonnypress:
Looks good, merging

PR #17, opened 2016-11-21
Moved sort slave config into TorQ

BenMcElderryAquaQ:


jonnypress:
Could you see if you can trim down that config a bit?  I think a lot of those values aren't applicable to the sort slave process as it just gets invoked by the main sort process. I think maybe it only needs the .servers stuff, but not sure.  Can you test with both the sort and merge approach?

PR #22, opened 2016-12-05
Fixed pointer size in mem.objsize for 32 bit kdb

AllanMoore20:


jonnypress:
Could you make the pointer size a global variable the same way that version is?

PR #23, opened 2017-01-05
Improved debugging features 

aidanogorman:
Improved debugging features in the loadf function so that errors inside files are no longer masked. If an error occurs when files are loaded in, it will flag at that point in the file and enter debug mode.

I'm happy in my eyes that this change won't have any other implications and there are no other error traps that will cause issue.


picoDoc:
looks good to me

PR #40, opened 2017-03-07
Added error trap for wdb writedown mode

aquaq:


jonnypress:
removing, was resolved another way

PR #42, opened 2017-03-08
Fixed issue where eod would fail when wdb was set to partbyattr and t…

AllanMoore20:
…here was no data in a table

Had to make a new branch to fix conflicts with Scott's recent change.


AllanMoore20:
I have combined the variables into a dictionary, I think that's what you wanted, should the tablist variable have a different name now that it' a dictionary, or is it fine the way it is?

PR #43, opened 2017-03-13
Added check when part by attr set, if directory exists delete it

claire-sloan:
Getting error in log - failure to remove directory
When there is no data flowing in, at eod empty tables will be saved down but as there was no data flowing in, no partition has been created thus added check, if the path exists delete it.

### DIFF
--- a/code/processes/wdb.q
+++ b/code/processes/wdb.q
@@ -300,8 +300,8 @@ endofdaymerge:{[dir;pt;tablist;mergelimits]
 		merge[dir;pt;;mergelimits] peach flip (key tablist;value tablist);];	
 		[.lg.o[`merge;"merging on master"];
 		merge[dir;pt;;mergelimits] each flip (key tablist;value tablist)]];
-	/- if directory exists delete it
-        if[0 < count key p:.Q.par[savedir;pt;`]; .os.deldir .os.pth[string p]];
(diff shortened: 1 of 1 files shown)


code/processes/wdb.q
```diff
@@ -300,8 +300,8 @@ endofdaymerge:{[dir;pt;tablist;mergelimits]
 		merge[dir;pt;;mergelimits] peach flip (key tablist;value tablist);];	
 		[.lg.o[`merge;"merging on master"];
 		merge[dir;pt;;mergelimits] each flip (key tablist;value tablist)]];
-	/- delete the empty date directory
-	.os.deldir .os.pth[string .Q.par[savedir;pt;`]];	
+	/- if directory exists delete it
+        $[`error ~ @[get; (hsym `$.os.pth[string .Q.par[savedir;pt;`]]); {`error}]; (); .os.deldir .os.pth[string .Q.par[savedir;pt;`]]];
```
jonnypress: You can derive the path just once rather than twice (i.e. this bit: .os.pth[string .Q.par[savedir;pt;`]])

To check existence it is better to use key

key on a directory that exists returns a symbol list of it's contents
key on a file that exists returns the name of the file as a symbol
key on a file/directory that doesn't exist returns empty list ()

PR #45, opened 2017-03-15
addprocscustom is now defined with an errortrap

mmeenagh1:


code/handlers/trackservers.q
```diff
@@ -235,6 +235,8 @@ addprocs:{[connectiontab;procs;connect]
 
 // addprocscustom is to allow bespoke extensions when adding processes
 addprocscustom:{[connectiontab;procs]}
```
jonnypress: surely you need to delete this line?

mmeenagh1: The error trap should work with this.  (it just gave a projection before)
addprocscustom:{$[x=0b;{[connectiontab;procs]};x]}[@[value;`.servers.addprocscustom;0b]]
is there anything more I need to do on it ?

mmeenagh1: Just a quick explanation of why I changed it:

If I set
addprocscustom:@[value;`.servers.addprocscustom;{[connectiontab;procs]}]
It gives a rank error, for example if a variable called var is not defined the result of the error trap is a projection:
@[value;`var;{[x1;x2]}] 
q) {[x1;x2]}["var"]

jonnypress: If you want an error trap to return a function then you have to wrap the function within the error trap function e.g. 

@[value;.servers.addprocscustom;{{[connectiontab;procs]}}]

mmeenagh1: Oh I see. I'll make that change now

PR #46, opened 2017-03-20
compression bug fix

mmeenagh1:


code/processes/wdb.q
```diff
@@ -232,14 +232,26 @@ doreload:{[pt]
...
+// set .z.zd to control how data gets compressed
+setcompression:{[compression] if[ 3= count compression;
+							.lg.o[`compression;"setting compression level to (",(";" sv string compression),")"];
+                        				.z.zd:compression;
+                        				.lg.o[`compression;".z.zd has been set to (",(";" sv string .z.zd),")"]]}
+//reset .z.zd to 16 0 0 (i.e no compression)
+removecompression:{[compression] if[ 3= count compression;
+					    		.z.zd:16 0 0;
+                                            		.lg.o[`compression;"resetting compression level to (",(";" sv string 16 0 0),")"]]}
+
```
jonnypress: Could you make "set compression" and removecompression a single function?  (or remove could be a projection of set).  So basically if count=3, then set it.  if count = 0 (empty list), then unset it

code/processes/wdb.q
```diff
@@ -232,14 +232,26 @@ doreload:{[pt]
...
 endofdaysortdate:{[dir;pt;tablist]
 	/-sort permitted tables in database
 	/- sort the table and garbage collect (if enabled)
 	.lg.o[`sort;"starting to sort data"];
 	
 	$[(0 < count .z.pd[]) and ((system "s")<0);
 		[.lg.o[`sort;"sorting on slave sort", string .z.p];
-		{[x] .sort.sorttab[x];if[gc;.gc.run[]]} peach tablist,'.Q.par[dir;pt;] each tablist;
+		{[x] setcompression[compression];.sort.sorttab[x];removecompression[compression];if[gc;.gc.run[]]} peach tablist,'.Q.par[dir;pt;] each tablist;
 		];
```
jonnypress: This isn't going to work.  The sorting here is done on a remote process, which might have different compression parameters defined.  You should pass the compression settings to the remote process (as an extra parameter to the function)

code/processes/wdb.q
```diff
@@ -296,8 +308,10 @@ merge:{[dir;pt;tableinfo;mergelimits]
 endofdaymerge:{[dir;pt;tablist;mergelimits]		
 	/- merge data from partitons
 	$[(0 < count .z.pd[]) and ((system "s")<0);
-		[.lg.o[`merge;"merging on slave"];
+		[.lg.o[`merge;"merging on slave"];		
+		setcompression[compression];
 		merge[dir;pt;;mergelimits] peach flip (key tablist;value tablist);];	
+		removecompression[compression];
 		[.lg.o[`merge;"merging on master"];
 		merge[dir;pt;;mergelimits] each flip (key tablist;value tablist)]];
```
jonnypress: I think you need to wrap this block in set/remove statements?  This is the bit that does the merging on the master process

code/processes/wdb.q
```diff
@@ -296,8 +308,10 @@ merge:{[dir;pt;tableinfo;mergelimits]
 endofdaymerge:{[dir;pt;tablist;mergelimits]		
 	/- merge data from partitons
 	$[(0 < count .z.pd[]) and ((system "s")<0);
-		[.lg.o[`merge;"merging on slave"];
+		[.lg.o[`merge;"merging on slave"];		
+		setcompression[compression];
 		merge[dir;pt;;mergelimits] peach flip (key tablist;value tablist);];	
+		removecompression[compression];
```
jonnypress: I don't think this will work. setcompression will be executed locally, whereas merge will be executed remotely.  You need to wrap the merge statement in a lambda and set/remove compression either side of the merge function call (the same as you do with the other peach statement)

code/processes/wdb.q
mmeenagh1: Oh ok, I had been setting values for "compression" in the sort/sortslaves config files. Instead, should the "compression" value set in the wdb config file get used in all 3 cases ?

jonnypress: Yes - it should just be set in one place and it should drive the other processes.

mmeenagh1: I've fixed that problem now, but I've had to add a compression parameter into a few functions (informsortandreload, endofdaysort, endofdaymerge, endofdaysortdate) so that it feeds through to the sort/sortslave processes. Is this ok or is there a better way to do it ?

code/processes/wdb.q
```diff
@@ -232,14 +232,18 @@ doreload:{[pt]
 	];
 	};
 
-endofdaysortdate:{[dir;pt;tablist]
+// set .z.zd to control how data gets compressed
+setcompression:{[compression] if[3=count compression;.lg.o[`compression;$[compression~16 0 0;"resetting";"setting"]," compression level to (",(";" sv string compression),")"];.z.zd:compression]}
```
jonnypress: could you split this across lines? 

you could also add a function of 
resetcompression:{setcompression 16 0 0}

which might be nice

PR #47, opened 2017-03-30
Add colouring to log output

jonathonmcmurray:


torq.q
```diff
@@ -142,8 +142,13 @@ getapplication:{$[0 = count a:@[{read0 x};hsym last getconfigfile"application.tx
...
 // Logging functions live in here
 
+colours:(`ERROR`ERR`WRN`WARN`INF!("\033[1;31m";"\033[1;31m";"\033[0;33m";"\033[0;33m";"\033[0m"));
 // Format a log message
-format:{[loglevel;proctype;proc;id;message] "|" sv (string .proc.cp[];string .z.h;string proctype;string proc;string loglevel;string id;message)}
+if[.z.o like "w*";
+ format:{[loglevel;proctype;proc;id;message] "|" sv (string .proc.cp[];string .z.h;string proctype;string proc;string loglevel;string id;message)}];
+
+if[not .z.o like "w*";
+ format:{[loglevel;proctype;proc;id;message]((colours loglevel), "|" sv (string .proc.cp[];string .z.h;string proctype;string proc;string loglevel;string id;message)),"\033[0m"}];
```
jonnypress: can you just make this an if-else ?

picoDoc: Jonathon's obviously a disciple of Rolf's style guide now! :-P

picoDoc:
Can I raise a general point about this: what happens if logs are not being viewed in putty?  how do other programs handle the \033 stuff?  what happens in vim, or windows based apps, do we just get \033 stuff everywhere?

torq.q
ThomasSmyth: Alternatively you could define a vanilla format function and then just wrap it if not on windows:
`format:{[func;loglevel;proctype;proc;id;message] colours[loglevel],func[loglevel;proctype;proc;id;message],"\033[0m"}[format]`

Also, what if someone doesn't want colour? Can there be some option to turn it off even if its in `torq.q`?

ThomasSmyth:
I tested a few logging commands and can see a load of special characters when I use vim on the log files.

picoDoc:
Instead of just poking holes here's an actual suggestion/request: can you find another unix app that colours it's logs like this?  If so fair enough, lets do it.  If not you gotta ask why no one else does this (personally I've never seen it before)

jonathonmcmurray:
In regards to special chars in vim, surely "less" is a more suitable tool for reading log files? with the -r option, less displays the colours correctly

I think you can maybe also get a vim plugin to show them properly there, there are possibly similar plugins for Notepad++ etc. on the Windows side, but I guess all this isn't necessarily very user-friendly.

Another option is making the colours only display when running interactively, seeing as the log files are already separated by type (err, out etc.), so colours in the files probably aren't as useful?

jonnypress:
Closing this - seems there would be other ways to do it if required.

PR #49, opened 2017-04-13
kafka for l64

mmeenagh1:


code/common/kafka.q
```diff
@@ -0,0 +1,52 @@
+\d .kafka
+
+// configuration
+enabled:@[value;`enabled;.z.o in `l32`l64]		// whether kafka is enabled
+
```
jonnypress: we should have a default definition here of kupd e.g. 

kupd:@[value;`kupd;{{[k;x] -1 `char$x;}}]

to allow this function to be set externally

code/common/kafka.q
```diff
@@ -0,0 +1,52 @@
...
+	/ publish[topic;partition;key;message]
+	/ publish 'message' byte vector to topic, partition. symbol key can be null
+	/ e.g. publish[`test;0;`;`byte$"hello world"]
+	publish: lib 2: (`publish;4);
+  ];
+ ];
+
+\d .
+if[.kafka.enabled;
```
jonnypress: Can you move this definition into the previous block (in the .kafka namespace) and use it to set kupd from the default value.  Note also that kupd should only be set if it is not already set.  I think it should look something like this: 

if[not `kupd in key`.; @[`.;`kupd;:;.kafka.kupd]]; 
.lg.o[`kafka;"kupd is set to ",-3!kupd]

config/settings/default.q
```diff
@@ -110,6 +110,13 @@ usessl:0b                              		// connect using SSL/TLS
 debug:0i                               		// debug level for email library: 0i = none, 1i=normal, 2i=verbose
 img:`$getenv[`KDBHTML],"/img/AquaQ-TorQ-symbol-small.png"	// default image for bottom of email
 
+
+// configuration for kafka
+\d .kafka
+enabled:1b                                      // whether kafka is enabled
+
```
jonnypress: add a kupd definition here

PR #62, opened 2017-12-06
Sean h utils

shehir94:
rack and interval code, along with rack doc and tests for rack


shehir94:
Hi Jonny, the updated general utilities doc is up now. The finished code is on Seans branch

PR #63, opened 2018-01-08
updated loadallfiles function to filter based on filext

michaeld1015:


michaeld1015:
It'd be nice to not have to check `0h<type loadparams[filext]` but `like` is funny with atoms and lists... I couldn't get it to work any other way basically

torq.q
```diff
@@ -426,17 +426,17 @@ createalias:{[logdir;filename;alias]
...
+	fileredirect[logdir;"out_",basename;"out_",alias;1];]]
 	.lg.banner[]}
 
 // function to produce the timestamp value for the log file
 logtimestamp:@[value;`logtimestamp;{[x] {[]`$ssr[;;"_"]/[string .z.z;".:T"]}}]
 
 rolllogauto:{[] 
 	.lg.o[`logging;"creating standard out and standard err logs"];
-	createlog[getenv`KDBLOG;procname;logtimestamp[];`suppressalias in key params]}
+	createlog[getenv`KDBLOG;procname;logtimestamp[];`suppressalias in key params;`onelog in key params]}
```
jonnypress: Will this work? looks like you are passing one param too many?

jonathonmcmurray:
@michaeld1015 instead of `0h<type loadparams[filext]` and separate approaches for list or atom, could you use `(),loadparams[filext]` to ensure it is a list, and just use the list method?

torq.q
```diff
@@ -426,17 +426,17 @@ createalias:{[logdir;filename;alias]
 // makealias = if true, will create alias files without the timestamp value
 createlog:{[logdir;logname;timestamp;suppressalias]
 	basename:(string logname),"_",(string timestamp),".log";
-	alias:$[suppressalias;"";(string logname),".log"];
+	alias:$[suppressalias;"";(string logname),".log"];$[onelog;fileredirect[logdir;"err_out_",basename;"err_out_",alias;2];[
```
jonnypress: could you stick a line break in here.  
if it's just a single log file could it not have a prefix - just the basename will do. 
Are you redirecting 1 correctly if there is a single log file?

michaeld1015:
tried that doesn't seem to work, if someone specifies more than one pattern then `type loadparams[filepattern]` is `10h` whereas one pattern type is `0h` even with `(),loadparams...`

jonnypress:
yes that's because (), on a string yields a string.  You could just do

michaeld1015:
Think I come up with a nice way to handle the filepattern problem, not overly sure about the onelog change I've made and if it'll work. I see there's another createlog function in logusage.q not sure if I need to change that as well. I'll try and take another look at this again...

torq.q
```diff
@@ -184,7 +184,7 @@ publish:{[loglevel;proctype;proc;id;message]
 
 // Dictionary of log levels mapped to standard out/err
 // Set to 0 if you don't want the log type to print
-outmap:@[value;`outmap;`ERROR`ERR`INF`WARN!2 2 1 1]
+outmap:@[{[x]value[x]&1};`outmap;`ERROR`ERR`INF`WARN!2 2 1 1]
```
ThomasSmyth: Can be `{1&value x}` or possibly just `1&value`

torq.q
```diff
@@ -424,19 +424,21 @@ createalias:{[logdir;filename;alias]
 // logname = base of log file
 // timestamp = optional timestamp value (e.g. .z.d, .z.p)
 // makealias = if true, will create alias files without the timestamp value
-createlog:{[logdir;logname;timestamp;suppressalias]
+createlog:{[logdir;logname;timestamp;suppressalias;onelog]
+        out_prefix:"out_";err_prefix:"err_";
 	basename:(string logname),"_",(string timestamp),".log";
```
ThomasSmyth: If you're editing you may as well move this to square brackets `string[...]`

torq.q
```diff
@@ -424,19 +424,21 @@ createalias:{[logdir;filename;alias]
...
-	alias:$[suppressalias;"";(string logname),".log"];
-	fileredirect[logdir;"err_",basename;"err_",alias;2];
-	fileredirect[logdir;"out_",basename;"out_",alias;1];
+        alias:$[suppressalias;"";(string logname),".log"];
```
ThomasSmyth: Use square brackets for `string[...]`. Also this line looks misaligned.

torq.q
```diff
@@ -424,19 +424,21 @@ createalias:{[logdir;filename;alias]
...
+	fileredirect[logdir;out_prefix,basename;out_prefix,alias;1];
+        fileredirect[logdir;err_prefix,basename;err_prefix,alias;2];
+        if[onelog;.lg.outmap[key .lg.outmap]:1];
```
ThomasSmyth: This and the line above look misaligned.

torq.q
```diff
@@ -424,19 +427,21 @@ createalias:{[logdir;filename;alias]
...
-	alias:$[suppressalias;"";(string logname),".log"];
-	fileredirect[logdir;"err_",basename;"err_",alias;2];
-	fileredirect[logdir;"out_",basename;"out_",alias;1];
+createlog:{[logdir;logname;timestamp;suppressalias;onelog]
+        out_prefix:"out_";
+	err_prefix:$[onelog;out_prefix;"err_"];
+	basename:string[logname],"_",string[timestamp],".log";
+        alias:$[suppressalias;"";string[logname],".log"];
+	fileredirect[logdir;out_prefix,basename;out_prefix,alias;1];
+        fileredirect[logdir;err_prefix,basename;err_prefix,alias;2];
```
jonnypress: CAn you fix up the indentations here?

michaeld1015: I can't see the indentations on vim. take it that's normal?

docs/utilities.md
```diff
@@ -570,6 +570,7 @@ dictionary should/can have the following fields:
 |    chunksize    |  N   |     int      | Data size in bytes to read in one chunk. Default is 100 MB |
 |   compression   |  N   |   int list   | Compression parameters to use e.g. 17 2 6. Default is empty list for no compression |
 |       gc        |  N   |   boolean    | Whether to run garbage collection at appropriate points. Default is 0b (false) |
+|  filepattern    |  N   | char\[list\] | Pattern used to only load certain files e.g. "*.csv"|
```
jonnypress: This has to be a char list or a list of char lists doesn't it?  i.e. a string or a list of strings.  Can you give two examples- one with a single string and one with multiple strings.

PR #68, opened 2018-01-31
added function to allow deletion or moving of duplicate files in file…

denismc88:
…alerter process.

Just took existing removedups function from housekeeping process and added an option to move files also if variable duplicateaction is set to move and the filealertercsv has a movetodirectory defined. Otherwise works as before if variable unset or set to anything else.


code/processes/filealerter.q
```diff
@@ -93,19 +93,35 @@ chktable:{[files] table:([]filename:files;md5hash:$[usemd5;gethash'[files];(coun
...
-        pf:action/:[DICT[`function];files];];
-        .lg.o[`alerter;"no new files found"]];
-        t:update function:(count toprocess)#DICT[`function],funcpassed:pf,moveto:(count toprocess)#enlist .rmvr.removeenvvar[DICT[`movetodirectory]] from toprocess; t}
-
-
+    /-find all matches to the file search string
+    matches:find[.rmvr.removeenvvar[DICT[`path]];DICT[`match]];
+    /-process duplicate files, either remove, move or leave depending on duplicateaction variable
+    processdups[;DICT] each matches;
+    toprocess:getunprocessed[matches;DICT[`newonly]];
```
jonnypress: I think I would do this the other way around.  So the duplicates are basically

matches except toprocess

Otherwise we will end up with duplicate detection logic in two places.

code/processes/filealerter.q
```diff
@@ -93,19 +93,35 @@ chktable:{[files] table:([]filename:files;md5hash:$[usemd5;gethash'[files];(coun
...
+    /-perform the function on the file
+    pf:action/:[DICT[`function];files];];
+    .lg.o[`alerter;"no new files found"]];
+    t:update function:(count toprocess)#DICT[`function],funcpassed:pf,moveto:(count toprocess)#enlist .rmvr.removeenvvar[DICT[`movetodirectory]] from toprocess; t}
+
+//-process duplicates (comparison to filealerterprocessed)
+duplicateaction:`delete;
+
+processdups:{
+    $[duplicateaction=`delete;
```
jonnypress: I think we can make this a bit more generic.  We can define a function, processduplicate, or something, and processdups just invokes processduplicate on each duplicate file.  

By default processduplicate will just be a null op e.g. 

processduplicate:@[value;`.fa.processduplicate;{{[x]}}]

And then we can override it in the config file with a delete/hdel operation

PR #69, opened 2018-01-31
onelog flag documentation

michaeld1015:


docs/gettingstarted.md
```diff
@@ -82,7 +82,7 @@ The available command line parameters are:
   |-nopi                     |Reset the definition of .z.pi to the initial value (useful for debugging)|
   |-debug                    |Equivalent to \[-nopi -noredirect\]|
   |-usage                    |Print usage info and exit|
-
+  |-onelog                   |Writes all messages to stdout log file, note non-trapped errors will still be written to stderr log file|
```
jonnypress: Can you add a bit more detail about this to the logging section: http://aquaqanalytics.github.io/TorQ/gettingstarted/#logging

Explain the short coming (i.e. TorQ uses \1 and \2 for redirects, but that doesn't allow both to be redirected to the same file) so any errors that are raised by kdb+ and unhandled will end up in the error log file.

docs/gettingstarted.md
```diff
@@ -195,6 +195,12 @@ The date suffix can be overridden by modifying the .proc.logtimestamp
 function and sourcing torq.q from another script. This could, for
 example, change the suffixing to a full timestamp.
 
+In the case where -onelog is flagged TorQ will attempt to redirect
+all output to the out log file, unfortunately this is not perfect.
+TorQ uses \1 and \2 to redirect stderr and stdout, onelog only
+overrides handled errors to the \1 redirect. Any errors that are
```
jonnypress: put in here "because \1 and \2 cannot be redirected to the same file"

jonathonmcmurray: Am I missing something Jonny? Seems like they can be redirected to the same file?

```jmcmurray@homer ~/scrap $ q
KDB+ 3.5 2017.08.22 Copyright (C) 1993-2017 Kx Systems
l64/ 8()core 16048MB jmcmurray homer.aquaq.co.uk 127.0.1.1 EXPIRE 2018.06.30 AquaQ #50170

q)\1 t1.txt
q)\2 t1.txt
1+1
2+2
`a+3
\\
jmcmurray@homer ~/scrap $ cat t1.txt
2
4
q)'type
  [0]  `a+3
         ^
q)jmcmurray@homer ~/scrap $
```

jonnypress: Maybe... Couldn't do that on the version Michael was working on!  We'll try on later version.  Thanks

jonnypress: Actually... no.  The problem is that the order in the file gets mixed up - 1 prints to top, 2 prints after.  So the documentation comment should reflect that.  When this was being developed it was tested by tailing the log file which never showed the -1 prints as they were going to the top

```
q)\1 t1.txt
q)\2 t1.txt
1+1
`a+2
1+3

jonnyp@unix:~$ cat t1.txt 
2
4  <---- this should be after the error
'type
  [0]  `a+2
         ^
q)q)jonnyp@unix:
```

PR #78, opened 2018-03-06
Porting Torq to kdb 3.6

Abhairien:
Since 3.6, conditionals inside a function must have an odd number of arguments, otherwise it throws a 'cond error.

### DIFF
--- a/code/common/compress.q
+++ b/code/common/compress.q
@@ -173,7 +173,7 @@ compress:{[filetoCompress;algo;blocksize;level;sizeuncomp]
                 $[not 0=algo;statstab ,: (filetoCompress;algo;(-21!filetoCompress)`compressedLength;sizeuncomp);statstab ,: (filetoCompress;algo;comprL;sizeuncomp)]];
             [$[(not count (-21!compressedFile));
 		[.lg.o[`compression; "Failed to compress file ",(string filetoCompress)];hdel compressedFile];
-		.lg.o[`compression; $[algo=0;"Decompressed ";"Compressed "], "file ", (string compressedFile), " doesn't match original. Deleting new file"]; hdel compressedFile;]]]
+		.lg.o[`compression;$[algo=0;"Decompressed ";"Compressed "],"file ",string[compressedFile]," doesn't match original. Deleting new file"];hdel compressedFile;]]]
         ];
         / if already compressed/decompressed, then log that and skip.
         [((not 0 = count -21!filetoCompress) & not 0 = algo)|((0 = count -21!filetoCompress) & 0 = algo)];
--- a/code/common/compress.q
+++ b/code/common/compress.q
@@ -170,9 +170,9 @@ compress:{[filetoCompress;algo;blocksize;level;sizeuncomp]
                 / move the hash files too.
                 if[78 <= type sf; system "r ", (last ":" vs string compressedFile),"# ", (last ":" vs string filetoCompress),"#"];
                 /-log to the table if the algo wasn't 0
-                $[not 0=algo;statstab ,: (filetoCompress;algo;(-21!filetoCompress)`compressedLength;sizeuncomp);statstab ,: (filetoCompress;algo;comprL;sizeuncomp)]];
+                statstab,:$[not 0=algo;(filetoCompress;algo;(-21!filetoCompress)`compressedLength;sizeuncomp);(filetoCompress;algo;comprL;sizeuncomp)]];
             [$[(not count (-21!compressedFile));
-		[.lg.o[`compression; "Failed to compress file ",(string filetoCompress)];hdel compressedFile];
+		[.lg.o[`compression; "Failed to compress file ",string[filetoCompress]];hdel compressedFile];
 		.lg.o[`compression;$[algo=0;"Decompressed ";"Compressed "],"file ",string[compressedFile]," doesn't match original. Deleting new file"];hdel compressedFile;]]]
         ];
         / if already compressed/decompressed, then log that and skip.
--- a/code/common/compress.q
+++ b/code/common/compress.q
@@ -173,8 +173,8 @@ compress:{[filetoCompress;algo;blocksize;level;sizeuncomp]
                 $[not 0=algo;statstab ,: (filetoCompress;algo;(-21!filetoCompress)`compressedLength;sizeuncomp);statstab ,: (filetoCompress;algo;comprL;sizeuncomp)]];
             [$[(not count (-21!compressedFile));
 		[.lg.o[`compression; "Failed to compress file ",(string filetoCompress)];hdel compressedFile];
-		.lg.o[`compression; $[algo=0;"Decompressed ";"Compressed "], "file ", (string compressedFile), " doesn't match original. Deleting new file"]; hdel compressedFile]]]
+		.lg.o[`compression; $[algo=0;"Decompressed ";"Compressed "], "file ", (string compressedFile), " doesn't match original. Deleting new file"]; hdel compressedFile;]]]
         ];
         / if already compressed/decompressed, then log that and skip.
         [((not 0 = count -21!filetoCompress) & not 0 = algo)|((0 = count -21!filetoCompress) & 0 = algo)];
-        .lg.o[`compression; "file ", (string filetoCompress), " is already ",$[0=algo; "decompressed";"compressed"],". Skipping this file"]]}
+        .lg.o[`compression; "file ", (string filetoCompress), " is already ",$[0=algo; "decompressed";"compressed"],". Skipping this file"];]}
--- a/code/common/compress.q
+++ b/code/common/compress.q
@@ -158,14 +158,15 @@ summarystats:{
 compress:{[filetoCompress;algo;blocksize;level;sizeuncomp]
     compressedFile: hsym `$(string filetoCompress),"_kdbtempzip";
     / compress or decompress as appropriate:
+    cmp:$[algo=0;"de";""];
     $[((0 = count -21!filetoCompress) & not 0 = algo)|((not 0 = count -21!filetoCompress) & 0 = algo);
-        [.lg.o[`compression;$[algo=0;"Decompressing ";"Compressing "],"file ", (string filetoCompress), " with algo: ", (string algo), ", blocksize: ", (string blocksize), ", and level: ", (string level), "."];
+        [.lg.o[`compression;cmp,"compressing ","file ", (string filetoCompress), " with algo: ", (string algo), ", blocksize: ", (string blocksize), ", and level: ", (string level), "."];
          / perform the compression/decompression
         if[0=algo;comprL:(-21!filetoCompress)`compressedLength];
         -19!(filetoCompress;compressedFile;blocksize;algo;level);
          / check the compressed/decomp file and move if appropriate; else delete compressed file and log error
         $[((get compressedFile)~sf:get filetoCompress) & (count -21!compressedFile) or algo=0;
-            [.lg.o[`compression;"File ", $[algo=0;"decompressed ";"compressed "],"successfully; matches orginal. Deleting original."];
+            [.lg.o[`compression;"File ",cmp,"compressed ","successfully; matches orginal. Deleting original."];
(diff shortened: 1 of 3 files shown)


code/common/compress.q
```diff
@@ -173,8 +173,8 @@ compress:{[filetoCompress;algo;blocksize;level;sizeuncomp]
                 $[not 0=algo;statstab ,: (filetoCompress;algo;(-21!filetoCompress)`compressedLength;sizeuncomp);statstab ,: (filetoCompress;algo;comprL;sizeuncomp)]];
             [$[(not count (-21!compressedFile));
 		[.lg.o[`compression; "Failed to compress file ",(string filetoCompress)];hdel compressedFile];
-		.lg.o[`compression; $[algo=0;"Decompressed ";"Compressed "], "file ", (string compressedFile), " doesn't match original. Deleting new file"]; hdel compressedFile]]]
+		.lg.o[`compression; $[algo=0;"Decompressed ";"Compressed "], "file ", (string compressedFile), " doesn't match original. Deleting new file"]; hdel compressedFile;]]]
```
ThomasSmyth: Use square brackets for `string`. If you were feeling particularly cheeky you could do `$[algo=0;"Dec";"C"]ompressed`

code/common/compress.q
```diff
@@ -173,8 +173,8 @@ compress:{[filetoCompress;algo;blocksize;level;sizeuncomp]
...
         ];
         / if already compressed/decompressed, then log that and skip.
         [((not 0 = count -21!filetoCompress) & not 0 = algo)|((0 = count -21!filetoCompress) & 0 = algo)];
-        .lg.o[`compression; "file ", (string filetoCompress), " is already ",$[0=algo; "decompressed";"compressed"],". Skipping this file"]]}
+        .lg.o[`compression; "file ", (string filetoCompress), " is already ",$[0=algo; "decompressed";"compressed"],". Skipping this file"];]}
```
ThomasSmyth: Squares for `string` + same alternative as line 176.

code/common/html.q
```diff
@@ -67,7 +67,7 @@ jsformat:{ k:cols x; flip k !(y value t:type each x)@'value x:flip 0!x}
 // Arg: dictionary decoded front end JSON  
 // format should be `func`arg1`arg2 ... `arg8!(function;arg1;arg2;...;arg3)
 // all args except arg1 are optional
-execdict:{$[not `func in key x;'"no func in dictionary";1=count key x;(value x`func) @ 1;1<count key x;(value x`func) . value x _ `func]}
+execdict:{$[not `func in key x;'"no func in dictionary";1=count key x;(value x`func) @ 1;1<count key x;(value x`func) . value x _ `func;]}
```
ThomasSmyth: ``value[x`func]`` should work. You can probably remove `@` here too and index without it.

code/common/compress.q
jonathonmcmurray: Even better, you could do something that'll work e.g. `$[algo=0;"Dec";"C"],"ompressed"`

code/common/html.q
ThomasSmyth: What might be even better is have:
```
if[not`func in key x;...];
:value[x`func].$[1=count key x;(),1;value x _`func];
```
Untested but I think it should work.

code/common/compress.q
jzabroski: @ThomasSmyth  I'm still a q newbie. Can I ask why you would want to write it that way? Is there some efficiency to be gained besides optimizing to reduce character count?

ThomasSmyth: Q programmers often strive for terseness in their code and this case is no different. My suggestion does not improve the efficiency, but will avoid the unnecessary duplication of strings.

I would argue that by making code more terse it can be made neater and cleaner. This should benefit those who have to read it later and can help simplify later maintenance of the file.

ThomasSmyth:
Please update this PR with a relevant title and description, this will help others understand the purpose of your changes at a glance.

code/common/compress.q
```diff
@@ -173,8 +173,8 @@ compress:{[filetoCompress;algo;blocksize;level;sizeuncomp]
                 $[not 0=algo;statstab ,: (filetoCompress;algo;(-21!filetoCompress)`compressedLength;sizeuncomp);statstab ,: (filetoCompress;algo;comprL;sizeuncomp)]];
             [$[(not count (-21!compressedFile));
 		[.lg.o[`compression; "Failed to compress file ",(string filetoCompress)];hdel compressedFile];
-		.lg.o[`compression; $[algo=0;"Decompressed ";"Compressed "], "file ", (string compressedFile), " doesn't match original. Deleting new file"]; hdel compressedFile]]]
+		.lg.o[`compression;$[algo=0;"Decompressed ";"Compressed "],"file ",string[compressedFile]," doesn't match original. Deleting new file"];hdel compressedFile;]]]
         ];
         / if already compressed/decompressed, then log that and skip.
         [((not 0 = count -21!filetoCompress) & not 0 = algo)|((0 = count -21!filetoCompress) & 0 = algo)];
```
ThomasSmyth: Although you haven't made any changes here it may be an idea to clean up this function while you're in here. Both conditions here are repeatedly calculated when they could just be calculated once at the beginning of the function then referred to.

code/common/compress.q
```diff
@@ -173,8 +173,8 @@ compress:{[filetoCompress;algo;blocksize;level;sizeuncomp]
                 $[not 0=algo;statstab ,: (filetoCompress;algo;(-21!filetoCompress)`compressedLength;sizeuncomp);statstab ,: (filetoCompress;algo;comprL;sizeuncomp)]];
```
ThomasSmyth: Pull `statstab,:` out in front of the conditional then it only needs written once. The lists in each condition are also repetitive, could be room for further trimming.

ThomasSmyth:
Great point, with the compress/decompress behaviour being set at the start it makes sense to pass it through! Apologies for not spotting this first time round.

code/processes/wdb.q
```diff
@@ -335,7 +335,7 @@ reloadproc:{[h;d;ptype]
 	.wdb.countreload:count[raze .servers.getservers[`proctype;;()!();1b;0b]each reloadorder];
 	$[eodwaittime>0;
 		{[x;y;ptype].[{neg[y]@x};(x;y);{[ptype;x].lg.e[`reloadproc;"failed to reload the ",string[ptype]];'x}[ptype]]}[({@[`. `reload;x;()]; (neg .z.w)(`.wdb.handler;1b); (neg .z.w)[]};d);h;ptype];
-		@[h;(`reload;d);{[ptype;e] .lg.e[`reloadproc;"failed to reload the ",string[ptype],".  The error was : ",e][ptype]}];
+		@[h;(`reload;d);{[ptype;e] .lg.e[`reloadproc;"failed to reload the ",string[ptype],".  The error was : ",e][ptype]}]
```
jonnypress: I'm pretty sure that the error trap in this else case is incorrect (not your code change, but was originally incorrect).  I think it should be 

@[h;(`reload;d);{[ptype;e] .lg.e[`reloadproc;"failed to reload the ",string[ptype],".  The error was : ",e]}[ptype]]

i.e. the projection is incorrect in the error trap.  Can you change to above and test?

code/common/compress.q
```diff
@@ -158,23 +158,24 @@ summarystats:{
...
                 / move the hash files too.
                 if[78 <= type sf; system "r ", (last ":" vs string compressedFile),"# ", (last ":" vs string filetoCompress),"#"];
                 /-log to the table if the algo wasn't 0
-                $[not 0=algo;statstab ,: (filetoCompress;algo;(-21!filetoCompress)`compressedLength;sizeuncomp);statstab ,: (filetoCompress;algo;comprL;sizeuncomp)]];
+                statstab,:$[not 0=algo;(filetoCompress;algo;(-21!filetoCompress)`compressedLength;sizeuncomp);(filetoCompress;algo;comprL;sizeuncomp)]];
             [$[(not count (-21!compressedFile));
-		[.lg.o[`compression; "Failed to compress file ",(string filetoCompress)];hdel compressedFile];
-		.lg.o[`compression; $[algo=0;"Decompressed ";"Compressed "], "file ", (string compressedFile), " doesn't match original. Deleting new file"]; hdel compressedFile]]]
+		[.lg.o[`compression; "Failed to compress file ",string[filetoCompress]];hdel compressedFile];
+		.lg.o[`compression;cmp,"compressed ","file ",string[compressedFile]," doesn't match original. Deleting new file"];hdel compressedFile;]]]
```
jonnypress: I think this is a bug in the original.  The last two statements are both for the case where the compressed and uncompressed file do not match.  So they should be wrapped in a block statement

[.lg.o[`compression;cmp,"compressed ","file ",string[compressedFile]," doesn't match original. Deleting new file"];hdel compressedFile]]]]

code/common/compress.q
```diff
@@ -158,23 +158,24 @@ summarystats:{
...
     $[((0 = count -21!filetoCompress) & not 0 = algo)|((not 0 = count -21!filetoCompress) & 0 = algo);
-        [.lg.o[`compression;$[algo=0;"Decompressing ";"Compressing "],"file ", (string filetoCompress), " with algo: ", (string algo), ", blocksize: ", (string blocksize), ", and level: ", (string level), "."];
+        [.lg.o[`compression;cmp,"compressing ","file ", (string filetoCompress), " with algo: ", (string algo), ", blocksize: ", (string blocksize), ", and level: ", (string level), "."];
          / perform the compression/decompression
         if[0=algo;comprL:(-21!filetoCompress)`compressedLength];
         -19!(filetoCompress;compressedFile;blocksize;algo;level);
          / check the compressed/decomp file and move if appropriate; else delete compressed file and log error
         $[((get compressedFile)~sf:get filetoCompress) & (count -21!compressedFile) or algo=0;
-            [.lg.o[`compression;"File ", $[algo=0;"decompressed ";"compressed "],"successfully; matches orginal. Deleting original."];
+            [.lg.o[`compression;"File ",cmp,"compressed ","successfully; matches orginal. Deleting original."];
```
jonnypress: Given that we are modifying these for style, the indentation on the case statements is confusing here.  Can you line it up so it is clear which lines which in which block?

code/common/compress.q
```diff
@@ -158,23 +158,24 @@ summarystats:{
...
             [$[(not count (-21!compressedFile));
-		[.lg.o[`compression; "Failed to compress file ",(string filetoCompress)];hdel compressedFile];
-		.lg.o[`compression; $[algo=0;"Decompressed ";"Compressed "], "file ", (string compressedFile), " doesn't match original. Deleting new file"]; hdel compressedFile]]]
+		[.lg.o[`compression; "Failed to compress file ",string[filetoCompress]];hdel compressedFile];
+		.lg.o[`compression;cmp,"compressed ","file ",string[compressedFile]," doesn't match original. Deleting new file"];hdel compressedFile;]]]
         ];
         / if already compressed/decompressed, then log that and skip.
         [((not 0 = count -21!filetoCompress) & not 0 = algo)|((0 = count -21!filetoCompress) & 0 = algo)];
-        .lg.o[`compression; "file ", (string filetoCompress), " is already ",$[0=algo; "decompressed";"compressed"],". Skipping this file"]]}
+        .lg.o[`compression; "file ", (string filetoCompress), " is already ",cmp,"compressed",". Skipping this file"];]}
```
jonnypress: This condition is a bit weird, from looking at the combinations we don't need this check: 

 [((not 0 = count -21!filetoCompress) & not 0 = algo)|((0 = count -21!filetoCompress) & 0 = algo)]

(which also doesn't need to be a block statement) 
as this could just be the else case.  So I think you can just get rid of 

 [((not 0 = count -21!filetoCompress) & not 0 = algo)|((0 = count -21!filetoCompress) & 0 = algo)]

and then have 

.lg.o[`compression; "file ", (string filetoCompress), " is already ",cmp,"compressed",". Skipping this file"]]}

code/common/compress.q
jzabroski: Thank you very much! Reading these commit code reviews are invaluable to me.

PR #80, opened 2018-03-07
Response formatter

paulrdouglas:
Response format for sync and async requests.

Takes 3 args - `status, sync, result` to allow specific functionality for each case.

Defaults to throw error for sync errors, else passes through the error string or result.

Updated `sendclientreply` to take extra argument, `status`, for async calls

### DIFF
--- a/code/processes/gateway.q
+++ b/code/processes/gateway.q
@@ -248,7 +248,7 @@ removeserverhandle:{[serverh]
  // 1) queries sent to this server but no reply back yet
  qids:where {[res;id] any (::)~/:res[1;where id=res[1;;0];1]}[;serverid] each results;
  // propagate an error back to each client
- sendclientreply[;.gw.errorprefix,"backend ",(string servertype)," server handling query closed the connection";0b] each qids;
+ sendclientreply[;.gw.errorprefix,"backend ",string[servertype]," server handling query closed the connection";0b] each qids;
  finishquery[qids;1b;serverh]; 
 
  // 2) queries partially run + waiting for this server
@@ -259,13 +259,13 @@ removeserverhandle:{[serverh]
 	s:where (::)~/:res[1;;1]; 
 	$[11h=type s; not all s in aTypes; not all any each s in\: aIDs] 
 	}[;serverid;activeServerIDs;activeServerTypes] each results _ 0Ni;
- sendclientreply[;.gw.errorprefix,"backend ",(string servertype)," server for running query closed the connection";0b] each qids2;
+ sendclientreply[;.gw.errorprefix,"backend ",string[servertype]," server for running query closed the connection";0b] each qids2;
  finishquery[qids2;1b;serverh]; 
 
  // 3) queries not yet run + waiting for this server
  qids3:exec queryid from .gw.queryqueue where null submittime, not `boolean${$[11h=type z; all z in x; all any each z in\: y]}[activeServerTypes;activeServerIDs] each servertype; 
  // propagate an error back to each client
- sendclientreply[;.gw.errorprefix,"backend ",(string servertype)," server for queued query closed the connection";0b] each qids3;
+ sendclientreply[;.gw.errorprefix,"backend ",string[servertype]," server for queued query closed the connection";0b] each qids3;
  finishquery[qids3;1b;serverh]; 
 
  // mark the server as inactive
@@ -446,7 +446,7 @@ asyncexecjpt:{[query;servertype;joinfunction;postback;timeout]
 	servertype:res;
  ]]];
  if[count errStr;
-  @[neg .z.w;.gw.formatresponse[0b;`async;$[()~postback;errStr;$[-11h=type postback;enlist postback;postback],(enlist query),enlist errStr]];()];
+  @[neg .z.w;.gw.formatresponse[0b;`async;$[()~postback;errStr;$[-11h=type postback;enlist postback;postback],enlist[query],enlist errStr]];()];
   :()];
 
  addquerytimeout[query;servertype;queryattributes;joinfunction;postback;timeout];
@@ -568,7 +568,8 @@ reloadstart:{
  /- extract ids of queries not yet returned
  qids:exec queryid from .gw.queryqueue where 1<count each distinct each{@[(exec serverid!servertype from .gw.servers)@;x;x]}each servertype,null returntime;
  /- propagate a timeout error to each client
- if[count qids;.gw.sendclientreply[;.gw.errorprefix,"query did not return prior to eod reload";0b]each qids;.gw.finishquery[qids;1b;0Ni]];}
+ if[count qids;.gw.sendclientreply[;.gw.errorprefix,"query did not return prior to eod reload";0b]each qids;.gw.finishquery[qids;1b;0Ni]];
+ };
 
 reloadend:{
  .lg.o[`reload;"reload end called"];
--- a/code/processes/gateway.q
+++ b/code/processes/gateway.q
@@ -73,7 +73,7 @@
 // addserver and asyncexec, with the (servertypes) parameter projected to a single server of (for example) `standardserver
 
 \d .gw
-
+formatresponse:@[value;`.gw.formatresponse;{[x;y]if[10h=type y;'y];y}]
 synccallsallowed:@[value;`.gw.synccallsallowed; 0b]		// whether synchronous calls are allowed
 querykeeptime:@[value;`.gw.querykeeptime; 0D00:30]		// the time to keep queries in the 
 errorprefix:@[value;`.gw.errorprefix; "error: "]		// the prefix for clients to look for in error strings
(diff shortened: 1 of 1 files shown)


code/processes/gateway.q
```diff
@@ -74,6 +74,7 @@
 
 \d .gw
 
+formatresponse:@[value;`.gw.formatresponse;{{[status;sync;result] :$[not[status]and sync=`sync;'result;result];}}]
```
jonathonmcmurray: maybe add a comment to this line to explain the default behaviour?

ThomasSmyth: Don't need the colon or semi colon in a single line lambda.

code/processes/gateway.q
```diff
@@ -245,7 +246,7 @@ removeserverhandle:{[serverh]
  // 1) queries sent to this server but no reply back yet
  qids:where {[res;id] any (::)~/:res[1;where id=res[1;;0];1]}[;serverid] each results;
  // propagate an error back to each client
- sendclientreply[;.gw.errorprefix,"backend ",(string servertype)," server handling query closed the connection"] each qids;
+ sendclientreply[;.gw.errorprefix,"backend ",(string servertype)," server handling query closed the connection";0b] each qids;
```
ThomasSmyth: Square brackets for passing arguments to string.

code/processes/gateway.q
```diff
@@ -256,13 +257,13 @@ removeserverhandle:{[serverh]
 	s:where (::)~/:res[1;;1]; 
 	$[11h=type s; not all s in aTypes; not all any each s in\: aIDs] 
 	}[;serverid;activeServerIDs;activeServerTypes] each results _ 0Ni;
- sendclientreply[;.gw.errorprefix,"backend ",(string servertype)," server for running query closed the connection"] each qids2;
+ sendclientreply[;.gw.errorprefix,"backend ",(string servertype)," server for running query closed the connection";0b] each qids2;
```
ThomasSmyth: Squares again.

code/processes/gateway.q
```diff
@@ -256,13 +257,13 @@ removeserverhandle:{[serverh]
...
  finishquery[qids2;1b;serverh]; 
 
  // 3) queries not yet run + waiting for this server
  qids3:exec queryid from .gw.queryqueue where null submittime, not `boolean${$[11h=type z; all z in x; all any each z in\: y]}[activeServerTypes;activeServerIDs] each servertype; 
  // propagate an error back to each client
- sendclientreply[;.gw.errorprefix,"backend ",(string servertype)," server for queued query closed the connection"] each qids3;
+ sendclientreply[;.gw.errorprefix,"backend ",(string servertype)," server for queued query closed the connection";0b] each qids3;
```
ThomasSmyth: Squares again.

code/processes/gateway.q
```diff
@@ -439,7 +440,7 @@ asyncexecjpt:{[query;servertype;joinfunction;postback;timeout]
 	servertype:res;
  ]]];
  if[count errStr;
-  @[neg .z.w;$[()~postback;errStr;$[-11h=type postback;enlist postback;postback],(enlist query),enlist errStr];()];
+  @[neg .z.w;.gw.formatresponse[0b;`async;$[()~postback;errStr;$[-11h=type postback;enlist postback;postback],(enlist query),enlist errStr]];()];
```
ThomasSmyth: Should be able to remove nested if-else squares + use squares for enlist.

code/processes/gateway.q
```diff
@@ -560,7 +562,7 @@ reloadstart:{
  /- extract ids of queries not yet returned
  qids:exec queryid from .gw.queryqueue where 1<count each distinct each{@[(exec serverid!servertype from .gw.servers)@;x;x]}each servertype,null returntime;
  /- propagate a timeout error to each client
- if[count qids;.gw.sendclientreply[;.gw.errorprefix,"query did not return prior to eod reload"]each qids;.gw.finishquery[qids;1b;0Ni]];}
+ if[count qids;.gw.sendclientreply[;.gw.errorprefix,"query did not return prior to eod reload";0b]each qids;.gw.finishquery[qids;1b;0Ni]];}
```
ThomasSmyth: Closing `}` should be on next line and followed by a semi colon.

code/processes/gateway.q
jonnypress: Yes, comment to define default would be good.
Explain what the parameters are e.g. status is 1b for success and 0b for error etc.

code/processes/gateway.q
```diff
@@ -412,13 +413,13 @@ getserveridstype:{[att;typ]
...
   serverids:first value flip $[99h=type res; key res; res];
-  if[all 0=count each serverids; '"no servers match ",string[typ]," requested attributes"];
+  if[all 0=count each serverids;.gw.formatresponse[0b;`sync;"no servers match ",string[typ]," requested attributes"]];
   :serverids;
  }
 
 // execute an asynchronous query
 asyncexecjpt:{[query;servertype;joinfunction;postback;timeout]
- if[.gw.permissioned;if[.pm.allowed[.z.u; query];'"User is not permissioned to run this query from the gateway"]];
+ if[.gw.permissioned;if[.pm.allowed[.z.u; query];.gw.formatresponse[0b;`sync;"User is not permissioned to run this query from the gateway"]]];
```
jonnypress: I don't think this is correct - I think this is a bug in the original.  This should also be propagated back via the neg handle

code/processes/gateway.q
```diff
@@ -330,7 +331,7 @@ getserversindependent:{[req;att;besteffort]
  alldone:1+first where all each all each' maxs value s;
 
  if[(null alldone) and not besteffort;
-        '"getserversindependent: cannot satisfy query as not all attributes can be matched"];
+  .gw.formatresponse[0b;`sync;"getserversindependent: cannot satisfy query as not all attributes can be matched"]];
```
jonnypress: I don't think you need the response formatter on any of these getservers* functions.  These are invoked in both asyncexec and syncexec. 

If you look at line 437 the getservers function is invoked in an error trap and then you have added a format function to the result of that.  So I think the way you have it currently it would get passed through the formatter twice.  

It is also invoked in syncexecj at line 458.  I think what you need to do is change the call at 458 to error trap the result, and then pass it through the result formatter e.g. something like: 

serverids:@[getserverids;servertype;{.gw.formatresponse[0b;`sync;.gw.errorprefix,x]}];

code/processes/gateway.q
```diff
@@ -412,13 +415,17 @@ getserveridstype:{[att;typ]
...
   :serverids;
  }
 
 // execute an asynchronous query
 asyncexecjpt:{[query;servertype;joinfunction;postback;timeout]
- if[.gw.permissioned;if[.pm.allowed[.z.u; query];'"User is not permissioned to run this query from the gateway"]];
+ if[.gw.permissioned;
+  if[.pm.allowed[.z.u; query];
+   @[neg .z.w;.gw.formatresponse[0b;`async;"User is not permissioned to run this query from the gateway"];()];
+   :()];
```
ThomasSmyth: Newline for closing bracket.

code/processes/gateway.q
```diff
@@ -474,9 +481,10 @@ syncexecj:{[query;servertype;joinfunction]
  // check if there are any errors in the returned results
  $[all res[;0];
   // no errors - join the results
-  @[joinfunction;res[;2];{'`$"failed to apply supplied join function to results: ",x}];
+  [s:@[{(1b;x y)}joinfunction;res[;2];{(0b;"failed to apply supplied join function to results: ",x)}];
+    .gw.formatresponse[s 0;`sync;s 1]];
   [failed:where not res[;0];
-   '`$"queries failed on server(s) ",(", " sv string exec servertype from servers where handle in handles failed),".  Error(s) were ","; " sv res[failed][;2]]] 
+   .gw.formatresponse[0b;`sync;"queries failed on server(s) ",(", " sv string exec servertype from servers where handle in handles failed),".  Error(s) were ","; " sv res[failed][;2]]]]
```
ThomasSmyth: Indented by 2 spaces on 485 and by 1 here, seems to be indentation of 1 space in conditionals throughout the file.

code/processes/gateway.q
```diff
@@ -474,9 +481,10 @@ syncexecj:{[query;servertype;joinfunction]
  // check if there are any errors in the returned results
  $[all res[;0];
   // no errors - join the results
-  @[joinfunction;res[;2];{'`$"failed to apply supplied join function to results: ",x}];
+  [s:@[{(1b;x y)}joinfunction;res[;2];{(0b;"failed to apply supplied join function to results: ",x)}];
+    .gw.formatresponse[s 0;`sync;s 1]];
   [failed:where not res[;0];
-   '`$"queries failed on server(s) ",(", " sv string exec servertype from servers where handle in handles failed),".  Error(s) were ","; " sv res[failed][;2]]] 
+   .gw.formatresponse[0b;`sync;"queries failed on server(s) ",(", " sv string exec servertype from servers where handle in handles failed),".  Error(s) were ","; " sv res[failed][;2]]]] 
  }
```
ThomasSmyth: Add semi colon.

PR #85, opened 2018-03-28
Gatewayqueries

JuliaPartridge:
Gateway script has been updated so that hdb queries require a date.

### DIFF
--- a/code/processes/gateway.q
+++ b/code/processes/gateway.q
@@ -284,8 +284,11 @@ checktimeout:{
   finishquery[qids;1b;0Ni]];
  }
 
-
-
+// check if hdb query has date
+checkdate:{[query;servertype]
+ if[servertype=`hdb;
+ if[not query like "*date*"; '"query requires a date"]]
+ }
 
 /- NEED TO FILTER ON PREFERENCES FIRST
 
@@ -453,6 +456,8 @@ syncexecj:{[query;servertype;joinfunction]
  if[not .gw.synccallsallowed; '`$"synchronous calls are not allowed"];
  // check if the gateway allows the query to be called
  if[.gw.permissioned;if[not .pm.allowed [.z.u;query];'"User is not permissioned to run this query from the gateway"]];
+ // check if hdb query has a date clause
+ checkdate[query;servertype];
  // check if we have all the servers active
  serverids:getserverids[servertype];
  // check if gateway in eod reload phase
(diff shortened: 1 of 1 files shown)


code/processes/gateway.q
```diff
@@ -284,8 +284,13 @@ checktimeout:{
   finishquery[qids;1b;0Ni]];
  }
 
-
-
+// check if hdb query has date
+checkdate:{[query;servertype]
+ if[servertype=`hdb;
+  $[type [query] = 0h;
+  if[not any raze[string query 2] like "*date*"; '"query requires a date"];
```
jonathonmcmurray: This isn't very nice, and it shouldn't be necessary to convert to a string...In fact, I think you'd probably be best to *not* work with strings at all....

```
q)0N!parse"select from trades where date=2018.01.01";
(?;`trades;,,(=;`date;2018.01.01);0b;())
q)q:(?;`trades;enlist(=;`date;2018.01.01);0b;())
q)q[2]
= `date 2018.01.01
q)any `date in/:q[2]
1b
```

Something like this will work if the query is passed as a string (in which case, `parse` it) or a functional form list (in which case, don't parse it)

Of course, you also need to know what type of partition the HDB is using - this is stored in `.Q.pf` http://code.kx.com/q/ref/dotq/#qpf-partition-type

Finally, you also need to handle cases where the client is calling a predefined function, rather than a Q-SQL query. E.g.

```
.gw.syncexec["vwap[`AAPL;2018.01.01]";`hdb]
```

Again, parsing this string may help...

```
q)parse"vwap[`AAPL;2018.01.01]"
`vwap
,`AAPL
2018.01.01
```

(of course, the string here doesn't contain the word `"date"`, but should absolutely be allowed to run)

code/processes/gateway.q
```diff
@@ -284,8 +284,13 @@ checktimeout:{
   finishquery[qids;1b;0Ni]];
  }
 
-
-
+// check if hdb query has date
+checkdate:{[query;servertype]
+ if[servertype=`hdb;
+  $[type [query] = 0h;
```
jonathonmcmurray: I don't think the `$` should be used like this - I think what you should be doing here is using the `$` to get the query into a useful format so that regardless of the incoming type, you can handle it the same way. My suggestion would be something along the lines of:

```
query:$[10h=type query;parse;]query
```

which will parse the query if it's a string, or not if it's a list and then either way, it's a list going forward.

Also, unnecessary whitespace between `type` and `[query]`, plus by rearranging as in my suggestion you can get rid of square brackets

PR #86, opened 2018-04-27
Add TorQ Low Power Mode; Improve Timer Performance

karanpatel12:
reduce freq of subscription check to 60s and .timer to 1s; disable heartbeat and timer call logging; .timer.nextruntime added to hold value of next time .timer has to execute

### DIFF
--- a/code/common/heartbeat.q
+++ b/code/common/heartbeat.q
@@ -7,7 +7,9 @@
 // Override the processwarning and processerror functions to implement required behaviour when warnings or errors are encountered
 // Override warningperiod and errorperiod functions to have bespoke warning and error periods for different process types
 // Use storeheartbeat function in the upd function to process heartbeats
+
 \d .hb
+
 enabled:@[value;`enabled;1b]					// whether the heartbeating is enabled
 subenabled:@[value;`subenabled;0b]                              // whether subcriptions to heartbeats are enabled
 debug:@[value;`debug;1b]					// whether to print debug information
@@ -104,10 +106,10 @@ getheartbeats:{[proctype]
     subscribe each handles]
  }
 
-
 \d .
-// set the heartbeat table to the top level namespace, to allow it to be initialis    ed in the pub/sub routine
-@[`.;`heartbeat;:;.hb.heartbeat];
+
+// set the heartbeat table to the top level namespace, to allow it to be initialised in the pub/sub routine
+heartbeat:.hb.heartbeat;
 if[(not @[value;`.proc.lowpowermode;0b]) & @[value;`enabled;1b];
  // add the checkheartbeat function to the timer
  $[@[value;`.timer.enabled;0b] and `publish in key `.ps;
--- a/code/common/heartbeat.q
+++ b/code/common/heartbeat.q
@@ -7,13 +7,11 @@
 // Override the processwarning and processerror functions to implement required behaviour when warnings or errors are encountered
 // Override warningperiod and errorperiod functions to have bespoke warning and error periods for different process types
 // Use storeheartbeat function in the upd function to process heartbeats
-
 \d .hb
-
 enabled:@[value;`enabled;1b]					// whether the heartbeating is enabled
-subenabled:@[value;`subenabled;0b]                                 // whether subcriptions to heartbeats are enabled
+subenabled:@[value;`subenabled;0b]                              // whether subcriptions to heartbeats are enabled
 debug:@[value;`debug;1b]					// whether to print debug information
-publishinterval:@[value;`publishinterval;0D00:00:30]		// how often heartbeats are published	
+publishinterval:@[value;`publishinterval;0D00:00:30]		// how often heartbeats are published
 checkinterval:@[value;`checkinterval;0D00:00:10]		// how often heartbeats are checked
 warningtolerance:@[value;`warningtolerance;1.5f]		// a process will move to warning state when it hasn't heartbeated in warningtolerance*checkinterval
 errortolerance:@[value;`errortolerance;2f]			// and to an error state when it hasn't heartbeated in errortolerance*checkinterval
@@ -24,7 +22,7 @@ subscribedhandles:0 0Ni
 // sym = proctype
 heartbeat:([]time:`timestamp$(); sym:`symbol$(); procname:`symbol$(); counter:`long$(); pid:`int$(); host:`$(); port:`int$())
 
-// process ID, hostname and port 
+// process ID, hostname and port
 pid:.z.i
 host:.z.h
 port:system"p"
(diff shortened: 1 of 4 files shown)


code/common/heartbeat.q
```diff
@@ -7,13 +7,11 @@
 // Override the processwarning and processerror functions to implement required behaviour when warnings or errors are encountered
 // Override warningperiod and errorperiod functions to have bespoke warning and error periods for different process types
 // Use storeheartbeat function in the upd function to process heartbeats
-
 \d .hb
-
```
ThomasSmyth: Bring the newlines back either side of `\d .hb`, will help keep the namespace change more visible.

code/common/heartbeat.q
```diff
@@ -108,10 +106,9 @@ getheartbeats:{[proctype]
 
 
 \d .
-
-if[.hb.enabled;
- // set the heartbeat table to the top level namespace, to allow it to be initialised in the pub/sub routine
- @[`.;`heartbeat;:;.hb.heartbeat];
+// set the heartbeat table to the top level namespace, to allow it to be initialis    ed in the pub/sub routine
+@[`.;`heartbeat;:;.hb.heartbeat];
```
ThomasSmyth: This was originally here, but if you're in the root namespace is there any need to use `@` here?

code/common/heartbeat.q
```diff
@@ -108,10 +106,9 @@ getheartbeats:{[proctype]
 
 
 \d .
-
-if[.hb.enabled;
- // set the heartbeat table to the top level namespace, to allow it to be initialised in the pub/sub routine
- @[`.;`heartbeat;:;.hb.heartbeat];
+// set the heartbeat table to the top level namespace, to allow it to be initialis    ed in the pub/sub routine
```
ThomasSmyth: Large space in this commented needs removed.

code/common/subscriptions.q
```diff
@@ -3,29 +3,31 @@
 \d .sub
 
 AUTORECONNECT:@[value;`AUTORECONNECT;1b];				//resubscribe to processes when they come back up
-checksubscriptionperiod:@[value;`checksubscribtionperiod;0D00:00:10];	//how frequently you recheck connections.  0D = never
+if[@[value;`.proc.lowpowermode;0b];
+  checksubscriptionperiod:@[value;`checksubscribtionperiod;0D00:01:00];	//how frequently you recheck connections.  0D = never
```
ThomasSmyth: Use spaces rather than tabs to bring this comment back in a bit.

code/common/timer.q
```diff
@@ -1,23 +1,22 @@
 // Functionality to extend the timer
 
 \d .timer
-
 enabled:@[value;`enabled;1b]				// whether the timer is enabled
 debug:@[value;`debug;0b]				// print when the timer runs any function
-logcall:@[value;`logcall;1b]				// log each timer call by passing it through the 0 handle
+logcall:(not @[value;`.proc.lowpowermode;0b]) & @[value;`logcall;1b]				// log each timer call by passing it through the 0 handle
```
ThomasSmyth: Comment is way out in space due to tabs.

jonathonmcmurray:
PR description and the commit message of only commit are both quite unreadable, please improve pr description (e.g. Bullet point things or something, as well as adding an overview of why the changes are being made), and use better commit messages going forward (same rules apply as in prodsupport stuff)

code/common/heartbeat.q
karanpatel12: so should I just replace this with heartbeat:.hb.heartbeat ?

jonnypress: yes sounds good

code/common/timer.q
jonnypress: yes line up the comments please

code/common/subscriptions.q
jonnypress: checksubscriptionperiod has a typo - b instead of p

I think it would be better to just turn this off (set to 0D) in low power mode.  Also, you still need a value for non lower power mode.  So the setting could probably be

checksubscriptionperiod:(not @[value;`.proc.lowpowermode;0b]) * @[value;`checksubscriptionperiod;0D00:00:10]

code/common/timer.q
```diff
@@ -74,24 +73,28 @@ runandreschedule:{
...
 if[.timer.enabled;
  $[@[{value x;1b};`.z.ts;0b];
    .z.ts:{[x;y] .timer.run now:.proc.cp[]; x@y}[.z.ts];
-   .z.ts:{.timer.run[.proc.cp[]]}];
+   .z.ts:{if[.proc.cp[]>.timer.nextruntime;.timer.run[.proc.cp[]]]}];
 
  // Set the timer to 200ms if not set already
  if[not system"t"; system"t 200"]];
-	
+if[.proc.lowpowermode;
```
jonnypress: could you add the error trap around this e.g. 

@[value;`.proc.lowerpowermode;0b]

config/settings/default.q
```diff
@@ -7,6 +7,7 @@ loadprocesscode:0b		// whether to load the process specific code defined at ${KD
 loadnamecode:0b			// whether to load the name specific code defined at ${KDBCODE}/{name of process}
 loadhandlers:1b			// whether to load the message handler code defined at ${KDBCODE}/handlers
 logroll:1b			// whether to roll the std out/err logs on a daily basis
+lowpowermode:0b			// Use TorQ in low cpu usage mode (No default logging for timer calls, Disabled heartbeating, Reduced timer frequency)
```
jonnypress: lower resubsciption check frequency as well

code/common/subscriptions.q
karanpatel12: line 40 in TorQ/config/settings/default.q defines the checksubscriptionperiod:0D00:00:10. So that would be the default for non low power mode. I don't think it needs to be explicitly defined in subscriptions.q?

jonnypress: The standard practice here is to redefine it in the file if it hasn't already been defined, rather than assuming it has already been defined.  That's why all those lines look like @[value.... ]

code/common/subscriptions.q
```diff
@@ -2,30 +2,32 @@
 
 \d .sub
 
-AUTORECONNECT:@[value;`AUTORECONNECT;1b];				//resubscribe to processes when they come back up
-checksubscriptionperiod:@[value;`checksubscribtionperiod;0D00:00:10];	//how frequently you recheck connections.  0D = never
+AUTORECONNECT:@[value;`AUTORECONNECT;1b];									//resubscribe to processes when they come back up
+if[@[value;`.proc.lowpowermode;0b];
+  checksubscriptionperiod:(not @[value;.proc.lowpowermode;0b]) * @[value;checksubscriptionperiod;0D00:00:10]  	//how frequently you recheck connections.  0D = never
```
jonnypress: We still need a default setting for checksubscriptionperiod if lowpowermode=0b

PR #87, opened 2018-04-30
Add initialisation function and start timer

JuliaPartridge:
Add a variable .proc.starttime which is set as soon as a process is started.
Add a default initialisation function .proc.init that is overridden when the script is loaded and the connections have been established.

### DIFF
--- a/torq.q
+++ b/torq.q
@@ -8,9 +8,8 @@ initialised:0b
 
 // function to add functions to initialisation list
 initlist:()
-addinitlist:{[x]if[1<count x;.proc.initlist,:enlist x]};
+addinitlist:{[x].proc.initlist,:enlist x};
   
-//add to init list functions explicitly defined on command line
 generalusage:@[value;`generalusage;"General:
  This script should form the basis of a production kdb+ environment. 
  It can be sourced from other files if required, or used as a launch script before loading other files/directories 
@@ -590,10 +589,13 @@ if[@[value;`.servers.STARTUP;0b]; .servers.startup[]]
 
 // function to execute functions in .proc.initlist
 .proc.init:{
-	if[count .proc.initlist;
-	{[a]@[value;a;{[x;a].lg.e[`init;"unable to run initialisation function for ",a," - ",x," error"]}[;a]]}
-	each .proc.initlist;.proc.initlist:()];
-	}
+	$[count .proc.initlist;
+		[{[a].lg.o[`init;"attemping to run initialisation: ",-3!a];
+		@[value;a;
+		{[x;a].lg.e[`init;x," error - failed to run initialisation: ",-3!a]}[;a]]}
+		each .proc.initlist;.proc.initlist:()];
+		.lg.o[`init;"no initialisation functions found"]];
+ }
 
 if[count .proc.initlist;.proc.init[]]
 
--- a/torq.q
+++ b/torq.q
@@ -6,6 +6,13 @@ loaded:1b
 // Initialised flag - used to check if the process is still initialisation
 initialised:0b
 
+// function to add functions to initialisation list
+initlist:()
+addinitlist:{[x]if[1<count x;.proc.initlist,:enlist x]};
+  
+  //$[10h=type x;.proc.initlist,:"S"$x;.proc.initlist,:x]};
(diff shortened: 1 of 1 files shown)


torq.q
```diff
@@ -6,6 +6,13 @@ loaded:1b
 // Initialised flag - used to check if the process is still initialisation
 initialised:0b
 
+// function to add functions to initialisation list
+initlist:()
+addinitlist:{[x]if[1<count x;.proc.initlist,:enlist x]};
+  
+  //$[10h=type x;.proc.initlist,:"S"$x;.proc.initlist,:x]};
+//add to init list functions explicitly defined on command line
```
ThomasSmyth: You can remove these if they are no longer required.

torq.q
```diff
@@ -583,11 +590,18 @@ if[@[value;`.ps.loaded;0b]; .ps.initialise[]]
 // initialise connections
 if[@[value;`.servers.STARTUP;0b]; .servers.startup[]]
 
+// function to execute functions in .proc.initlist
+.proc.init:{if[count .proc.initlist;{[a]@[value;a;{[x;a].lg.e[`init;"unable to run initialisation function for ",a," - ",x," error"]}[;a]]} each .proc.initlist;.proc.initlist:()]};
```
ThomasSmyth: Split this up over multiple lines, it's hard to read as is.

JuliaPartridge: When I split it up should I do double space indentation or do I keep it consistent with the rest of the script?

ThomasSmyth: Looks like indentation is not consistent across this file, possibly because it's a mix of tabs and spaces. Try copy that style if possible.

torq.q
```diff
@@ -6,6 +6,11 @@ loaded:1b
 // Initialised flag - used to check if the process is still initialisation
 initialised:0b
 
+// function to add functions to initialisation list
+initlist:()
+addinitlist:{[x]if[1<count x;.proc.initlist,:enlist x]};
```
jonnypress: What is the reason for the 1<count x check?  Which case is that catching?

torq.q
```diff
@@ -6,6 +6,11 @@ loaded:1b
...
+  
+//add to init list functions explicitly defined on command line
```
jonnypress: Is this comment necessary? It looks like it should be the other side of the new code, if it is still valid.

torq.q
```diff
@@ -583,11 +588,22 @@ if[@[value;`.ps.loaded;0b]; .ps.initialise[]]
 // initialise connections
 if[@[value;`.servers.STARTUP;0b]; .servers.startup[]]
 
+// function to execute functions in .proc.initlist
+.proc.init:{
+	if[count .proc.initlist;
+	{[a]@[value;a;{[x;a].lg.e[`init;"unable to run initialisation function for ",a," - ",x," error"]}[;a]]}
+	each .proc.initlist;.proc.initlist:()];
```
jonnypress: stick the second statement on a new line

torq.q
```diff
@@ -583,11 +588,22 @@ if[@[value;`.ps.loaded;0b]; .ps.initialise[]]
 // initialise connections
 if[@[value;`.servers.STARTUP;0b]; .servers.startup[]]
 
+// function to execute functions in .proc.initlist
+.proc.init:{
+	if[count .proc.initlist;
+	{[a]@[value;a;{[x;a].lg.e[`init;"unable to run initialisation function for ",a," - ",x," error"]}[;a]]}
```
jonnypress: I think you need to indent the condition

Will this actually work?  What if a is a function like (`myfunc;1;2) - surely the error trap will fail as it can't print it?  I think you need to -3! it. 

Also can you add more logging.   For each initialisation function that runs we need a log line 

"running initialisation function: ",-3!a

If we don't have any to run then we need a log line to say that, e.g. "no initialisation functions found"

torq.q
JuliaPartridge: I think it was to make sure a function was passed with at least an argument. I don't think it's very good though, I can remove that.

torq.q
JuliaPartridge: This comment was supposed to be deleted

PR #88, opened 2018-04-30
Start stop script

JuliaPartridge:
This is the new TorQ start and stop script which includes the following features:
- Starting one or multiple processes
- Printing the start up line for one or multiple processes
- Stopping one or multiple processes
- Summary table of processes - including PID and port number of each process
- Debug mode for a process
- Different q commands
- Different csv files
- Appending/overriding default start up lines


jonathonmcmurray:
Please make the description more descriptive. What's new about it? What's changed? Why do we need a new script?

setenv.sh
```diff
@@ -6,9 +6,24 @@ export KDBCODE=${TORQHOME}/code
 export KDBLOG=${TORQHOME}/logs
 export KDBHTML=${TORQHOME}/html
 export KDBLIB=${TORQHOME}/lib
+export KDBAPPCODE=${TORQHOME}/code/tick
```
jonnypress: This isn't right I think - should have the /tick on the end

startstoptorq.sh
```diff
@@ -0,0 +1,275 @@
+# Load the environment
+. ./setenv.sh
+
+# sets the base port for a default TorQ installation
+export KDBHDB=${PWD}/hdb/database
+export KDBWDB=${PWD}/wdbhdb
+export KDBSTACKID="-stackid ${KDBBASEPORT}"
+export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$KDBLIB/l32
+export DEFAULTCSV=${KDBAPPCONFIG}/process.csv
```
jonnypress: Can we move these into the setenv.sh script?  I think it's historical reasons that we can probably get rid of that these are in separate places.  

Also I think we should define three variables below TORQHOME.  We should have 
TORQDATA
TORQAPPHOME
 
By default, these can each just get ${PWD} as their value, but the idea is to make it easier for people to modify.  

KDBLOG can then be ${TORQDATA}/logs
KDBHDB can be ${TORQDATA}/hdb/database
KDBWDB can be ${TORQDATA}/wdbhdb
KDBAPPCODE can be ${TORQAPPHOME}/code
KDBAPPCONFIG can be ${TORQAPPHOME}/config

I think that would make it backwardly compatible, but more flexible - so a new installation just has to change TORQHOME, TORQAPPHOME and TORQDATA to have all three in different places

jonnypress:
Oh also... I think the new start/stop script should just be called torq.sh

startstoptorq.sh
```diff
@@ -0,0 +1,275 @@
...
+
+getfield() {
+  fieldno=`awk -F, '{if(NR==1) for(i=1;i<=NF;i++){if($i=="'$2'") print i}}' $csvpath` 		    # get number for field based on headers
+  fieldval=`awk -F, '{if(NR == '$1') print $'$fieldno'}' $csvpath`                    		    # pull one field from one line of file
+  echo $fieldval | envsubst                                                                         # substitute env vars
```
jonathonmcmurray: misaligned comment

startstoptorq.sh
```diff
@@ -0,0 +1,275 @@
...
+  fieldval=`getfield $1 $2`
+  if [ "" == "$fieldval" ]; then                                                                    # check for empty string
+    echo ""
+  else
+    echo " -"$2 $fieldval
+  fi
+ }
+
+getqcmd() {
+  fieldval=`getfield $1 $2`
```
jonathonmcmurray: seems like this is a function for getting a specific field, why does it need to have two params? Surely second is always going to be `"qcmd"`, why not just hardcode this in here?

startstoptorq.sh
```diff
@@ -0,0 +1,275 @@
...
+  fi
+ }
+
+findproc() {
+  pgrep -f "\-procname $1 $KDBSTACKID \-proctype $(getfield $1 proctype)"
+ }
+
+startline() {
+  procno=`awk '/'$1'/{print NR}' $csvpath`							    # get line number for file
+  qcmd=`awk '/'$1'/{print NR}' $csvpath`
```
jonathonmcmurray: is this used for anything? doesn't seem like it is, and it's the same as `procno` so surely if you do need it you can just use that?

startstoptorq.sh
```diff
@@ -0,0 +1,275 @@
...
+  params="proctype U localtime g T w load"  							    # list of params to read from config
+  sline="${TORQHOME}/torq.q -procname $1 ${KDBSTACKID}"					   	    # base part of startup line
```
jonathonmcmurray: misaligned comment

startstoptorq.sh
```diff
@@ -0,0 +1,275 @@
...
+    a=`parameter $procno $p`;									    # get param
+    sline="$sline$a";     									    # append to startup line
+  done
+  qcmd=`getqcmd $procno "qcmd"`;
+  sline="$qcmd $sline $(getfield $procno extras) -procfile $csvpath"
+  echo $sline
+ }
+
+start() {
+  if [ -z `findproc $1` ]; then									    # check process not running
```
jonathonmcmurray: misaligned comment

startstoptorq.sh
```diff
@@ -0,0 +1,275 @@
...
+    sline="nohup $(startline $1) $extras </dev/null >${KDBLOG}/torq${1}.txt 2>&1 &"                 # line to run each process
+    echo `date '+%H:%M:%S'` "| Starting $1..."
+    eval $sline                                                                                     # redirect output and run in background
+  else
+    echo `date '+%H:%M:%S'` "| $1 already running"
+  fi 
+ }
+
+print() {
+  sline="nohup $(startline $1) $extras </dev/null >${KDBLOG}/torq${1}.txt 2>&1 &"                   # line to run each process in print mode
```
jonathonmcmurray: why isn't all of this stuff done in the `startline` function? Seems like you add it most times you call the function, perhaps could have an additional arg to function if you *don't* want this stuff added. Or a wrapper function to add it or something?

startstoptorq.sh
```diff
@@ -0,0 +1,275 @@
...
+    sline="$sline$a";     									    # append to startup line
+  done
+  qcmd=`getqcmd $procno "qcmd"`;
+  sline="$qcmd $sline $(getfield $procno extras) -procfile $csvpath"
+  echo $sline
+ }
+
+start() {
+  if [ -z `findproc $1` ]; then									    # check process not running
+    sline="nohup $(startline $1) $extras </dev/null >${KDBLOG}/torq${1}.txt 2>&1 &"                 # line to run each process
```
jonathonmcmurray: where is `$extras` set? is this a global variable? I think convention in shell scripts is to make global names all caps

startstoptorq.sh
```diff
@@ -0,0 +1,275 @@
...
+    eval "kill -15 $pid"                                                                            # kill process pid
+  fi
+ }
+
+getall() {
+  procs=`awk -F, '{if(NR>1) print $4}' $csvpath`                                                    # get all processes from csv
+  start=""
+  for a in $procs;
+  do
+    procno=`awk '/'$a'/{print NR}' $csvpath`  		                                            # get line number for file
```
jonathonmcmurray: misaligned comment

startstoptorq.sh
```diff
@@ -0,0 +1,275 @@
...
+  avail=()
+  for i in $input;
+  do 
+    if [[ `echo "$procs" | grep -w "$i"` ]]; then                                                   # check input process is valid
+      avail+="$i "                                                                                  # get only valid processes
+    else 
+      echo `date '+%H:%M:%S'` "| $i failed - unavailable processname"
+    fi
+  done
+  procs=$avail                                                                                      # assign valid processes
```
jonathonmcmurray: setting a global, should use caps I think

startstoptorq.sh
```diff
@@ -0,0 +1,275 @@
...
+ }
+
+getprocs() {
+  if [ "$2" == "all" ]; then 
+    procs=$(getall)                                                                                 # get all processes
```
jonathonmcmurray: global - caps again

startstoptorq.sh
```diff
@@ -0,0 +1,275 @@
...
+  else
+    shift                                                                                           # ignore first argument
+    checkinput $@                                                                                   # check input process names
+  fi
+ }
+
+flag() {
```
jonathonmcmurray: this function could maybe do with an explanation, not very clear what's going on here

startstoptorq.sh
```diff
@@ -0,0 +1,275 @@
...
+getextrascsv() {
+  eval flagextras "extras";
+  eval flagcsv "csv";
+  length=$(($#-$n-2));
+  array=${@:1:$length};                                                                             # arguments without extras and csv flag
+  getprocs $array;
+ }
+
+checkextrascsv() {
+  if [[ `echo ${BASH_ARGV[*]} | grep -e extras` ]] && [[ `echo ${BASH_ARGV[*]} | grep -e csv` ]]; then
```
jonathonmcmurray: In some places single `[` are used in if statements, here it's `[[` - make this consistent please

startstoptorq.sh
```diff
@@ -0,0 +1,275 @@
...
+stopprocs() {
+  for p in $procs;
+  do
+    stop $p;
+  done
+ }
+
+usage() {
+  printf -- "Arguments:\n"
+  printf -- "-procs                               to list all processes\n"
```
jonathonmcmurray: I think maybe some additional indentation on these strings would be good, to separate from section headers

setenv.sh
JuliaPartridge: It does have /tick on the end - did you mean shouldn't? If so, I changed this because for the feed process.csv loads the script is with using `${KDBAPPCODE}/feed.q`. This was how it was set up in the process.csv I was given.

I will change it so `export KDBAPPCODE=${TORQHOME}/code` and the feed loads `${KDBAPPCODE}/tick/feed.q`

setenv.sh
```diff
@@ -1,14 +1,44 @@
...
-export KDBCONFIG=${TORQHOME}/config
-export KDBCODE=${TORQHOME}/code
-export KDBLOG=${TORQHOME}/logs
-export KDBHTML=${TORQHOME}/html
-export KDBLIB=${TORQHOME}/lib
+export TORQDATA=${PWD}
+export TORQAPPHOME=${PWD}
+export KDBLOG=${TORQDATA}/logs
+export KDBHTML=${TORQDATA}/html
+export KDBLIB=${TORQDATA}/lib
```
jonathonmcmurray: I don't think `KDBHTML` and `KDBLIB` should be relative to `TORQDATA`, should be relative to `TORQHOME` surely?

setenv.sh
```diff
@@ -1,14 +1,44 @@
...
+	mkdir ${PWD}/certs
+	curl -s  https://curl.haxx.se/ca/cacert.pm > ${PWD}/certs/cabundle.pem
+	echo "`date`    The SSL securiity certificate has been downloaded to ${PWD}/certs/cabundle.pem" </dev/null >>$KDBLOG/torqsslcert.txt 
+	export SSL_CA_CERT_FILE=${PWD}/certs/cabundle.pem
+else
+	echo "`date`    The SSL security certificate already exists. If https requests fail it may be because of inappropriate certification." </dev/null >>$KDBLOG/torqsslcert.txt 
+fi
+
+
+# sets the base port for a default TorQ installation
```
jonathonmcmurray: this comment refers to 3 lines later?

setenv.sh
```diff
@@ -1,14 +1,44 @@
...
+export KDBHDB=${TORQDATA}/hdb/database
+export KDBWDB=${TORQDATA}/wdbhdb
+export KDBSTACKID="-stackid ${KDBBASEPORT}"
+export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$KDBLIB/l32
```
jonathonmcmurray: I think all these env vars should be with the other env vars at the top of the file

Also above comment (l20-24) mentions setting this last one differently depending on OS, should it be set like this in this file?

setenv.sh
```diff
@@ -1,14 +1,44 @@
...
+export KDBLIB=${TORQDATA}/lib
+export KDBAPPCODE=${TORQAPPHOME}/code
+export KDBCONFIG=${TORQAPPHOME}/config
+export KDBCODE=${TORQAPPHOME}/code
+
+# sets the application specific configuration directory
+export KDBAPPCONFIG=${TORQHOME}/appconfig
+# set KDBBASEPORT to the default value for a TorQ Installation
+export KDBBASEPORT=41000
+# set DEFAULTCSV to the default process csv
```
jonathonmcmurray: inline comments seem like they would be more consistent with the style of `torq.sh`

torq.sh
```diff
@@ -0,0 +1,263 @@
...
+ }
+
+usage() {
+  printf -- "Arguments:\n"
+  printf -- "-procs       	                        to list all processes\n"
+  printf -- "-summary             	                to view summary table\n"
+  printf -- "<processname> -debug			to debug process\n"
+  printf -- "start all -print                   	to view all default startup lines\n"
+  printf -- "start <processname(s)> -print		to view default startup lines\n"
+  printf -- "start all		           	to start all processes\n"
```
jonathonmcmurray: looks like incorrect alignment on this line

JuliaPartridge: If I correctly align this in the script it doesn't align when the usage statement prints to screen.

jonathonmcmurray: Hmm, seems strange, are tabs being used, or spaces? (it should be spaces) 

If everything lines up when it's run, that's fine then

ThomasSmyth: There's tabs in there, you can see if you highlight over the whitespace.

setenv.sh
```diff
@@ -1,14 +1,31 @@
...
-export KDBCODE=${TORQHOME}/code
-export KDBLOG=${TORQHOME}/logs
+export TORQHOME=${PWD}                                                                              # if running the kdb+tick example, change these to full paths
+export TORQDATA=${PWD}                                                                              # some of the kdb+tick processes will change directory, and these will no longer be valid
+export TORQAPPHOME=${PWD}
+export KDBLOG=${TORQDATA}/logs
 export KDBHTML=${TORQHOME}/html
 export KDBLIB=${TORQHOME}/lib
+export KDBAPPCODE=${TORQAPPHOME}/code
+export KDBCONFIG=${TORQAPPHOME}/config
```
jonnypress: This should be TORQHOME

And we should also have 

KDBAPPCONFIG=${TORQAPPHOME}/config

setenv.sh
```diff
@@ -1,14 +1,31 @@
...
+export KDBCODE=${TORQAPPHOME}/code
```
jonnypress: should be TORQHOME

Also can you like up the KDBCODE, KDBCONFIG and KDBAPPCODE and KDBAPPCONFIG environemnt variables so they sit together

torq.sh
```diff
@@ -0,0 +1,263 @@
...
+usage() {
+  printf -- "Arguments:\n"
+  printf -- "  start all                                to start all processes\n"
+  printf -- "  start <processname(s)>                   to start process(es)\n"
+  printf -- "  stop all                                 to stop all processes\n"
+  printf -- "  stop <processname(s)>                    to stop process(es)\n"
+  printf -- "  start all -print                         to view all default startup lines\n"
+  printf -- "  start <processname(s)> -print            to view default startup lines\n"
+  printf -- "  -procs                                   to list all processes\n"
+  printf -- "  -summary                                 to view summary table\n"
```
jonathonmcmurray: seems to me like `-procs` and `-summary` shouldn't have a hyphen in front of them? Seems more consistent with the `start` & `stop` commands IMO

torq.sh
```diff
@@ -0,0 +1,263 @@
...
+ }
+
+usage() {
+  printf -- "Arguments:\n"
+  printf -- "  start all                                to start all processes\n"
+  printf -- "  start <processname(s)>                   to start process(es)\n"
+  printf -- "  stop all                                 to stop all processes\n"
+  printf -- "  stop <processname(s)>                    to stop process(es)\n"
+  printf -- "  start all -print                         to view all default startup lines\n"
+  printf -- "  start <processname(s)> -print            to view default startup lines\n"
```
jonathonmcmurray: Stick `-print` in the "optional flags section and get rid of 216 & 217

torq.sh
```diff
@@ -0,0 +1,263 @@
...
+  for p in $PROCS;
+  do
+    stop $p;                                                                                        # kill each process in variable 
+  done
+ }
+
+usage() {
+  printf -- "Arguments:\n"
+  printf -- "  start all                                to start all processes\n"
+  printf -- "  start <processname(s)>                   to start process(es)\n"
```
jonathonmcmurray: condense 212 and 213 into something like `start all|<processname(s)>`, this seems kinda redundant as-is. Same with the `stop` commands on 215 & 216

docs/gettingstarted.md
```diff
@@ -173,6 +219,20 @@ procname were both passed in as command line parameters. If both of
...
+  |---------------------------------| ----------------------------|
+  |U|                                  Authentication|
+  |localtime|                          localtime or UTC|
+  |g|                                  Garbage collection|
+  |T|                                  Timeout|
+  |w|                                  Memory limit|
+  |load|                               Path to load process file|
+  |startwithall|                       Start using torq.sh| 
+  |extras|                             Add arguments to start up line|
+  |qcmd|                               Command to start q session|
```
jonnypress: WE need a bit more detail on each of these.  e.g. 

start with all is to determnine if this process gets started when all is specified
qcmd allows different versions of q to be used, or different command line options e.g. rlwrap, numactl
U/g/T/w are the standard q commnd line args
extras allows you to specify any addition parameters

docs/gettingstarted.md
```diff
@@ -93,6 +93,11 @@ qualification e.g. -.servers.HOPENTIMEOUT 5000.
 
 <a name="env"></a>
 
+Using torq.sh
+---------------------
+torq.sh is a script that runs processes in torq with added functionality, one key enhancement is all the process configuration is now in one place. The default process file is located in $KDBCONFIG/process.csv. This script is only available on Linux. It requires environment variables to be set, similar to torq.q. A usage statement for the script can be seen by running the following in a unix environment: `./torq.sh`.
```
ThomasSmyth: Need to wrap these lines at 80 chars.

docs/gettingstarted.md
```diff
@@ -173,6 +178,68 @@ procname were both passed in as command line parameters. If both of
...
+  |---------------------------------| ----------------------------|
+  |U|                                  Authentication requiring a usr:pwd file|
+  |localtime|                          Sets process running in local time rather than GMT|
+  |g|                                  Garbage collection immediate (1) or deferred (0)|
+  |T|                                  Timeout in seconds for client queries, 0 for no timeout|
+  |w|                                  Workspace MB limit|
+  |load|                               Files or database directory to load|
+  |startwithall|                       Determine if process is started when all is specified| 
+  |extras|                             Specify any additional parameters|
+  |qcmd|                               Allows different versions of q to be used or different command line options - rlwap, numactl|
```
ThomasSmyth: Line up this table correctly, so its easy to view for anyone editing this file in future.

jonathonmcmurray:
All clear from me once those two things @ThomasSmyth pointed out are sorted

PR #95, opened 2018-05-30
Startup dependent function

lholmes8:
Add general function to check that a list of required process connections are made, else process is blocked while re-connection attempted

### DIFF
--- a/code/handlers/trackservers.q
+++ b/code/handlers/trackservers.q
@@ -335,4 +335,19 @@ pc:{[result;W] update w:0Ni,endp:.proc.cp[] from`.servers.SERVERS where w=W;clea
 if[enabled;
 	.z.pc:{.servers.pc[x y;y]}.z.pc;
 	if[DISCOVERYRETRY > 0; .timer.repeat[.proc.cp[];0Wp;DISCOVERYRETRY;(`.servers.retrydiscovery;`);"Attempt reconnections to the discovery service"]];
-	if[RETRY > 0; .timer.repeat[.proc.cp[];0Wp;RETRY;(`.servers.retry;`);"Attempt reconnections to closed server handles"]]];
+    if[RETRY > 0; .timer.repeat[.proc.cp[];0Wp;RETRY;(`.servers.retry;`);"Attempt reconnections to closed server handles"]]];
+
+// Check if required processes all connected
+reqprocsnotconn:{[requiredprocs] 
+    (count requiredprocs)=sum requiredprocs in exec u from .clients.clients
+  }
+
+// Check all required processes are connected before starting process
+startupdependent:{[requiredprocs;timeintv]
+    while[.servers.reqprocsnotconn[requiredprocs];
+              .os.sleep[timeintv];
+              .servers.startup;
+              subscribe[];
+              ]
+  }
+
--- a/code/handlers/trackservers.q
+++ b/code/handlers/trackservers.q
@@ -9,25 +9,25 @@ if[not @[value;`.proc.loaded;0b]; '"environment is not initialised correctly to 
 
 SERVERS:@[value;`.servers.SERVERS;([]procname:`symbol$();proctype:`symbol$();hpup:`symbol$();w:`int$();hits:`int$();startp:`timestamp$();lastp:`timestamp$();endp:`timestamp$();attributes:())]
 
-enabled:@[value;`enabled;1b]                                        // whether server tracking is enabled
-CONNECTIONS:@[value;`CONNECTIONS;`]                             // the list of connections to make at start up
-DISCOVERYREGISTER:@[value;`DISCOVERYREGISTER;1b]                        // whether to register with the discovery service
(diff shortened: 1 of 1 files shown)


code/handlers/trackservers.q
```diff
@@ -335,4 +335,19 @@ pc:{[result;W] update w:0Ni,endp:.proc.cp[] from`.servers.SERVERS where w=W;clea
 if[enabled;
 	.z.pc:{.servers.pc[x y;y]}.z.pc;
 	if[DISCOVERYRETRY > 0; .timer.repeat[.proc.cp[];0Wp;DISCOVERYRETRY;(`.servers.retrydiscovery;`);"Attempt reconnections to the discovery service"]];
-	if[RETRY > 0; .timer.repeat[.proc.cp[];0Wp;RETRY;(`.servers.retry;`);"Attempt reconnections to closed server handles"]]];
+    if[RETRY > 0; .timer.repeat[.proc.cp[];0Wp;RETRY;(`.servers.retry;`);"Attempt reconnections to closed server handles"]]];
+
+// Check if required processes all connected
+reqprocsnotconn:{[requiredprocs] 
+    (count requiredprocs)<>sum requiredprocs in exec proctype from .servers.SERVERS
```
jonnypress: Think this could just be 
not all requiredprocs in exec proctype from ... 
Also, you need to check if the handle is valid.

code/handlers/trackservers.q
```diff
@@ -335,4 +335,19 @@ pc:{[result;W] update w:0Ni,endp:.proc.cp[] from`.servers.SERVERS where w=W;clea
...
+  }
+
+// Block process until all required processes are connected
+startupdependent:{[requiredprocs;timeintv]
+    while[.servers.reqprocsnotconn[requiredprocs];
+              .os.sleep[timeintv];
+              .servers.startup[];
+              subscribe[]
```
jonnypress: You don't need subscribe here I think

code/handlers/trackservers.q
```diff
@@ -335,4 +335,19 @@ pc:{[result;W] update w:0Ni,endp:.proc.cp[] from`.servers.SERVERS where w=W;clea
 if[enabled;
 	.z.pc:{.servers.pc[x y;y]}.z.pc;
 	if[DISCOVERYRETRY > 0; .timer.repeat[.proc.cp[];0Wp;DISCOVERYRETRY;(`.servers.retrydiscovery;`);"Attempt reconnections to the discovery service"]];
-	if[RETRY > 0; .timer.repeat[.proc.cp[];0Wp;RETRY;(`.servers.retry;`);"Attempt reconnections to closed server handles"]]];
+    if[RETRY > 0; .timer.repeat[.proc.cp[];0Wp;RETRY;(`.servers.retry;`);"Attempt reconnections to closed server handles"]]];
```
ThomasSmyth: Keep indentation consistent with above, may need to use tabs or changes the others to spaces.

code/handlers/trackservers.q
lholmes8: That is much better, thanks

code/handlers/trackservers.q
```diff
@@ -243,96 +243,109 @@ procupdate:{[procs] addprocs[procs;exec distinct proctype from procs;0b];}
...
-	.z.pc:{.servers.pc[x y;y]}.z.pc;
-	if[DISCOVERYRETRY > 0; .timer.repeat[.proc.cp[];0Wp;DISCOVERYRETRY;(`.servers.retrydiscovery;`);"Attempt reconnections to the discovery service"]];
-	if[RETRY > 0; .timer.repeat[.proc.cp[];0Wp;RETRY;(`.servers.retry;`);"Attempt reconnections to closed server handles"]]];
+    .z.pc:{.servers.pc[x y;y]}.z.pc;
+    if[DISCOVERYRETRY > 0; .timer.repeat[.proc.cp[];0Wp;DISCOVERYRETRY;(`.servers.retrydiscovery;`);"Attempt reconnections to the discovery service"]];
+    if[RETRY > 0; .timer.repeat[.proc.cp[];0Wp;RETRY;(`.servers.retry;`);"Attempt reconnections to closed server handles"]]];
+
+// Check if required processes all connected
+reqprocsnotconn:{[requiredprocs] 
+    not all requiredprocs in exec proctype from .servers.SERVERS where w in key .z.W
```
jonnypress: could you use .dotz.liveh here to be consistent e.g. 

where .dotz.liveh[w] 

I think would work

code/handlers/trackservers.q
```diff
@@ -243,96 +243,109 @@ procupdate:{[procs] addprocs[procs;exec distinct proctype from procs;0b];}
...
+  }
+
+// Block process until all required processes are connected
+startupdependent:{[requiredprocs;timeintv]
```
jonnypress: Could you move this and previous function to be defined just after start up (before pc)

PR #96, opened 2018-05-30
Add a parentproctype flag for torq.q

mkelly95:
Add the ability to specify a parentproctype for a process when starting using torq.

If specified then the specified parent process configuration will be loaded after default.q but before the process type.

Then the parent process code in [parentproctype] subdirectory of the various code directories will be loaded after default and before the specific process

### DIFF
--- a/code/handlers/trackservers.q
+++ b/code/handlers/trackservers.q
@@ -41,7 +41,7 @@ loadpassword:{
            .lg.o[`conn;"password file ",(string file)," not found"];
            [.lg.o[`conn;"password file ",(string file)," found"];
             .servers.USERPASS:first`$read0 hsym file]]};
-    files:{.proc.getconfig["passwords/",(string x),".txt";2]} each `default,.proc.parentproctype,.proc.proctype,.proc.procname;
+	files:{.proc.getconfig["passwords/",(string x),".txt";2]} each `default,.proc.parentproctype,.proc.proctype,.proc.procname;
 	loadpassfile each (distinct files[;1],files[;0]) except `;
 	}
 loadpassword[]
--- a/code/handlers/trackservers.q
+++ b/code/handlers/trackservers.q
@@ -41,7 +41,7 @@ loadpassword:{
            .lg.o[`conn;"password file ",(string file)," not found"];
            [.lg.o[`conn;"password file ",(string file)," found"];
             .servers.USERPASS:first`$read0 hsym file]]};
-	files:{.proc.getconfig["passwords/",(string x),".txt";2]} each `default,.proc.proctype,.proc.procname;
+	files:{.proc.getconfig["passwords/",(string x),".txt";2]} each `default,.proc.parentproctype,.proc.proctype,.proc.procname;
 	loadpassfile each (distinct files[;1],files[;0]) except `;
 	}
 loadpassword[]
--- a/code/handlers/trackservers.q
+++ b/code/handlers/trackservers.q
@@ -41,7 +41,7 @@ loadpassword:{
            .lg.o[`conn;"password file ",(string file)," not found"];
            [.lg.o[`conn;"password file ",(string file)," found"];
             .servers.USERPASS:first`$read0 hsym file]]};
-	files:{.proc.getconfig["passwords/",(string x),".txt";2]} each `default,.proc.parentproctype,.proc.proctype,.proc.procname;
+    files:{.proc.getconfig["passwords/",(string x),".txt";2]} each `default,.proc.parentproctype,.proc.proctype,.proc.procname;
 	loadpassfile each (distinct files[;1],files[;0]) except `;
 	}
 loadpassword[]
--- a/docs/gettingstarted.md
+++ b/docs/gettingstarted.md
@@ -54,6 +54,10 @@ parameters. An example is below.
     $ . setenv.sh
     $ q torq.q -debug -proctype testproc -procname test1 
 
+To specify the parent process type, do:
+
+    $ q torq.q -debug -parentproctype -testparentproc -proctype testproc -procname test1
+
 To load a file, do:
 
     $ q torq.q -load myfile.q -debug -proctype testproc -procname test1
@@ -70,6 +74,7 @@ The available command line parameters are:
   |Cmd Line Param|            Description|
   |-------------------------| --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
   |-procname x -proctype y   |The process name and process type|
+  |-parentproctype x         |The parent process type|
   |-procfile x               |The name of the file to get the process information from|
   |-load x \[y..z\]          |The files or database directory to load|
   |-loaddir x \[y..z\]       |Load all .q, .k files in specified directories|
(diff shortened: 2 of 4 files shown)


docs/gettingstarted.md
```diff
@@ -54,6 +54,10 @@ parameters. An example is below.
     $ . setenv.sh
     $ q torq.q -debug -proctype testproc -procname test1 
 
+To specify the parent process type, do:
+
+    $ q torq.q -debug -parentproctype -testparentproc -proctype testproc -procname test1
```
jonathonmcmurray: I think `-testparentproc` should be `testparentproc`

torq.q
```diff
@@ -536,37 +546,40 @@ reloadnamecode:{
...
 	// load TorQ Default configuration module
-	.proc.loadconfig[getenv[`KDBCONFIG],"/settings/";] each `default,.proc.proctype,.proc.procname;
+	.proc.loadconfig[getenv[`KDBCONFIG],"/settings/";] each `default,.proc.parentproctype,.proc.proctype,.proc.procname;
   // check if KDBSERVCONFIG is set and load Service Layer specific configuration module
   $[""~getenv`KDBSERVCONFIG;
     .lg.o[`fileload;"environment variable KDBSERVCONFIG not set, not loading app specific config"];
     [.proc.servconfig:getenv[`KDBAPPCONFIG],"/settings/";
     .lg.o[`fileload;"environment variable KDBSERVCONFIG set, loading app specific config from ",.proc.servconfig];
-    .proc.loadconfig[.proc.servconfig;] each `default,.proc.proctype,.proc.procname]
+	.proc.loadconfig[.proc.servconfig;] each `default,.proc.parentproctype,.proc.proctype,.proc.procname]
```
jonathonmcmurray: looks like indentation was increased here, I think it should be the same as before

code/handlers/trackservers.q
```diff
@@ -41,7 +41,7 @@ loadpassword:{
            .lg.o[`conn;"password file ",(string file)," not found"];
            [.lg.o[`conn;"password file ",(string file)," found"];
             .servers.USERPASS:first`$read0 hsym file]]};
-	files:{.proc.getconfig["passwords/",(string x),".txt";2]} each `default,.proc.proctype,.proc.procname;
+	files:{.proc.getconfig["passwords/",(string x),".txt";2]} each `default,.proc.parentproctype,.proc.proctype,.proc.procname;
```
ThomasSmyth: Could go for `` `default,.proc`parentproctype`proctype`procname``

torq.q
```diff
@@ -521,6 +527,10 @@ reloadcommoncode:{
 	// Load common code from each directory if it exists
 	loadspeccode["/common"]'[`KDBCODE`KDBSERVCODE`KDBAPPCODE];
 	};
+reloadparentprocesscode:{
+  // Load parentproctype code from each directory if it exists
+  loadspeccode["/",string parentproctype]'[`KDBCODE`KDBSERVCODE`KDBAPPCODE];
+  };
```
ThomasSmyth: Indentation of this block is inconsistent with those around it.

torq.q
```diff
@@ -536,37 +546,40 @@ reloadnamecode:{
...
-    .proc.loadconfig[.proc.servconfig;] each `default,.proc.proctype,.proc.procname]
+	.proc.loadconfig[.proc.servconfig;] each `default,.proc.parentproctype,.proc.proctype,.proc.procname]
   ];
 	// check if KDBAPPCONFIG is set and load Appliation specific configuration module 
   $[""~getenv`KDBAPPCONFIG;	
 	  .lg.o[`fileload;"environment variable KDBAPPCONFIG not set, not loading app specific config"];
 	  [.proc.appconfig:getenv[`KDBAPPCONFIG],"/settings/";
 	  .lg.o[`fileload;"environment variable KDBAPPCONFIG set, loading app specific config from ",.proc.appconfig];
-	  .proc.loadconfig[.proc.appconfig;] each `default,.proc.proctype,.proc.procname]
+	  .proc.loadconfig[.proc.appconfig;] each `default,.proc.parentproctype,.proc.proctype,.proc.procname]
```
ThomasSmyth: Again, could use `` `default,.proc`parentproctype`proctype`procname`` on 549, 555 and 562

docs/gettingstarted.md
```diff
@@ -128,19 +133,24 @@ Process Identification
...
-determine the code base and configuration loaded, and how they are
-connected to by other processes. If both of these are not defined, the
-TorQ will attempt to use the port number a process was started on to
-determine the code base and configuration loaded.
+defined by two required variables - .proc.proctype and .proc.procname
+which are the type and name of the process respectively and an optional
+variable parentproctype which is the type of the parent process. These
+three values determine the code base and configuration loaded, and how
+they are connected to by other processes. If both of the required
+variables are not defined, the TorQ will attempt to use the port number
```
jonnypress: Get rid of “the” before TorQ

docs/gettingstarted.md
```diff
@@ -128,19 +133,24 @@ Process Identification
...
 
 At the crux of AquaQ TorQ is how processes identify themselves. This is
-defined by two variables - .proc.proctype and .proc.procname which are
-the type and name of the process respectively. These two values
-determine the code base and configuration loaded, and how they are
-connected to by other processes. If both of these are not defined, the
-TorQ will attempt to use the port number a process was started on to
-determine the code base and configuration loaded.
+defined by two required variables - .proc.proctype and .proc.procname
+which are the type and name of the process respectively and an optional
```
jonnypress: Can you make this a new sentence (after respectively). Something like “An optional variable parentproctype allows an initial codebase and configuration to be loaded”

torq.q
```diff
@@ -536,37 +546,40 @@ reloadnamecode:{
...
   ];
 	// Override config from the command line
 	.proc.override[]]
 
 // Load library code 
 .proc.loadcommoncode:@[value;`.proc.loadcommoncode;1b];
+.proc.loadparentprocesscode:@[value;`.proc.loadparentprocesscode;1b];
```
jonnypress: I think we need to simplify this a bit. We don’t need additional variables for loadparentprocesscode- just drive them from loadprocesscode. So if parent proctype is defined and loadprocesscode is true, then load the parent process code as well. So you can then drop the reloadparentprocesscode method.

torq.q
mkelly95: When parentproctype isn't supplied it is set as () so they way I did it means that it wont appear in the list as an empty list is joined on

torq.q
mkelly95: By mistake used a tab

torq.q
```diff
@@ -344,6 +344,12 @@ $[count[req] = count req inter key params;
 	.lg.o[`init;"ignoring partial subset of required process parameters found on the command line - reading from file"];
   ()];		 
 
+// If parentproctype has been supplied then set it
+parentproctype:`$();
```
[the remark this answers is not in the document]
mkelly95: It could be, in which case I would have to change
``` not null first .proc.parentproctype;.proc.reloadparentprocesscode[]```
to
```not null first `$.proc.parentproctype;.proc.reloadparentprocesscode[]```

as applying null on a general empty list gives a generic null

But I think the cast is neater inside the function instead of on assignment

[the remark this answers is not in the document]
jonnypress: Good point. 
Just leave as is.  It's not that important

PR #99, opened 2018-06-26
Error in non-torq process

Aqualex:
Added checkNonTorq function in trackservers.q which checks if .proc.getattributes is defined in the non torq process


code/handlers/trackservers.q
```diff
@@ -175,12 +173,15 @@ retrydiscovery:{
 // Called by the discovery service when it restarts
 autodiscovery:{if[DISCOVERYRETRY>0; .servers.retrydiscovery[]]}
 
+//check if .proc.getattributes exists in the nontorqprocess
+checkNonTorq:{$[(`proc in key`)&(first distinct`getattributes`cd in key`.proc);1;0]};
```
jonathonmcmurray: no other function here appears to use camelCase, so for consistency (and because it's awful) please avoid

code/handlers/trackservers.q
```diff
@@ -34,28 +34,26 @@ PASSWORDS:@[value;`PASSWORDS;enlist[`]!enlist `]
 // Otherwise just load the usernames and passwords from the passwords directory
 // using the usual hierarchic approach
 loadpassword:{
+
```
ThomasSmyth: Remove this line.

code/handlers/trackservers.q
```diff
@@ -34,28 +34,26 @@ PASSWORDS:@[value;`PASSWORDS;enlist[`]!enlist `]
 // Otherwise just load the usernames and passwords from the passwords directory
 // using the usual hierarchic approach
 loadpassword:{
+    
     .lg.o[`conn;"attempting to load external connection username:password from file"];
     // load a password file
     loadpassfile:{[file]
-         $[()~key hsym file;
+         $[()~key hsym file;
```
ThomasSmyth: Remove the added space.

code/handlers/trackservers.q
```diff
@@ -137,8 +135,8 @@ addh:{[hpuP]
     addhw[hpuP;W]]}
 
 // return the details of the current process
-getdetails:{(.z.f;.z.h;system"p";@[value;`.proc.procname;`];@[value;`.proc.proctype;`];@[value;(`.proc.getattributes;`);()!()])}
-
+getdetails:{(.z.f;.z.h;system"p";@[value;`.proc.procname;`];@[value;`.proc.proctype;`];@[value;(`.proc.ges;`);()!()])}
+//getdetails:{(.z.f;.z.h;system"p";@[value;`.proc.procname;`];@[value;`.proc.proctype;`])}
```
ThomasSmyth: Remove comment.

code/handlers/trackservers.q
```diff
@@ -175,12 +173,15 @@ retrydiscovery:{
 // Called by the discovery service when it restarts
 autodiscovery:{if[DISCOVERYRETRY>0; .servers.retrydiscovery[]]}
 
+//check if .proc.getattributes exists in the nontorqprocess
+checknontorqattr:{$[(`proc in key`)&(first distinct`getattributes`cd in key`.proc);1;0]};
```
ThomasSmyth: I don't get the second condition here ``first distinct`getattributes`cd in key`.proc``, won't `first` return the same as `first distinct`? And if you're running `first` then why not just check for `` `getattributes`` only as you'll never get to `` `cd``. Also why is this a conditional? Should just return from the condition directly unless it needs to be a long.

code/handlers/trackservers.q
```diff
@@ -175,12 +173,15 @@ retrydiscovery:{
...
+
 // Attempt to make a connection for specified row ids
 retryrows:{[rows]
     // opencon, amends global tables, cannot be used inside of a select statement
     handles:.servers.opencon each exec hpup from`.servers.SERVERS where i in rows;
-    update lastp:.proc.cp[],w:handles from`.servers.SERVERS where i in rows;
-        update attributes:{$[null x;()!();@[x;(`.proc.getattributes;`);()!()]]} each w,startp:?[null w;0Np;.proc.cp[]] from `.servers.SERVERS where i in rows;
+    update lastp:.proc.cp[],w:handles from`.servers.SERVERS where i in rows;
```
ThomasSmyth: Remove added space.

code/handlers/trackservers.q
jonathonmcmurray: even if it does need to be a long, could just cast it & still remove conditional

code/handlers/trackservers.q
```diff
@@ -175,12 +173,15 @@ retrydiscovery:{
...
+checknontorqattr:{$[(`proc in key`)&(first distinct`getattributes`cd in key`.proc);1;0]};
+
 // Attempt to make a connection for specified row ids
 retryrows:{[rows]
     // opencon, amends global tables, cannot be used inside of a select statement
     handles:.servers.opencon each exec hpup from`.servers.SERVERS where i in rows;
-    update lastp:.proc.cp[],w:handles from`.servers.SERVERS where i in rows;
-        update attributes:{$[null x;()!();@[x;(`.proc.getattributes;`);()!()]]} each w,startp:?[null w;0Np;.proc.cp[]] from `.servers.SERVERS where i in rows;
+    update lastp:.proc.cp[],w:handles from`.servers.SERVERS where i in rows; 
+        update attributes:{$[null x;()!();$[@[x;(checknontorqattr;())];@[x;(`.proc.getattributes;`);()!()];()!()]]} each w,startp:?[null w;0Np;.proc.cp[]] from `.servers.SERVERS where i in rows;
```
jonathonmcmurray: this line is a little unwieldy... nested `$` conditional can probably be simplified a bit?

At the very least, can get rid of inner `$` and use the 'if...else if...else' `$[cond1;val1;cond2;val2;val3]` style

code/handlers/trackservers.q
Aqualex: That was not one of my changes, but I will remove it.

ThomasSmyth: You added it in this commit https://github.com/AquaQAnalytics/TorQ/commit/7f48c88246a459581982241b82dc46027633b9b3

code/handlers/trackservers.q
Aqualex: yeah. I have seen now that it basically returns the same thing. I will remove the conditional form as well.

code/handlers/trackservers.q
Aqualex: Will change it now.

code/handlers/trackservers.q
```diff
@@ -175,13 +173,16 @@ retrydiscovery:{
...
 retryrows:{[rows]
     // opencon, amends global tables, cannot be used inside of a select statement
     handles:.servers.opencon each exec hpup from`.servers.SERVERS where i in rows;
-    update lastp:.proc.cp[],w:handles from`.servers.SERVERS where i in rows;
-        update attributes:{$[null x;()!();@[x;(`.proc.getattributes;`);()!()]]} each w,startp:?[null w;0Np;.proc.cp[]] from `.servers.SERVERS where i in rows;
-        if[ count connectedrows:select from `.servers.SERVERS where i in rows, .dotz.liveh0 w;
+    update lastp:.proc.cp[],w:handles from`.servers.SERVERS where i in rows; 
+    update attributes:{$[null x;()!();@[x;(checknontorqattr;())];@[x;(`.proc.getattributes;`);()!()];()!()]} each w,startp:?[null w;0Np;.proc.cp[]] from `.servers.SERVERS where i in rows;
+    if[ count connectedrows:select from `.servers.SERVERS where i in rows, .dotz.liveh0 w;
     connectcustom[connectedrows]]}
```
ThomasSmyth: Not touched by you but this action needs indented for some clarity :+1:

code/handlers/trackservers.q
```diff
@@ -175,13 +173,16 @@ retrydiscovery:{
 // Called by the discovery service when it restarts
 autodiscovery:{if[DISCOVERYRETRY>0; .servers.retrydiscovery[]]}
 
+//check if .proc.getattributes exists in the nontorqprocess
+checknontorqattr:{(`proc in key`)&(`getattributes in key`.proc)};
```
jonathonmcmurray: Is the first condition even necessary?

```q
q).proc
'.proc
  [0]  .proc
       ^
q)key`.proc
q)`getattributes in key`.proc
0b
```
Is there a condition under which this is insufficient?

Also, even if it is necessary, parentheses around second condition can be removed

code/handlers/trackservers.q
```diff
@@ -175,13 +173,16 @@ retrydiscovery:{
...
+
 // Attempt to make a connection for specified row ids
 retryrows:{[rows]
     // opencon, amends global tables, cannot be used inside of a select statement
     handles:.servers.opencon each exec hpup from`.servers.SERVERS where i in rows;
-    update lastp:.proc.cp[],w:handles from`.servers.SERVERS where i in rows;
-        update attributes:{$[null x;()!();@[x;(`.proc.getattributes;`);()!()]]} each w,startp:?[null w;0Np;.proc.cp[]] from `.servers.SERVERS where i in rows;
-        if[ count connectedrows:select from `.servers.SERVERS where i in rows, .dotz.liveh0 w;
+    update lastp:.proc.cp[],w:handles from`.servers.SERVERS where i in rows; 
+    update attributes:{$[null x;()!();@[x;(checknontorqattr;())];@[x;(`.proc.getattributes;`);()!()];()!()]} each w,startp:?[null w;0Np;.proc.cp[]] from `.servers.SERVERS where i in rows;
```
jonathonmcmurray: Haven't tested this to be sure, but I think this can be further simplified something like

```q
$[not[null x]&@[x;(checknontorqattr;`);0b];@[x;(`.proc.getattributes;`);()!()];()!()]
```

This lambda is possibly approaching the length where it could be split out into it's own function? Could be more nicely laid out that way to be more readable I think

code/handlers/trackservers.q
```diff
@@ -175,13 +173,21 @@ retrydiscovery:{
 // Called by the discovery service when it restarts
 autodiscovery:{if[DISCOVERYRETRY>0; .servers.retrydiscovery[]]}
 
+//check if .proc.getattributes exists in the nontorqprocess
+checknontorqattr:{(`proc in key`)&`getattributes in key`.proc};
```
jonathonmcmurray: Repasting "outdated" comment that still applies:

Is the first condition even necessary?
```q
q).proc
'.proc
  [0]  .proc
       ^
q)key`.proc
q)`getattributes in key`.proc
0b
```
Is there a condition under which this is insufficient?

code/handlers/trackservers.q
```diff
@@ -175,13 +173,21 @@ retrydiscovery:{
...
+
 // Attempt to make a connection for specified row ids
 retryrows:{[rows]
+    //function a checks if the handle passed is empty and also invokes checknontorqattr function 
+    //which checks if .proc.getattributes is defined on the nontorqprocess and executes it 
+    //only if it is defined
+    a:{$[(first not null x)&@[first x;(checknontorqattr;`);0b];@[first x;(`.proc.getattributes;`);()!()];()!()]};
```
jonathonmcmurray: I think this should maybe be a standalone function for checking attributes on a remote process

I thnk the error trap in first clause of `$` should also remove the need to check if `x` is null - if it is, error trap will catch it anyway.

Also if you're only passing one handle in, as the call below appears to do, should be no need for all these `first`s

Finally, according to our internal style suggestions, it is preferred to control execution order with square brackets rather than parentheses i.e.
```q
first[not null x]
```
rather than
```q
(first not null x)
```
(but as above, I think this part can actually be removed completely)

code/handlers/trackservers.q
Aqualex: I just forgot to remove it.

code/handlers/trackservers.q
jonnypress: Yes remove commented out getdetails definition

code/handlers/trackservers.q
```diff
@@ -177,11 +173,16 @@ autodiscovery:{if[DISCOVERYRETRY>0; .servers.retrydiscovery[]]}
 
 // Attempt to make a connection for specified row ids
 retryrows:{[rows]
+    //function a checks if the handle passed is empty and also invokes checknontorqattr function 
+    //which checks if .proc.getattributes is defined on the nontorqprocess and executes it 
+    //only if it is defined
+    a:{$[(first not null x)&@[first x;"`getattributes in key`.proc";0b];:@[first x;(`.proc.getattributes;`);()!()];()!()]};
```
jonnypress: Why do you have the "first"s in here?  Are there cases where x is passed in as a list? 

This current version is making two calls to the remote process - can you just make it one call?  So a lambda which checks if getattributes exists and then calls it if it does?

ThomasSmyth:
Make sure you run `dos2unix` on any file you change here, I've purged windows line endings from the codebase.

code/handlers/trackservers.q
```diff
@@ -42,23 +42,19 @@ loadpassword:{
            [.lg.o[`conn;"password file ",(string file)," found"];
             .servers.USERPASS:first`$read0 hsym file]]};
     files:{.proc.getconfig["passwords/",(string x),".txt";2]} each `default,.proc.parentproctype,.proc.proctype,.proc.procname;
-    loadpassfile each (distinct files[;1],files[;0]) except `;
+    loadpassfile each (distinct files[;1],files[;0]) except`;
```
ThomasSmyth: Is `raze files[;1 0]` the same thing?

code/handlers/trackservers.q
```diff
@@ -125,7 +121,7 @@ addnthawc:{[name;proctype;hpup;attributes;W;checkhandle]
     if[checkhandle and not isalive:.dotz.liveh W;'"invalid handle"];
     cleanup[];
     $[not hpup in (exec hpup from .servers.SERVERS) inter (exec hpup from .servers.nontorqprocesstab);
-        `.servers.SERVERS insert(name;proctype;lower hpup;W;0i;$[isalive;.proc.cp[];0Np];.proc.cp[];0Np;attributes);
+        [`.servers.SERVERS insert(name;proctype;lower hpup;W;0i;$[isalive;.proc.cp[];0Np];.proc.cp[];0Np;attributes);.servers.SERVERS];
```
ThomasSmyth: Would ``value`.servers.SERVERS insert...`` avoid the need for this code block and returning the servers separately?

code/handlers/trackservers.q
```diff
@@ -137,13 +133,14 @@ addh:{[hpuP]
     addhw[hpuP;W]]}
 
 // return the details of the current process
+// getdetails:{(.z.f;.z.h;system"p";@[value;`.proc.procname;`];@[value;`.proc.proctype;`];@[value;(`.proc.getattributes;`);()!()])}
```
ThomasSmyth: Commented duplicate of line below?

code/handlers/trackservers.q
```diff
@@ -137,13 +133,14 @@ addh:{[hpuP]
...
 getdetails:{(.z.f;.z.h;system"p";@[value;`.proc.procname;`];@[value;`.proc.proctype;`];@[value;(`.proc.getattributes;`);()!()])}
-
 / add session behind a handle
 addhw:{[hpuP;W]
     // Get the information around a process
     / info:`f`h`port`procname`proctype`attributes!(@[W;"(.z.f;.z.h;system\"p\";@[value;`.proc.procname;`];@[value;`.proc.proctype;`];@[value;(`.proc.getattributes;`);()!()])";(`;`;0Ni;`;`;()!())]);
-    info:`f`h`port`procname`proctype`attributes!(@[W;(`.servers.getdetails;`);(`;`;0Ni;`;`;()!())]);
+    //info:`f`h`port`procname`proctype`attributes!(@[W;(`.servers.getdetails;`);(`;`;0Ni;`;`;()!())]);
```
ThomasSmyth: Looks like you don't need this anymore.

code/handlers/trackservers.q
```diff
@@ -137,13 +133,14 @@ addh:{[hpuP]
...
+    info:`f`h`port`procname`proctype`attributes!(@[W;"$[`getdetails in key`.proc;.proc.getdetails[];(.z.f;.z.h;system\"p\";`;`;$[`getattributes in key`.proc;.proc.getattributes[];()!()])]";(`;`;0Ni;`;`;()!())]);
```
ThomasSmyth: Instead of checking for existence would ``@[.proc.getdetails;`;<alt operation>]`` work instead?

code/handlers/trackservers.q
```diff
@@ -177,11 +174,16 @@ autodiscovery:{if[DISCOVERYRETRY>0; .servers.retrydiscovery[]]}
 
 // Attempt to make a connection for specified row ids
 retryrows:{[rows]
+    //function a checks if the handle passed is empty and also invokes checknontorqattr function 
+    //which checks if .proc.getattributes is defined on the nontorqprocess and executes it 
+    //only if it is defined
+    a:{$[(not null x);@[x;"$[`getattributes in key`.proc;.proc.getattributes[];()!()]";()!()];()!()]};
```
ThomasSmyth: Same comment as line 143 may apply here + don't need the brackets round `not null x`.

code/handlers/trackservers.q
Aqualex: They are the same but I will have to add something to sort the result so that `TorQ` files will be loaded before `TorQ-Finance` files

If you do `files[;1 0]`  you get:
```
/home/arisnoveanu/TorQProdSupp/TorQDev/TorQ/config/passwords/default.txt /home/arisnoveanu/TorQProdSupp/TorQDev/TorQ-Finance-Starter-Pack/appconfig/passwords/default.txt
/home/arisnoveanu/TorQProdSupp/TorQDev/TorQ/config/passwords/rdb.txt     /home/arisnoveanu/TorQProdSupp/TorQDev/TorQ-Finance-Starter-Pack/appconfig/passwords/rdb.txt
/home/arisnoveanu/TorQProdSupp/TorQDev/TorQ/config/passwords/rdb1.txt    /home/arisnoveanu/TorQProdSupp/TorQDev/TorQ-Finance-Starter-Pack/appconfig/passwords/rdb1.txt
```

and if you raze it: 

```
`/home/arisnoveanu/TorQProdSupp/TorQDev/TorQ/config/passwords/default.txt`/home/arisnoveanu/TorQProdSupp/TorQDev/TorQ-Finance-Starter-Pack/appconfig/passwords/default.txt`/home/arisnoveanu/TorQProdSupp/TorQDev/TorQ/config/passwords/rdb.txt`/home/arisnoveanu/TorQProdSupp/TorQDev/TorQ-Finance-Starter-Pack/appconfig/passwords/rdb.txt`/home/arisnoveanu/TorQProdSupp/TorQDev/TorQ/config/passwords/rdb1.txt`/home/arisnoveanu/TorQProdSupp/TorQDev/TorQ-Finance-Starter-Pack/appconfig/passwords/rdb1.txt
```

The order of the files is different and because these files have to be checked in order I will need to sort the list such that the `TorQ` files will be loaded before the `TorQ-Finance` files. 

I am not sure if it is an easy way to sort that list. I tried a few things but in my opinion it makes the code slightly more complicated and confusing then it is right now.

jonathonmcmurray: what if you do `raze flip files[;1 0]` ?

ThomasSmyth: Try `flip` then `raze`?

code/handlers/trackservers.q
Aqualex: I am trying to avoid errors being raised in the nontorq process. from what I know the apply function will still try to execute the function.

ThomasSmyth: Okay, so what I said earlier wouldn't work at all, but if you wanted a way to change that then this might work:
```
@[value;`.proc.getattributes;{::}]()!()
```

code/handlers/trackservers.q
```diff
@@ -133,14 +133,12 @@ addh:{[hpuP]
...
 // return the details of the current process
-// getdetails:{(.z.f;.z.h;system"p";@[value;`.proc.procname;`];@[value;`.proc.proctype;`];@[value;(`.proc.getattributes;`);()!()])}
 getdetails:{(.z.f;.z.h;system"p";@[value;`.proc.procname;`];@[value;`.proc.proctype;`];@[value;(`.proc.getattributes;`);()!()])}
 / add session behind a handle
 addhw:{[hpuP;W]
     // Get the information around a process
-    / info:`f`h`port`procname`proctype`attributes!(@[W;"(.z.f;.z.h;system\"p\";@[value;`.proc.procname;`];@[value;`.proc.proctype;`];@[value;(`.proc.getattributes;`);()!()])";(`;`;0Ni;`;`;()!())]);
-    //info:`f`h`port`procname`proctype`attributes!(@[W;(`.servers.getdetails;`);(`;`;0Ni;`;`;()!())]);
-    info:`f`h`port`procname`proctype`attributes!(@[W;"$[`getdetails in key`.proc;.proc.getdetails[];(.z.f;.z.h;system\"p\";`;`;$[`getattributes in key`.proc;.proc.getattributes[];()!()])]";(`;`;0Ni;`;`;()!())]);
+    break;
```
Aqualex: I'll remove this ...

code/handlers/trackservers.q
```diff
@@ -227,7 +226,8 @@ addprocs:{[connectiontab;procs;connect]
     // we've dropped some items - maybe there are updated attributes
     if[not count[res]=count connectiontab;
         if[`attributes in cols connectiontab;
-            .servers.SERVERS:.servers.SERVERS lj 3!select procname,proctype,hpup,attributes from connectiontab]];
+            a:select hpup,attributes from .servers.SERVERS;
```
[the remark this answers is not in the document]
Aqualex: That's strange. I remember I removed that a when I changed line 230. I assume the changes were not updated through WinScp.

code/handlers/trackservers.q
ThomasSmyth: Got it. Why do you only want to return `.servers.SERVERS` if this is true? Looks like value doesn't get used as this function outputs `W`.

Aqualex: I don't think there's a particular reason for that.  It was in torq from before. It doesn't seem to be used anywhere else. Probably it was used in the past and nobody changed it when they decided to output just the handle from the function. I think in can be removed from the code.

ThomasSmyth: It's not in master, the change was made in this branch.

PR #102, opened 2018-06-28
Fix process variable to get all processes

JuliaPartridge:
This PR will fix the problem of not being able to access processes with `startwithall` set to 0 in `process.csv`.  

I have made the following changes:

- `getall`  is a function that gets all processes with `startwithall` set to 1. I have replaced some of these with `awk -F, '{if(NR>1) print $4}' $CSVPATH` which will get all the processes. I have only changed this for the `PROCS` variables that are used to check the process exists. 
- the debug flag was echoing the wrong start line before evaluating the correct line. I have modified this so that it doesn't echo the line it doesn't use.
- small change to the port variable in the summary function. Additional `-n` flag to `netstat` to get port numbers where there were occasionally names.

### DIFF
--- a/torq.sh
+++ b/torq.sh
@@ -3,110 +3,110 @@
 . ./setenv.sh                                                                                       # load the environment
 
 getfield() {
-  fieldno=`awk -F, '{if(NR==1) for(i=1;i<=NF;i++){if($i=="'$2'") print i}}' $CSVPATH`               # get number for field based on headers
-  fieldval=`awk -F, '{if(NR == '$1') print $'$fieldno'}' $CSVPATH`                                  # pull one field from one line of file
-  echo $fieldval | envsubst                                                                         # substitute env vars
+  fieldno=$(awk -F, '{if(NR==1) for(i=1;i<=NF;i++){if($i=="'$2'") print i}}' "$CSVPATH")            # get number for field based on headers
+  fieldval=$(awk -F, '{if(NR == '$1') print $'$fieldno'}' "$CSVPATH")                               # pull one field from one line of file
+  echo "$fieldval" | envsubst                                                                       # substitute env vars
  }
 
 parameter() {
-  fieldval=`getfield $1 $2`
+  fieldval=$(getfield "$1" "$2")
   if [[ "" == "$fieldval" ]]; then                                                                  # check for empty string
     echo ""
   else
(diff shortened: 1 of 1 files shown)


torq.sh
```diff
@@ -69,7 +68,7 @@ summary() {
     printf "%-8s | %-14s | %-6s |\n" `date '+%H:%M:%S'` "$1" "down"                                 # summary table row for down process
   else
     pid=$(findproc $1)
-    port=`netstat -pl 2>/dev/null | grep $pid | awk '{ print $4 }' | head -1 | cut -c 3-`           # get port process is running on 
+    port=`netstat -nlp 2>/dev/null | grep $pid | awk '{ print $4 }' | head -1 | awk -F: '{ print $2 }'`
```
jonathonmcmurray: Can you put the comment back on this line? it'll have to be slightly out of line with the others given this is a longer line, but that's better than no comment

torq.sh
```diff
@@ -241,15 +240,15 @@ elif [[ "$1" == "stop" ]]; then
   stopprocs $PROCS;
 elif [[ "$1" == "summary" ]]; then
   allcsv $@;
-  PROCS=$(getall);
+  PROCS=`awk -F, '{if(NR>1) print $4}' $CSVPATH`;
```
jonathonmcmurray: Seems like this global is set a bunch of different places under different conditions, but all in the same way - could this just be set once when the script starts & used wherever it's needed?

JuliaPartridge:
After running `shellcheck torq.sh`, I have gone through the script and changed the majority of the lines it was telling me to update. There were a few where the suggestions stopped the script from working so I left those. 

With setting a global variable for ``` PROCS=`awk -F, '{if(NR>1) print $4}' $CSVPATH` ```, I was having difficulty setting this variable in the script. It either wouldn't run or give me an error such as `./torq.sh: line 228: -F,: command not found`

ThomasSmyth:
Make sure you run `dos2unix` on any file you change here, I've purged windows line endings from the codebase.

jonathonmcmurray:
Apologies I forgot about this PR for a bit, looks good! I assume you have tested all the different functionality being provided (starting all processes, starting some processes, starting a process in debug mode etc.)? Assuming that has been done, I'm going to approve, and you can merge

PR #105, opened 2018-07-17
[WIP] Integrating Monit for monitoring TorQ

Aqualex:
This will add the basic configuration needed to use `monit` to monitor TorQ processes on UNIX-like systems. The processes will be read from the process.csv file, started and monitored by monit. The dashboard can be accessed on localhost port 2812 and will include basic information about the TorQ processes. If any of the processes fail to start, monit will try to restart the process and send an alert e-mail which contains information about the service.


code/scripts/templates/monitrc
```diff
@@ -0,0 +1,321 @@
...
+## Set the list of mail servers for alert delivery. Multiple servers may be
+## specified using a comma separator. If the first mail server fails, Monit
+# will use the second mail server in the list and so on. By default Monit uses
+# port 25 - it is possible to override this with the PORT option.
+#
+# set mailserver mail.bar.baz,               # primary mailserver
+#                backup.bar.baz port 10025,  # backup mailserver on port 10025
+#                localhost                   # fallback relay
+  set mailserver smtp.gmail.com port 587
+        username "\"torqmonit@gmail.com\"" password "\"torqmonit2018\""
```
jonathonmcmurray: is this a real email address & password? these should be dummy placeholders in here, and this password is now in git history so should also be changed

code/scripts/templates/monitrc
```diff
@@ -0,0 +1,321 @@
...
+#
+## Monit has an embedded HTTP interface which can be used to view status of
+## services monitored and manage services from a web interface. The HTTP
+## interface is also required if you want to issue Monit commands from the
+## command line, such as 'monit status' or 'monit restart service' The reason
+## for this is that the Monit client uses the HTTP interface to send these
+## commands to a running Monit daemon. See the Monit Wiki if you want to
+## enable SSL for the HTTP interface.
+#
+ set httpd port 2812 and
```
jonathonmcmurray: anyway to have this port based on TorQ baseport variable? what happens if a user already has something running on port 2812?

code/scripts/templates/testmail.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+echo -e "TORQMONIT\r\rThis is an automated message\r\rProcess SERVICE is down"|mail.mailutils -s \"SUBJECT\" rdanutalexandrugmail.com
```
jonathonmcmurray: I assume this file should be removed? Doesn't seem like something that should be merged to master

setenv.sh
```diff
@@ -12,8 +12,10 @@ export KDBAPPCODE=${TORQAPPHOME}/code
 export KDBHDB=${TORQDATA}/hdb/database
 export KDBWDB=${TORQDATA}/wdbhdb
 
-export KDBBASEPORT=6000                                                                             # set KDBBASEPORT to the default value for a TorQ Installation
+export KDBBASEPORT=6000                                                                            # set KDBBASEPORT to the default value for a TorQ Installation
```
jonathonmcmurray: looks like alignment changed on this comment bringing it out of line, please put it back

fill_templates.sh
```diff
@@ -0,0 +1,59 @@
...
+}
+
+#SETTING DEFAULT ENVIRONMENT VARIABLES #############################################################
+if [ "-bash" = $0 ]; then
+    dirpath="${BASH_SOURCE[0]}"
+else
+    dirpath="$0"
+fi
+                            
+eval ". $(dirname "$dirpath")/setenv.sh"                                                            #set environment variables
```
jonathonmcmurray: why is `eval` needed here? Why not just source the file with `.` directly? (this pattern exists in a few places)

code/scripts/templates/monitrc
Aqualex: It is an e-mail address that I will change. I only use this to test if monit works as it should. I wasn't able to make the e-mail work yet.

jonathonmcmurray: Yeah, but regardless you should never commit a password to any email account to git - right now anyone could log into that email account (and I did, to check)

code/scripts/templates/monitrc
Aqualex: this is the default port for monit. It wasn't my focus to change it, but it should be really easy to do so. monitrc is created using the environment variables which are set when `setenv.sh` is executed. Should I look at a way to set that port taking in consideration the `baseport` and the number of processes or should I maybe define a an env variable and let the user set the port (just like we do now with the `baseport`)

code/scripts/templates/testmail.sh
Aqualex: This is my attempt to use an external e-mail to generate the alerts. Not very successful so far.

setenv.sh
Aqualex: will do

fill_templates.sh
Aqualex: This was already here. I think eval is used to execute `". $(dirname "$dirpath")/setenv.sh"`

jonathonmcmurray: Not sure what you mean by "already here", this is a new file, no? I get what `eval` is doing, I'm asking _why_ though? Seems like you could just do `. $(dirname "$dirpath")/setenv.sh`

code/scripts/templates/monitrc
jonathonmcmurray: personally I think the less a user has to think about configuring the better, so tying it to the baseport seems most logical to me

code/scripts/templates/monitconfig.cfg
```diff
@@ -0,0 +1 @@
+${TORQHOME}/appconfig/process.csv     ${TORQHOME}/torq.sh     monitconfig.cfg
```
ThomasSmyth: This + `fill_templates.sh` was set up by me a long time ago rather naively, there's probably better ways to do it and definitely better ways to set up this config file, would be worth looking into it.

Aqualex:
bash monit.sh takes me in the root directory and tries to execute everything from there. This creates a number of issues. I have seen someone using `start program = "/bin/bash -c 'cd location && ./torq.sh $procname'"`. At the moment I am defining TORQHOME in a file called .bashrc_torq and source it so I can use the full path. Anyone knows a better way for doing solving this problem?

monit.sh
```diff
@@ -0,0 +1,67 @@
...
+  checkst "$HOME/.bashrc_torq" "Creating .bashrc_torq" "exist"
+  echo "export TORQHOME=${PWD}" >> $HOME/.bashrc_torq
+  echo "export BASEDIR=${PWD}" >> $HOME/.bashrc_torq
+  checkst "cat $HOME/.bashrc_torq | grep 'TORQHOME\|BASEDIR'|wc -l" "Environment variables are set" "npres" "2"
+  echo "source $HOME/.bashrc_torq" >> $HOME/.bashrc
+  source $HOME/.bashrc
+  checkst "echo $TORQHOME|wc -l" "TORQHOME present" "npres" "1"
+  checkst "echo $BASEDIR|wc -l" "BASEDIR present" "npres" "1"
+else
+  echo "export TORQHOME=${PWD}" >> $HOME/.bashrc_torq
```
jonathonmcmurray: Does all this stuff need to be duplicated? Can you not just move it out of the `if` statement & run it regardless? Or is there a difference I'm missing?

jonathonmcmurray:
@Aqualex I don't think I understand the problem...why can't you just source `setenv.sh` the same way the TorQ start script does and get `TORQHOME` defined that way? Seems really messy to write a second script and then source it, plus the name `.bashrc_torq` doesn't really make sense; `.bashrc` is a file loaded by bash at the start of a shell session, that's not really related to this. I don't get why it's necessary to write a second script, if the purpose is to set environment variables, can you just do that directly if required? Why write into a second script & source it instead of just doing it directly?

config/settings/gateway.q
```diff
@@ -11,5 +11,5 @@ clearinactivetime:0D01:00	// the time to keep inactive handle data
 
 // Server connection details
 \d .servers
-CONNECTIONS:`rdb`hdb		// list of connections to make at start up
+CONNECTIONS:`rdb`hdb`wdb		// list of connections to make at start up
```
jonnypress: why is this here?

fill_templates.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
```
jonnypress: I'd rather this script didn't sit at the top level directory of TorQ

torq.sh
```diff
@@ -221,8 +221,14 @@ usage() {
   exit 1
```
jonnypress: Why do you need to change torq.sh in this PR?

setenv.sh
```diff
@@ -1,6 +1,12 @@
-export TORQHOME=${PWD}                                                                              # if running the kdb+tick example, change these to full paths
-export TORQDATA=${PWD}                                                                              # some of the kdb+tick processes will change directory, and these will no longer be valid
-export TORQAPPHOME=${PWD}
+if [ "-bash" = $0 ]; then
```
jonnypress: Are these changes strictly for this PR?

monit/bin/monit.sh
```diff
@@ -0,0 +1,162 @@
+#!/bin/bash
+#FUNTION DECLARATION ###############################################################################
```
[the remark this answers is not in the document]
Aqualex: fixed

monit/templates/monitalert.cfg
```diff
@@ -0,0 +1,12 @@
+check system homer.aquaq.co.uk
```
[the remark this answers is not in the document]
Aqualex: done

monit/templates/monitrc
```diff
@@ -0,0 +1,321 @@
+###############################################################################
```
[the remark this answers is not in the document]
Aqualex: so you want me to remove this altogether and everything the bit in monit.sh that creates it? I believe we should have monit setup in such way that it has some basic functionality straigth out of the box so that our users will see that it works at least.

torq.sh
Aqualex: monit executes torq.sh from the root directory. Torq.sh uses `. setenv.sh` and obviously the `setenv.sh` is not in the root directory

setenv.sh
Aqualex: yes

jonnypress:
As per conversation, add some documentation and we are good to go.  We can start with basic documentation and flesh it out.

PR #107, opened 2018-07-27
Add variable to store executed functions in .proc.initlist

JuliaPartridge:
This PR will close the following issue in Redmine: http://smithers.aquaq.co.uk/issues/2099

The functions executed when .proc.init is invoked will be stored in the variable .proc.initexecuted.


torq.q
```diff
@@ -616,7 +617,7 @@ if[@[value;`.servers.STARTUP;0b]; .servers.startup[]]
 		[{[a].lg.o[`init;"attemping to run initialisation: ",-3!a];
```
ThomasSmyth: I think instead of if-else this would be better as an exit early check if `.proc.initlist` is empty. Something like:
```q
if[0=count .proc.initlist;:.lg.o[`init;"no initialisation functions found"]];
{[a].lg.o....
```
Avoids the need for a code block. Thoughts @jonnypress and @jonathonmcmurray?

jonathonmcmurray: I agree this is much neater. Only caveat is that the '`$` conditional with code blocks' pattern is used in a lot of places in TorQ, so this may be inconsistent, but I think refactoring this way is more readable & my opinion would be that takes precedence.

torq.q
```diff
@@ -612,12 +613,11 @@ if[@[value;`.servers.STARTUP;0b]; .servers.startup[]]
...
-		[{[a].lg.o[`init;"attemping to run initialisation: ",-3!a];
-		@[value;a;
-		{[x;a].lg.e[`init;x," error - failed to run initialisation: ",-3!a]}[;a]]}
-		each .proc.initlist;.proc.initlist:()];
-		.lg.o[`init;"no initialisation functions found"]];
+    if[0=count .proc.initlist;:.lg.o[`init;"no initialisation functions found"]];
+    {[a].lg.o[`init;"attemping to run initialisation: ",-3!a];
+        @[value;a;
+        {[x;a].lg.e[`init;x," error - failed to run initialisation: ",-3!a]}[;a]]}
+        each .proc.initlist;.proc.initexecuted,:.proc.initlist;.proc.initlist:();
```
jonathonmcmurray: I personally think this is more clear if the closing `}` for the lambda is on this line before `each`, and the `.proc.initexecuted,:.proc.initlist;.proc.initlist:();` is broken up over the next few lines

Thoughts @ThomasSmyth ?

torq.q
```diff
@@ -612,12 +613,11 @@ if[@[value;`.servers.STARTUP;0b]; .servers.startup[]]
...
 // function to execute functions in .proc.initlist
 .proc.init:{
-	$[count .proc.initlist;
-		[{[a].lg.o[`init;"attemping to run initialisation: ",-3!a];
-		@[value;a;
-		{[x;a].lg.e[`init;x," error - failed to run initialisation: ",-3!a]}[;a]]}
-		each .proc.initlist;.proc.initlist:()];
-		.lg.o[`init;"no initialisation functions found"]];
+    if[0=count .proc.initlist;:.lg.o[`init;"no initialisation functions found"]];
+    {[a].lg.o[`init;"attemping to run initialisation: ",-3!a];
```
jonathonmcmurray: Should avoid internal functions where possible, use `.Q.s` instead of `-3!` (or maybe just `string`?)

torq.q
ThomasSmyth: I think this needs cleaned up a bit so it's easier to see where the lambda starts + stops:
```
{[a]
  .log.o...
  @[value;a;{[x;a].lg.e...
}each .proc.initlist;
.proc.initexecuted,:.proc.initlist
.proc.initlist:();
```
Thoughts? @jonnypress + @jonathonmcmurray

jonathonmcmurray: perhaps this multiline lambda could be made into it's own function `.proc.try` or something like that? so then in this function we would do `.proc.try each .proc.initlist`

I think even with good formatting, multiline lambdas can be confusing and hard to read, best to avoid them wherever we can

Maybe we can make it more generic as well, have it accept (in this case) `` `init``, and remove the word "initialisation " from the log messages; then this could in theory be used in other places if necessary. And here it would be ``.proc.try[`init] each .proc.initlist``

torq.q
```diff
@@ -612,12 +613,11 @@ if[@[value;`.servers.STARTUP;0b]; .servers.startup[]]
...
+        @[value;a;
+        {[x;a].lg.e[`init;x," error - failed to run initialisation: ",-3!a]}[;a]]}
```
ThomasSmyth: Given we have complete control of this function I think the args should be rearranged for simplicity and to remove some squares:
```
{[a;x].lg.e[`init;x," error - failed to run initialisation: ",-3!a]}a
```

torq.q
JuliaPartridge: Reason why I'm using `-3!` here is because I get a type error with `string` and `.Q.s`  adds an `\n`:

```q)).lg.o[`init;"attemping to run initialisation: ",string a]
'type
  [7]  /home/jpartridge/torq/deploy/torq.q:188: .lg.format:{[loglevel;proctype;proc;id;message] "|" sv (string .proc.cp[];string .z.h;string proctype;string proc;string loglevel;string id;message)}
                                                                                                    ^
q.lg)))\
q)).lg.o[`init;"attemping to run initialisation: ",-3!a]
2018.07.31D15:12:41.462410000|homer.aquaq.co.uk|discovery|discovery1|INF|init|attemping to run initialisation: "func[2]"
q)).lg.o[`init;"attemping to run initialisation: ",.Q.s a]
2018.07.31D15:12:50.649729000|homer.aquaq.co.uk|discovery|discovery1|INF|init|attemping to run initialisation: "func[2]"

```

jonathonmcmurray: ah was thinking `a` was a symbol, guess not. I think maybe `.Q.s1` is the direct equivalent of `-3!` but I suppose that's getting into even more internal stuff. Suppose `-3!` is maybe fine in this case

ThomasSmyth: `.Q.s1` is the function for `-3!`.

torq.q
```diff
@@ -610,14 +611,18 @@ if[@[value;`.ps.loaded;0b]; .ps.initialise[]]
 // initialise connections
 if[@[value;`.servers.STARTUP;0b]; .servers.startup[]]
 
+.proc.run:{[id;a]
+    .lg.o[id;"attemping to run: ",.Q.s1 a];
+    @[value;a;
+    {[id;x;a].lg.e[id;x," error - failed to run: ",.Q.s1 a]}[id;;a]];
```
ThomasSmyth: Combine 616+617 and change 617 to `{[id;a;x]...}[id;a]` for simplicity.

torq.q
```diff
@@ -610,14 +611,18 @@ if[@[value;`.ps.loaded;0b]; .ps.initialise[]]
...
 // function to execute functions in .proc.initlist
 .proc.init:{
-	$[count .proc.initlist;
-		[{[a].lg.o[`init;"attemping to run initialisation: ",-3!a];
-		@[value;a;
-		{[x;a].lg.e[`init;x," error - failed to run initialisation: ",-3!a]}[;a]]}
-		each .proc.initlist;.proc.initlist:()];
-		.lg.o[`init;"no initialisation functions found"]];
+    if[0=count .proc.initlist;:.lg.o[`init;"no initialisation functions found"]];
+    .proc.run[`INIT;] each .proc.initlist;
```
ThomasSmyth: Should be lower case `init` + can slim this down:
```
.proc.run[`init]each .proc.initlist;
```
OR
```
.proc.run[`init]'[.proc.initlist];
```

torq.q
jonathonmcmurray: similar to prior comment from @ThomasSmyth I would rearrange args here to `[id;a;x]` so you can remove a `;` in `[id;;a]` after lambda

torq.q
```diff
@@ -610,14 +611,18 @@ if[@[value;`.ps.loaded;0b]; .ps.initialise[]]
 // initialise connections
 if[@[value;`.servers.STARTUP;0b]; .servers.startup[]]
 
+.proc.run:{[id;a]
+    .lg.o[id;"attemping to run: ",.Q.s1 a];
```
jonathonmcmurray: think the style guideline is 2 spaces of indentation within functions (same in `.proc.init`)

torq.q
```diff
@@ -610,14 +611,18 @@ if[@[value;`.ps.loaded;0b]; .ps.initialise[]]
 // initialise connections
 if[@[value;`.servers.STARTUP;0b]; .servers.startup[]]
 
+.proc.run:{[id;a]
```
jonathonmcmurray: I prefer `.proc.try` given we are using a try-catch / error trap, but maybe `run` is more explanatory

torq.q
jonathonmcmurray: @ThomasSmyth what's the ruling here in regards to maintaining consistency?

torq.q
```diff
@@ -610,14 +611,17 @@ if[@[value;`.ps.loaded;0b]; .ps.initialise[]]
...
 // function to execute functions in .proc.initlist
 .proc.init:{
-	$[count .proc.initlist;
-		[{[a].lg.o[`init;"attemping to run initialisation: ",-3!a];
-		@[value;a;
-		{[x;a].lg.e[`init;x," error - failed to run initialisation: ",-3!a]}[;a]]}
-		each .proc.initlist;.proc.initlist:()];
-		.lg.o[`init;"no initialisation functions found"]];
+  if[0=count .proc.initlist;:.lg.o[`init;"no initialisation functions found"]];
+  .proc.try[`init;]each .proc.initlist;
```
jonathonmcmurray: one last thing... don't need the `;` after `` `init``

PR #108, opened 2018-07-27
Exit when setenv not sourced

JuliaPartridge:
This PR will close the following issue on Redmine: http://smithers.aquaq.co.uk/issues/2101

`torq.sh` will exit if setenv is not sourced and no longer hang.


torq.sh
```diff
@@ -224,6 +224,13 @@ usage() {
  }
 
 
+if [ -f $SETENV ]; then                                                                             # check script exists 
+  . ./$SETENV                                                                                       # load the environment
```
jonathonmcmurray: this still requires `$SETENV` to be located in current working directory, I think it'd be more generic if the default value of `SETENV` was something like `$(pwd)/setenv.sh` and this line was just `. $SETENV` to allow people to use a different location if they want - they'd maybe have to have a more custom `setenv.sh` for that to still work, but we shouldn't exclude the option.

torq.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash 
 
-. ./setenv.sh                                                                                       # load the environment
+SETENV=setenv.sh                                                                                    # set the environment
```
jonathonmcmurray: maybe this should check for an env var `SETENV` first & if it's set use that value? So someone could in theory do something like

```
$ SETENV=/home/user/setenv.sh start.sh all
```

or whatever to start up using a custom located env file without having to modify the start script. Obviously default would still be `setenv.sh` within current dir

torq.sh
```diff
@@ -255,7 +262,7 @@ elif [[ "$2" == "-debug" ]]; then
 elif [[ "$#" -eq 0 ]]; then                                                                             
   usage
 else
-  echo "Invalid argument(s)"
+  echo "ERROR - Invalid argument(s)"
```
jonathonmcmurray: Line 230 has `ERROR: `, this line has `ERROR - `. Can you pick one and use it for all error messages?

torq.sh
JuliaPartridge: This works fine when I move `setenv.sh` out of the current working directory and provide the path from the working directory to the file. However, your suggestion is a better idea.

torq.sh
```diff
@@ -223,6 +221,18 @@ usage() {
   exit 1
  }
 
+if [[ $(printenv | grep SETENV) ]]; then
```
jonathonmcmurray: I think this can be simplified, something like:

```bash
jmcmurray@homer ~ $ if [[ -z "$SETENV" ]]; then echo "blank"; else echo "not blank"; fi
blank
jmcmurray@homer ~ $ SETENV=test
jmcmurray@homer ~ $ if [[ -z "$SETENV" ]]; then echo "blank"; else echo "not blank"; fi
not blank
```

torq.sh
```diff
@@ -223,6 +221,18 @@ usage() {
...
+  SETENV=$(printenv | grep SETENV | sed -e 's#.*=\(\)#\1#')                                         # set predefined environment
```
jonathonmcmurray: why is this needed? If it's already set, it's all good. I think this should be a case of setting it if it's not set, otherwise leaving it e.g.

```bash
jmcmurray@homer ~ $ echo $SETENV  # not set

jmcmurray@homer ~ $ if [[ -z "$SETENV" ]]; then SETENV="test"; fi
jmcmurray@homer ~ $ echo $SETENV
test
jmcmurray@homer ~ $ SETENV=example #pre-set value
jmcmurray@homer ~ $ if [[ -z "$SETENV" ]]; then SETENV="test"; fi
jmcmurray@homer ~ $ echo $SETENV
example
```

Alternatively you may be able to use a default value expansion, e.g.

```bash
jmcmurray@homer ~ $ echo $SETENV
example
jmcmurray@homer ~ $ echo ${SETENV:-test}
example
jmcmurray@homer ~ $ unset SETENV
jmcmurray@homer ~ $ echo ${SETENV:-test}
test
```

torq.sh
JuliaPartridge: Problem is when I run the following in the script with SETENV in the env var it returns blank. This is why I have done it in the way that I have.

JuliaPartridge:
So the setenv variable can be set using `export SETENV=${PWD}` as we are running the script using `./torq.sh` which starts a sub process. `SETENV=${PWD}` will fail to set the environment variable and running the script using `. ./torq.sh` exits the current session.

When trying out `${SETENV:-${PWD}/setenv.sh}` I got a permissions error. So I settled for the if statement .

PR #110, opened 2018-08-13
Update print and debug arguments 

JuliaPartridge:
This PR will update how to run `print` and `debug` using `torq.sh`. 

Debug can now be run using the following:
`./torq.sh debug <proc>`

Print can now be run using the following:
`./torq.sh print <proc(s)>`

Previously the flags `-csv` and `-extras` did not work with debug, this has been fixed and additional error traps have been added.

### DIFF
--- a/torq.sh
+++ b/torq.sh
@@ -25,8 +25,7 @@ startline() {
   proctype=$(getfield "$procno" "proctype")                                                         # get proctype for process 
   params="U localtime g T w load"                                                                   # list of params to read from config 
   sline="${TORQHOME}/torq.q -stackid ${KDBBASEPORT} -proctype $proctype -procname $1"               # base part of startup line
-  for p in $params;                                                                                 # iterate over params
-  do
+  for p in $params; do                                                                              # iterate over params
     a=$(parameter "$procno" "$p");                                                                  # get param
     sline="$sline$a";                                                                               # append to startup line
   done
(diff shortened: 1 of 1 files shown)


torq.sh
```diff
@@ -234,19 +233,27 @@ else
...
 elif [[ "$1" == "stop" ]]; then
   checkextrascsv $@;
   stopprocs "$PROCS";
+elif [[ "$1" == "debug" ]]; then
+  checkextrascsv "$*";
+  if [[ $(echo $PROCS | wc -w) -gt 1 ]]; then 
+    echo "ERROR: Cannot debug more than one process at a time"
+  else 
+    for p in $PROCS;
+    do
```
ThomasSmyth: I think typical bash style is to put `do` on the same line as `for` (similar to `then`).

torq.sh
```diff
@@ -234,19 +233,27 @@ else
...
-    done
-  else 
-    checkextrascsv "$*";
-    startprocs "$PROCS";
-  fi
+  checkextrascsv "$*";
+  startprocs "$PROCS";
+elif [[ "$1" == "print" ]]; then         
+  checkextrascsv "$*";
+  for p in $PROCS;
```
ThomasSmyth: This do something for all procs is often repeated, something that could be turned into a function? @jonnypress @jonathonmcmurray

jonathonmcmurray:
I agree that a case statement would be a bit neater, but don't think it matters too much; if it'll be quick to change that, I'd say go ahead @JuliaPartridge

torq.sh
jonathonmcmurray: I'm not sure how easy that is, my bash-fu isn't the strongest...can you pass a function as an argument to another function? I think that'd be required here, essentially pass the list and the function to run on list. Could possibly be done passing the name of the function & using `eval` but `eval` makes me nervous

jonathonmcmurray: Actually, it can be done without `eval`

```bash
jmcmurray@homer ~ $ function f() { echo $1; }
jmcmurray@homer ~ $ function each() { for x in $2; do $1 $x; done; }
jmcmurray@homer ~ $ each f "a b c d"
a
b
c
d
```

ThomasSmyth: Great! Even though I brought this up I think that should be moved to another branch so as not to derail this one.

JuliaPartridge:
Alright, I will go ahead and merge.

PR #120, opened 2018-09-28
Add absolute path to bash scripts

ThomasSmyth:
A few of the bash scripts are using relative paths instead of absolute paths, I've updated 3 of them for a start to ensure absolute paths are used. I've also made some minor modifications to the layout of the bash scripts in accordance to this [style guide](https://google.github.io/styleguide/shell.xml) (although I'm, sure I haven't got everything it says).

I've rearranged `torq.sh` so the check for missing `setenv.sh` occurs near the start of the file.

It also looks like this monit addition is a little fragile, I'm going to add some comments later on to highlight a couple of the issues I think exist with it.

### DIFF
--- a/monit/bin/monit.sh
+++ b/monit/bin/monit.sh
@@ -1,8 +1,13 @@
 #!/bin/bash
 #FUNCTION DECLARATION ###############################################################################
 
- cd ../.. && export BASEDIR="${PWD}" && cd - > /dev/null;                                           # set BASEDIR to the root of the torq directory
- mkdir -p ../logs  
+if [ "-bash" = $0 ]; then
+  BASEDIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+else
+  BASEDIR="$(cd "$(dirname "$0")" && pwd)"
+fi
+BASEDIR=$(dirname $(dirname $BASEDIR))                                                              # set BASEDIR to root of TorQ directory
+mkdir -p ${BASEDIR}/monit/logs                                                                      # create directory for monit logs
 
 checkst(){
 #function to check if file exists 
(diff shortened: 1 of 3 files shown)


monit/bin/monit.sh
```diff
@@ -1,8 +1,13 @@
...
 #FUNCTION DECLARATION ###############################################################################
 
- cd ../.. && export BASEDIR="${PWD}" && cd - > /dev/null;                                           # set BASEDIR to the root of the torq directory
- mkdir -p ../logs  
+if [ "-bash" = $0 ]; then
+  BASEDIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+else
+  BASEDIR="$(cd "$(dirname "$0")" && pwd)"
+fi
+BASEDIR=$(dirname $(dirname $BASEDIR))                                                              # set BASEDIR to root of TorQ directory
```
ThomasSmyth: This addition is to mimic the `cd ../..` seen previously, but I'm not dying about it.

Should this file just point to `setenv.sh` and have some monit location variable declared there instead of `BASEDIR`?
The template `monit/templates/monitrc` would need to be updated to follow suit. I only mention this as the template file has some commented out references to `TORQMONIT` which is defined in `setenv.sh`

setenv.sh
```diff
@@ -24,4 +26,4 @@ export TORQPROCESSES=${KDBAPPCONFIG}/process.csv
 
 export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$KDBLIB/l32
 
-export TORQMONIT=${TORQHOME}/logs/monit                                                             # set the folder for monit outputs  
+export TORQMONIT=${TORQHOME}/logs/monit                                                             # set the folder for monit outputs
```
ThomasSmyth: This is currently unused and actually refers to a different location to the log location defined in `monit/bin/monit.sh` (which is `monit/logs`).

PR #121, opened 2018-09-28
Improve .proc.init functionality

FionaMorgan:
# Improve .proc.init functionality! :computer:

### PR general description
This pull request adds a function to take command line functions from -initlist and run the new function .proc.initcmd prior to the current .proc.initlist as called by .proc.init[]. It has updated .proc.init and .proc.try to also wipe .proc.initlist, update .init.initexectued and to print success log messages.

### Why is this PR necessary?
To keep track of additional functions being executed.
 
### Final Checklist
- [x] My code follows the code style of this project
- [x] My code is commented appropriately and concisely
- [x] If necessary, I have added new appropriate labels to this PR
- [x] I have assigned the appropriate users to this PR

### DIFF
--- a/torq.q
+++ b/torq.q
@@ -614,26 +614,22 @@ if[@[value;`.servers.STARTUP;0b]; .servers.startup[]]
 // function to execute functions in .proc.initlist
 .proc.try:{[id;a]
   .lg.o[id;"attempting to run: ",.Q.s1 a];
-  success:@[value;a;{[id;a;x].lg.e[id;x," error - failed to run: ",.Q.s1 a];:1b}[id;a]];
-  if[-1h=type success;:()];
+  success:@[{value x;1b};a;{[id;a;x].lg.e[id;x," error - failed to run: ",.Q.s1 a];0b}[id;a]];
+  if[not success;:()];
   .proc.initexecuted,:a;
   .lg.o[id;"run successful: ",.Q.s1 a];
  }
 
 .proc.initcmd:{
   // command line functions executed first
-  $[`initlist in key .proc.params;
-    {.proc.try[`init;(x;`)]}each .proc.params`initlist;
-    .lg.o[`init;"no initialisation functions found in cmd line args"]
-   ];
+  if[not`initlist in key .proc.params;:.lg.o[`init;"no initialisation functions found in cmd line args"]];
+  {.proc.try[`init;(x;`)]}each .proc.params`initlist;
  }
 
 .proc.init:{
   .proc.initcmd[];
-  $[count .proc.initlist;
-    .proc.try[`init]each .proc.initlist;
-    .lg.o[`init;"no initialisation functions found"]
-   ];
+  if[0=count .proc.initlist;:.lg.o[`init;"no initialisation functions found"]];
+  .proc.try[`init]each .proc.initlist;
   .proc.initlist:();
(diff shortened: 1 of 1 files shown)


torq.q
```diff
@@ -611,16 +611,29 @@ if[@[value;`.ps.loaded;0b]; .ps.initialise[]]
...
+   ];
  }
 
-// function to execute functions in .proc.initlist
 .proc.init:{
-  if[0=count .proc.initlist;:.lg.o[`init;"no initialisation functions found"]];
-  .proc.try[`init]each .proc.initlist;
-  .proc.initexecuted,:.proc.initlist;
+  .proc.initcmd[];
+  $[count .proc.initlist;
```
ThomasSmyth: Looks like you've only added line 632 so I think you should revert the rest of the changes to this function.

torq.q
```diff
@@ -611,16 +611,29 @@ if[@[value;`.ps.loaded;0b]; .ps.initialise[]]
...
+  .lg.o[id;"attempting to run: ",.Q.s1 a];
+  success:@[value;a;{[id;a;x].lg.e[id;x," error - failed to run: ",.Q.s1 a];:1b}[id;a]];
+  if[-1h=type success;:()];
+  .proc.initexecuted,:a;
+  .lg.o[id;"run successful: ",.Q.s1 a];
+ }
+
+.proc.initcmd:{
+  // command line functions executed first
+  $[`initlist in key .proc.params;
```
ThomasSmyth: If think you should change this to error out, like `.proc.init` was before your changes:
```
if[not`initlist in key .proc.params;:.lg.o ... ];
{.proc.try[`init;(x;`)]}each .proc.params`initlist;
```

torq.q
```diff
@@ -611,16 +611,29 @@ if[@[value;`.ps.loaded;0b]; .ps.initialise[]]
 // initialise connections
 if[@[value;`.servers.STARTUP;0b]; .servers.startup[]]
 
+// function to execute functions in .proc.initlist
 .proc.try:{[id;a]
-  .lg.o[id;"attemping to run: ",.Q.s1 a];
-  @[value;a;{[id;a;x].lg.e[id;x," error - failed to run: ",.Q.s1 a]}[id;a]];
+  .lg.o[id;"attempting to run: ",.Q.s1 a];
+  success:@[value;a;{[id;a;x].lg.e[id;x," error - failed to run: ",.Q.s1 a];:1b}[id;a]];
```
ThomasSmyth: I think success should be a Bool like so:
```
@[{value x;1b};a;{[id;a;x].lg.e[id;x," error - failed to run: ",.Q.s1 a];0b}[id;a]];
```
You'll get `1b` if it runs and `0b` if it fails, which would make the next line:
```
if[not success;:()];
```

torq.q
```diff
@@ -611,16 +611,29 @@ if[@[value;`.ps.loaded;0b]; .ps.initialise[]]
...
+  if[-1h=type success;:()];
```
ThomasSmyth: Should you have some logging if not a success?

FionaMorgan: Can you expand upon this? Think that's done in line 617

ThomasSmyth: Got it, no need to worry about this comment, I'll mark as resolved.

PR #128, opened 2018-10-08
Gateway Null Hang

ruelaidler:
There's a bug in the gateway that makes it hang if a request is sent async and returns a null (::) result. Basically, the GW uses null as a placeholder for results and the request then hangs and doesn't get returned to the client. To fix this, it's probably more suitable to replace the null and use a random GUID as the placeholder value.

### DIFF
--- a/code/processes/gateway.q
+++ b/code/processes/gateway.q
@@ -103,7 +103,6 @@ clients:([]time:`timestamp$(); clienth:`g#`int$(); user:`symbol$(); ip:`int$(); 
 //Function to generate random placeholder
 genrand:{system ("S ",string `int$.z.T);-1?0Ng}
 
-
 //Generate random placeholder
 placehold:.gw.genrand[]
 
--- a/code/processes/gateway.q
+++ b/code/processes/gateway.q
@@ -151,7 +151,7 @@ canberun:{
 
 // Manage client queries
 addquerytimeout:{[query;servertype;queryattributes;join;postback;timeout;sync]
-  `.gw.queryqueue upsert (nextqueryid[];.proc.cp[];.z.w;query;servertype;queryattributes;join;postback;timeout;0Np;0Np;0b;0<count queryattributes;sync)
+  `.gw.queryqueue upsert (nextqueryid[];.proc.cp[];.z.w;query;servertype;queryattributes;join;postback;timeout;0Np;0Np;0b;0<count queryattributes;sync);
  };
 removeclienthandle:{
  update submittime:2000.01.01D0^submittime,returntime:2000.01.01D0^returntime from `.gw.queryqueue where clienth=x;
@@ -184,7 +184,7 @@ finishquery:{[qid;err;serverh]
  deleteresult[qid];
  update error:err,returntime:.proc.cp[] from `.gw.queryqueue where queryid in qid;
  setserverstate[serverh;0b];
- }  
+ }
 
 // Get a list of pending and running queries
 getqueue:{select queryid,time,clienth,query,servertype,status:?[null submittime;`pending;`running],submittime from .gw.queryqueue where null returntime}
@@ -212,7 +212,7 @@ addservererror:{[queryid;error]
  }
 // check if all results are in.  If so, send the results to the client
 checkresults:{[queryid]
- if[not any (.gw.placehold)~/: value (r:.gw.results[queryid])[1;;1];
+ if[not any .gw.placehold~/: value (r:.gw.results[queryid])[1;;1];
   // get the rest of the detail from the query table
   querydetails:queryqueue[queryid];
   // apply the join function to the results
@@ -258,7 +258,7 @@ removeserverhandle:{[serverh]
  // get the list of effected query ids
 
  // 1) queries sent to this server but no reply back yet
- qids:where {[res;id] any (.gw.placehold)~/:res[1;where id=res[1;;0];1]}[;serverid] each results;
+ qids:where {[res;id] any .gw.placehold~/:res[1;where id=res[1;;0];1]}[;serverid] each results;
  // propagate an error back to each client
  sendclientreply[;.gw.errorprefix,"backend ",string[servertype]," server handling query closed the connection";0b] each qids;
  finishquery[qids;1b;serverh]; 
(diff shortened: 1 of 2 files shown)


code/processes/gateway.q
```diff
@@ -100,10 +100,16 @@ queryqueue:([queryid:`u#`long$()] time:`timestamp$(); clienth:`g#`int$(); query:
 // client details
 clients:([]time:`timestamp$(); clienth:`g#`int$(); user:`symbol$(); ip:`int$(); host:`symbol$())
 
+//Function to generate random placeholder
+genrand:{system "S";-1?0Ng}
```
ThomasSmyth: Whats the purpose of `system"S"` here? Doesn't seem to do anything.

code/processes/gateway.q
```diff
@@ -178,15 +183,14 @@ getnextquery:{
 finishquery:{[qid;err;serverh] 
  deleteresult[qid];
  update error:err,returntime:.proc.cp[] from `.gw.queryqueue where queryid in qid;
- if[.gw.querykeeptime=0D; .gw.removequeries[.gw.querykeeptime]];
  setserverstate[serverh;0b];
- } 
+ }
```
ThomasSmyth: Remove that 1 extra space please.

code/processes/gateway.q
```diff
@@ -145,9 +151,8 @@ canberun:{
 
 // Manage client queries
 addquerytimeout:{[query;servertype;queryattributes;join;postback;timeout;sync]
-  `.gw.queryqueue upsert (nextqueryid[];.proc.cp[];.z.w;query;servertype;queryattributes;join;postback;timeout;0Np;0Np;0b;0<count queryattributes;sync);
+  `.gw.queryqueue upsert (nextqueryid[];.proc.cp[];.z.w;query;servertype;queryattributes;join;postback;timeout;0Np;0Np;0b;0<count queryattributes;sync)
```
ThomasSmyth: Add semi colon back in, doesn't look this this should return anything.

code/processes/gateway.q
```diff
@@ -100,10 +100,16 @@ queryqueue:([queryid:`u#`long$()] time:`timestamp$(); clienth:`g#`int$(); query:
 // client details
 clients:([]time:`timestamp$(); clienth:`g#`int$(); user:`symbol$(); ip:`int$(); host:`symbol$())
 
+//Function to generate random placeholder
+genrand:{system "S";-1?0Ng}
+
+//Generate random placeholder
+placehold:.gw.genrand[]
```
ThomasSmyth: Should a different random value be returned each time or does it not really matter what the value is?

code/processes/gateway.q
```diff
@@ -208,7 +212,7 @@ addservererror:{[queryid;error]
  }
 // check if all results are in.  If so, send the results to the client
 checkresults:{[queryid]
- if[not any (::)~/: value (r:.gw.results[queryid])[1;;1];
+ if[not any (.gw.placehold)~/: value (r:.gw.results[queryid])[1;;1];
```
ThomasSmyth: Can remove the brackets round `.gw.placehold`.

code/processes/gateway.q
```diff
@@ -254,7 +258,7 @@ removeserverhandle:{[serverh]
  // get the list of effected query ids
 
  // 1) queries sent to this server but no reply back yet
- qids:where {[res;id] any (::)~/:res[1;where id=res[1;;0];1]}[;serverid] each results;
+ qids:where {[res;id] any (.gw.placehold)~/:res[1;where id=res[1;;0];1]}[;serverid] each results;
```
ThomasSmyth: Remove brackets.

code/processes/gateway.q
```diff
@@ -264,7 +268,7 @@ removeserverhandle:{[serverh]
  activeServerTypes:distinct exec servertype from .gw.servers where active, handle<>serverh;
 
  qids2:where {[res;id;aIDs;aTypes] 
-  s:where (::)~/:res[1;;1]; 
+  s:where (.gw.placehold)~/:res[1;;1];
```
ThomasSmyth: Remove brackets.

code/processes/gateway.q
```diff
@@ -503,9 +507,9 @@ syncexecjpre36:{[query;servertype;joinfunction]
  // to allow parallel execution, send an async query up each handle, then block and wait for the results
  (neg handles)@\:({@[neg .z.w;@[{(1b;.z.p;value x)};x;{(0b;.z.p;x)}];{@[neg .z.w;(0b;.z.p;x);()]}]};query);
  // flush
- (neg handles)@\:(::);
+ (neg handles)@\:(.gw.placehold);
```
ThomasSmyth: Remove brackets.

code/processes/gateway.q
```diff
@@ -503,9 +507,9 @@ syncexecjpre36:{[query;servertype;joinfunction]
...
  // block and wait for the results
- res:handles@\:(::);
+ res:handles@\:(.gw.placehold);
```
ThomasSmyth: Remove brackets.

code/processes/gateway.q
ruelaidler: When I spoke to Jonny last, he said that the GUID could just be generated upon start-up. I assume that this method is what he meant, but I can check with him.

ThomasSmyth: That's all good then, I will mark as resolved.

code/processes/gateway.q
ruelaidler: OH, I was trying to reset the seed for the process but forgot to fix this function. Is this necessary or will it reset automatically dependent on time?

ThomasSmyth: The seed defaults to -314159 and I don't think it ever changes unless it's done explicitly.

ruelaidler: Ok, I'll have a look into it. Thanks Thomas!

code/processes/gateway.q
```diff
@@ -100,10 +100,16 @@ queryqueue:([queryid:`u#`long$()] time:`timestamp$(); clienth:`g#`int$(); query:
 // client details
 clients:([]time:`timestamp$(); clienth:`g#`int$(); user:`symbol$(); ip:`int$(); host:`symbol$())
 
+//Function to generate random placeholder
+genrand:{system ("S ",string `int$.z.T);-1?0Ng}
```
jonathonmcmurray: don't need the brackets here

ruelaidler: cool, will take out now

jonnypress: It doesn't hugely matter given you are always using ~ for comparison, but could you change this to rand 0Ng? -1?0Ng will generate a 1 item list, an atom would be more appropriate.

ruelaidler: Understood. I'll fix this now!

ruelaidler: The function doesn't work when I take out the space after system?

PR #129, opened 2018-10-08
edit debug arg check to allow extra flags

FionaMorgan:
The argument constraint in the function rundebug in torq.sh stops from flags being able to use. Without this argument if two (or more) arguments are given to the debug flag and one of them is a process while the others aren't then the process that exists successfully goes into debug mode.

Have replaced the argument constraint with a count of the specific debug flag args for now so additional flags can be used


jonnypress:
Thomas can you check this one please?

PR #130, opened 2018-10-12
WDB reloadend function call

Joe-Griffiths:
At the moment the WDB process sends reloadend[] as a string to the gateway. This PR alters this functionality to send it as a functional call, this includes a slight change to the logging to prevent formatting issues.

### DIFF
--- a/code/processes/wdb.q
+++ b/code/processes/wdb.q
@@ -226,7 +226,7 @@ d:()!()
 doreload:{[pt]
 	.wdb.reloadcomplete:0b;
 	/-inform gateway of reload start
-	informgateway["reloadstart[]"];
+	informgateway[(`reloadstart;[])];
 	getprocs[;pt] each reloadorder;
 	if[eodwaittime>0;
 		.timer.one[.wdb.timeouttime:.proc.cp[]+.wdb.eodwaittime;(value;".wdb.flushend[]");"release all hdbs and rdbs as timer has expired";0b];
--- a/code/processes/wdb.q
+++ b/code/processes/wdb.q
@@ -226,7 +226,7 @@ d:()!()
 doreload:{[pt]
 	.wdb.reloadcomplete:0b;
 	/-inform gateway of reload start
-	informgateway[(`reloadstart;[])];
+	informgateway(`reloadstart;);
 	getprocs[;pt] each reloadorder;
 	if[eodwaittime>0;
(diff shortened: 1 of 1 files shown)


code/processes/wdb.q
```diff
@@ -226,7 +226,7 @@ d:()!()
 doreload:{[pt]
 	.wdb.reloadcomplete:0b;
 	/-inform gateway of reload start
-	informgateway["reloadstart[]"];
+	informgateway[(`reloadstart;[])];
```
jonnypress: I'm not sure if [] works here, but it's not the standard way to do it.  Can you change to

(`reloadstart;`)

code/processes/wdb.q
```diff
@@ -375,7 +375,7 @@ informgateway:{[message]
 	$[count gateways:.servers.getservers[`proctype;gatewaytypes;()!();1b;0b];
 	   [
 		   {.[@;(y;x);{.lg.e[`informgateway;"unable to run command on gateway"];'x}]}[message;] each exec w from gateways;
-		   .lg.o[`informgateway;"the message - ", message, " was sent to the gateways"]
+		   .lg.o[`informgateway;raze "the message - ",(1# string message), " was sent to the gateways"]
```
jonnypress: Just use -3! here 

,(-3!message)," was ...

code/processes/wdb.q
ThomasSmyth: Should this be ``(`reloadstart;`)``?

code/processes/wdb.q
ThomasSmyth: Think you can get rid of `raze` with:
```
"the message - ",first[string message]`," was sent to the gateways"
```

jonathonmcmurray: Is `.Q.s` preferred over `-3!`? https://code.kx.com/q/ref/internal/#-3x-qs

ThomasSmyth: Yes, it should be consistent even if the internal function changes.

code/processes/wdb.q
```diff
@@ -226,7 +226,7 @@ d:()!()
 doreload:{[pt]
 	.wdb.reloadcomplete:0b;
 	/-inform gateway of reload start
-	informgateway["reloadstart[]"];
+	informgateway(`reloadstart;);
```
ThomasSmyth: Typo, as per both comments this should be ``(`reloadstart;`)``.

Joe-Griffiths: That was foolish of me

Joe-Griffiths:
Good spot @jonnypress added that now, that is all instances of informgateway sorted.

PR #137, opened 2019-02-05
Grafana update

kalkis:
To summarize additions to the adapter:

1. functions can be applied to create table panels for timeseries and non timeseries data
2. functions can be applied to create graph panels for timeseries data only
3. Previously if you selected "t.trades" or "t.trades.MSFT" for your query in a table panel, this would only work if you had also selected timeserie from the response dropdown; now, if you select table, this will create a table as expected
4. Keyed tables can now be displayed.

Things that are not yet working/implemented:

1. Creating a graph from non-timeseries data is not yet implemented. Further modification of tbfunc will be needed to return the right grafana response.

For review by @FionaMorgan & @Joe-Griffiths

### DIFF
--- a/code/common/grafana.q
+++ b/code/common/grafana.q
@@ -51,10 +51,7 @@ search:{[rqt]
   tabs:tables[];
   symtabs:tabs where sym in'cols each tabs;
   timetabs:tabs where timecol in'cols each tabs;
-  func: string system "f";
   rsp:string tabs;
-  // prefixes "f" to each user defined function on the server
-  rsp,:raze prefix["f";] each (func; raze prefix[;func] each "gto");
   
   if[count timetabs;
     rsp,:s1:prefix["t";string timetabs];
--- a/code/common/grafana.q
+++ b/code/common/grafana.q
@@ -57,8 +57,8 @@ search:{[rqt]
   rsp,:raze prefix["f";] each (func; raze prefix[;func] each "gto");
   
   if[count timetabs;
-	rsp,:s1:prefix["t";string timetabs];
-	rsp,:s2:prefix["g";string timetabs];
+    rsp,:s1:prefix["t";string timetabs];
+    rsp,:s2:prefix["g";string timetabs];
     rsp,:raze(s2,'del),/:'c1:string {cols[x] where`number=types (0!meta x)`t}each timetabs;
     rsp,:raze(prefix["o";string timetabs],'del),/:'c1;
     if[count symtabs;
--- a/code/common/grafana.q
+++ b/code/common/grafana.q
@@ -29,7 +29,7 @@ epoch:946684800000;
 zpp:{
   // get API url from request
   // cuts at first whitespace char to avoid splitting function params
-  r: (0;n?" ") cut n:first x;
+  r:(0;n?" ")cut n:first x;
   // convert grafana message to q
   rqt:.j.k r 1;
   $["query"~r 0;query[rqt];"search"~r 0;search rqt;`$"Annotation url nyi"]
@@ -59,7 +59,7 @@ search:{[rqt]
   if[count timetabs;
 	rsp,:s1:prefix["t";string timetabs];
 	rsp,:s2:prefix["g";string timetabs];
-    rsp,:raze(s2,'del),/:'c1:string {(cols x) where`number=types (0!meta x)`t}each timetabs;
+    rsp,:raze(s2,'del),/:'c1:string {cols[x] where`number=types (0!meta x)`t}each timetabs;
     rsp,:raze(prefix["o";string timetabs],'del),/:'c1;
     if[count symtabs;
       rsp,:raze(s1,'del),/:'c2:string each finddistinctsyms'[timetabs];
@@ -100,8 +100,8 @@ tbfunc:{[rqt]
 tsfunc:{[x]	
   targ: raze x[`targets]`target;
   / split arguments
-  $[isfunc[targ]; numargs:count args:(0;1+targ?del) cut targ:2_targ; numargs:count args:`$del vs targ];
-  $[10h=abs type args 0; tyargs: `$1#args 0; tyargs: args 0];
+  numargs:count args:$[isfunc targ;(0;1+targ?del)cut targ:2_targ;`$del vs targ];
+  tyargs:$[10h=abs type args 0;`$1#;]args 0;
   // manipulate queried table
   coln:cols rqt:0!value args 1;
   // function to convert time to milliseconds, takes timestamp
--- a/code/common/grafana.q
+++ b/code/common/grafana.q
@@ -28,8 +28,9 @@ epoch:946684800000;
 // retrieve and convert Grafana HTTP POST request then process as either timeseries or table
 zpp:{
   // get API url from request
-  r:" " vs first x;
-  // convert grafana mesage to q
+  // cuts at first whitespace char to avoid splitting function params
+  r: (0;n?" ") cut n:first x;
+  // convert grafana message to q
   rqt:.j.k r 1;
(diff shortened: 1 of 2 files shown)


code/common/grafana.q
```diff
@@ -65,23 +73,41 @@ diskvals:{c:(count[x]-ticks)+til ticks;get'[.Q.ind[x;c]]};
...
+// builds body of table response in Json adaptor schema
+tabresponse:{[colname;coltype;rqt] .j.j enlist`columns`rows`type!(flip`text`type!(colname;coltype);catchvals rqt;`table)};
+
 // process a table request and return in JSON format
 tbfunc:{[rqt]
-  rqt:value raze rqt[`targets]`target;
+  rqt: raze rqt[`targets]`target;
+  symname:0b;
+  // if f.t.func, drop first 4 chars
+  $[isfunc[rqt] & istab[2_rqt]; rqt:4_rqt;
```
FionaMorgan: if else $ returns a value so you could use `rqt:[isfunc[rqt] & istab[2_rqt]; 4_rqt; ...` instead

code/common/grafana.q
```diff
@@ -65,23 +73,41 @@ diskvals:{c:(count[x]-ticks)+til ticks;get'[.Q.ind[x;c]]};
...
+
 // process a table request and return in JSON format
 tbfunc:{[rqt]
-  rqt:value raze rqt[`targets]`target;
+  rqt: raze rqt[`targets]`target;
+  symname:0b;
+  // if f.t.func, drop first 4 chars
+  $[isfunc[rqt] & istab[2_rqt]; rqt:4_rqt; 
+	// if f.func drop first 2
+	$[isfunc[rqt]; rqt:2_rqt;
```
FionaMorgan: you're already in an ifelse statement so no need for this second one

code/common/grafana.q
FionaMorgan:
```suggestion
  rqt:$[isfunc[rqt] & istab[2_rqt]; 4_rqt; 
    isfunc[rqt]; 2_rqt;
    istab[rqt]; [rqt: `$del vs rqt; if[2<count rqt; symname: rqt 2]; rqt 1]
    ]
```

code/common/grafana.q
```diff
@@ -42,20 +43,27 @@ query:{[rqt]
...
+  func: string system "f";
   rsp:string tabs;
+  // prefixes "f" to each user defined function on the server
+  rsp,:raze prefix["f";] each (func; raze prefix[;func] each "gto");
+  
   if[count timetabs;
-    rsp,:s1:("t",del),/:string timetabs;
-    rsp,:s2:("g",del),/:string timetabs; 
+	rsp,:s1:prefix["t";string timetabs];
+	rsp,:s2:prefix["g";string timetabs];
```
ThomasSmyth: Please convert these tabs to spaces.

code/common/grafana.q
```diff
@@ -65,23 +73,37 @@ diskvals:{c:(count[x]-ticks)+til ticks;get'[.Q.ind[x;c]]};
...
+// builds body of table response in Json adaptor schema
+tabresponse:{[colname;coltype;rqt] .j.j enlist`columns`rows`type!(flip`text`type!(colname;coltype);catchvals rqt;`table)};
+
 // process a table request and return in JSON format
 tbfunc:{[rqt]
-  rqt:value raze rqt[`targets]`target;
+  rqt: raze rqt[`targets]`target;
+  symname:0b;
+  // if f.t.func, drop first 4 chars
+  rqt:0!value $[isfunc[rqt] & istab[2_rqt]; 4_rqt;
```
ThomasSmyth: Is there a neater way to do this?

code/common/grafana.q
```diff
@@ -65,23 +73,37 @@ diskvals:{c:(count[x]-ticks)+til ticks;get'[.Q.ind[x;c]]};
...
+                isfunc[rqt]; 2_rqt;
+                istab[rqt]; [rqt: `$del vs rqt; if[2<count rqt; symname: rqt 2]; rqt 1];
+                rqt];
   // get column names and associated types to fit format
   colname:cols rqt;
   coltype:types (0!meta rqt)`t;
-  // build body of response in Json adaptor schema
-  :.j.j enlist`columns`rows`type!(flip`text`type!(colname;coltype);catchvals rqt;`table);
+  // search rqt for sym if symname was set
+  if[-11h=type symname; rqt:?[rqt;enlist(=;sym;enlist symname);0b;()]];
```
ThomasSmyth: Does this need to be a functional select? Is this the same as:
```
select from rqt where sym=symname;
```

code/common/grafana.q
```diff
@@ -65,23 +73,37 @@ diskvals:{c:(count[x]-ticks)+til ticks;get'[.Q.ind[x;c]]};
...
 
 // process a timeseries request and return in Json format, takes in query and information dictionary
-tsfunc:{[x]
+tsfunc:{[x]	
+  targ: raze x[`targets]`target;
   / split arguments
-  numargs:count args:`$del vs raze x[`targets]`target;
-  tyargs:args 0;
+  $[isfunc[targ]; numargs:count args:(0;1+targ?del) cut targ:2_targ; numargs:count args:`$del vs targ];
+  $[10h=abs type args 0; tyargs: `$1#args 0; tyargs: args 0];
```
ThomasSmyth: Can simplify:
```
numargs:count args:$[isfunc targ;(0;1+targ?del)cut targ:2_targ;`$del vs targ];
tyargs:$[10h=abs type args 0;`$1#;]args 0;
```

code/common/grafana.q
```diff
@@ -28,8 +28,9 @@ epoch:946684800000;
 // retrieve and convert Grafana HTTP POST request then process as either timeseries or table
 zpp:{
   // get API url from request
-  r:" " vs first x;
-  // convert grafana mesage to q
+  // cuts at first whitespace char to avoid splitting function params
+  r: (0;n?" ") cut n:first x;
```
ThomasSmyth: Can trim the space here.

code/common/grafana.q
```diff
@@ -42,20 +43,27 @@ query:{[rqt]
...
+
 search:{[rqt]
   // build drop down case options from tables in port
   tabs:tables[];
   symtabs:tabs where sym in'cols each tabs;
   timetabs:tabs where timecol in'cols each tabs;
+  func: string system "f";
   rsp:string tabs;
+  // prefixes "f" to each user defined function on the server
+  rsp,:raze prefix["f";] each (func; raze prefix[;func] each "gto");
```
ThomasSmyth:
```
rsp,:raze prefix["f"]each(func;raze prefix[;func]each"gto");
```

code/common/grafana.q
jonathonmcmurray: It's not quite the same, `sym` is a variable by the looks of it

ThomasSmyth: I don't see `sym` defined elsewhere in the function. @kalkis can let us know.

kalkis: @ThomasSmyth 

`sym` is defined at the top of the script in lines 5-6 here:

```// user defined column name of sym column
sym:@[value;`.grafana.sym;`sym];```

It's equal to whatever value the user has set for .grafana.sym, or `sym if none is given. Therefore the functional select is looking for values in the column named sym where they are equal to the symname.

ThomasSmyth: See that now, all good then.

ThomasSmyth:
Thanks for neatening up, I'll let the appropriate parties comment on the functionality.

FionaMorgan:
Looks good! You can squash and merge.

PR #138, opened 2019-02-11
Changes made to <stop> function to accommodate a child/parent pair of PIDs

samuelmcginn:
Issue:
Any torq processes that consisted of a child/parent pair of PIDs would need `./torq stop` to be run twice in order to kill the process properly

Fix:
The `stop` function in `torq.sh` has been altered to do the following
- pull the port number from the `process.csv` (determined by the process name argument as standard)
- use lsof to return only the child PID using the port number provided
- kill the process using the the child PID

### DIFF
--- a/torq.sh
+++ b/torq.sh
@@ -93,8 +93,9 @@ stop() {
     echo "$(date '+%H:%M:%S') | $1 is not currently running"
   else
     echo "$(date '+%H:%M:%S') | Shutting down $1..."
-    pid=$(findproc "$1")
-    eval "kill -15 $pid"                                                                            # kill process pid
+    procno=$(awk '/,'$1',/{print NR}' "$CSVPATH")
+    port=$(eval echo \$"$(getfield "$procno" port)" | bc)
+    eval "kill -15 `lsof -i :$port -sTCP:LISTEN | awk '{if(NR>1)print $2}'`"
   fi
  }
 


samuelmcginn:
Thanks Jonathon,
Jonny, Jamie, James, Thomas, when you get a minute would you mind checking this over?

jprgrant:
Can you start up some processes linked in this way on homer so I can run the unix commands?
Is starting up child processes like this part of one of the core TorQ libraries, or is it something that is happening in user code we need to cater for?

samuelmcginn:
Yeah Jamie I've got a stack running on baseport 40100, my repo has the change put into torq.sh here:
`/home/smcginn/TorQ/TorQ`

I don't think that this is a very common scenario encountered by anyone using TorQ, it's an issue we're hitting at FXCM because of the way we run certain processes using an external `taskset.q` script which is responsible for setting up the child/parent pair as opposed to a single PID

frenzy21:
Silly question but is this a good idea generally (Sam go ahead and change bc to expr )

Surely we should be connecting and killing these gracefully? [ Yes I know we are using -15 ]

jonathonmcmurray:
@frenzy21 I think connecting and killing gracefully would be nice, but what if a process is unresponsive?

frenzy21:
That's fine but should we essentially not have a "try" first and then if that fails or time sout for any reason then smash it?

jonathonmcmurray:
Actually, come to think of it we already have a `kill` TorQ process (https://github.com/AquaQAnalytics/TorQ/blob/master/code/processes/kill.q), which was used by the old stop script:

https://github.com/AquaQAnalytics/TorQ-Finance-Starter-Pack/blob/master/stop_torq_demo.sh

Should probably use that, at least as a first attempt

jprgrant:
Sorry, I'm not understanding the benefits of having 2 different kill methods in the code base. Can someone explain what we get from a q initiated 'graceful' shutdown that we don't get from SIGTERM?

```
jgrant@homer:~$ echo ".z.exit:{-1\"Goodbye cruel world\";}" > die.q
jgrant@homer:~$ nohup `which q` die.q -p 0W </dev/null >/tmp/out.txt 2>&1 &
[1] 9364
jgrant@homer:~$ kill -15 9364
jgrant@homer:~$ cat /tmp/out.txt
Goodbye cruel world
```

frenzy21:
I would like to see the q session killed differently as I would like to know when it was killed with the torq script ( gracefully) and when it was killed "out of process" or forceably.

Essentially if we kill with the q script and by an explicit call we can also add a message " Killed by x with torq script" or something.

When the torq process is killed with a kill -15 it just dies, and I have no idea who did it? [ Yes the exit message would help and should be implemented but that only partially solves the issue. ]

Secondly as the gracefull kill may not work, as a last resort we could have a second option to force kill. [ with kill -s 9 or something if the ones above dont work. 

[ Yes, I missed this was doing -SIGINT ]

PR #139, opened 2019-02-14
torq.sh stop updated to use $(( )) instead of bc

samuelmcginn:
No change to previous PR for altering the `./torq.sh stop` function - only difference is that `bc` is not implemented, as servers are not guaranteed to have it installed - now using `$(( ))` to evaluate the `{baseport}+offset` instead 

Note* `expr` could be used, but would require spaces to work and so would mean adding more logic to insert spaces


samuelmcginn:
Hi all,
Going to merge this - unless anyone has any issues?

PR #141, opened 2019-02-15
Allow in process calls of .gw.syncexec

FionaMorgan:
When .gw.syncexec is called in process (including when qconed in) it now returns the result of the query.

Only works for kdb+ version 3.5 or below.

### DIFF
--- a/code/processes/gateway.q
+++ b/code/processes/gateway.q
@@ -483,7 +483,7 @@ asyncexec:asyncexecjpt[;;raze;();0Wn]
 // execute a synchronous query
 syncexecjpre36:{[query;servertype;joinfunction]
  // Check correct function called
- if[not .gw.call .z.w;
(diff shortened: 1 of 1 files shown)


code/processes/gateway.q
```diff
@@ -483,7 +483,7 @@ asyncexec:asyncexecjpt[;;raze;();0Wn]
 // execute a synchronous query
 syncexecjpre36:{[query;servertype;joinfunction]
  // Check correct function called
- if[not .gw.call .z.w;
+ if[(.z.w<>0)&(.z.w in key .gw.call)&not .gw.call .z.w;
```
jonathonmcmurray: do you need both these checks? will `0` be in `key .gw.call`? my understanding is it wouldn't be and if that's the case, I think the second condition implies the first?

FionaMorgan: Zero is in .gw.call after the first call. The second one is to allow for calls to work in qcon

PR #142, opened 2019-02-15
Summary now prints all ports that are listening, not just a single port

samuelmcginn:
This change is to cover scenarios where a single TorQ process has multiple ports listening

The solution is to swap the position of `PID` and `PORT` in the summary table, so that the `PORT` column can now print out multiple ports separated by spaces

Changes:
swapping `$port` and `$pid` in the summary `printf`
swapping the headers `PORT` and `PID` in the summary printf
changing `$port` variable to a list of PIDs rather than pulling only one using `netstat`

### DIFF
--- a/torq.sh
+++ b/torq.sh
@@ -102,8 +102,8 @@ summary() {
     printf "%-8s | %-14s | %-6s |\n" "$(date '+%H:%M:%S')" "$1" "down"                              # summary table row for down process
   else
     pid=$((findproc "$1")|awk 'END{print}')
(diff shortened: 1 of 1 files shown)


jonathonmcmurray:
Can a q process be listening on multiple ports? I didn't know that was possible, but I don't understand what this PR is for if not?

frenzy21:
The process is listening on multiple ports.



On 15 Feb 2019, at 12:43, Jonathon McMurray <notifications@github.com<mailto:notifications@github.com>> wrote:


Can a q process be listening on multiple ports? I didn't know that was possible, but I don't understand what this PR is for if not?

samuelmcginn:
Hi all - does anyone have any issues with this?

samuelmcginn:
Hi would someone mind approving this if there are no issues so that I can merge?

PR #143, opened 2019-02-19
Monitor checks

ruelaidler:
This PR requests to add a new feature to the monitor process. 

It adds checkmonitor.q script to its own directory, which is called from monitor.q. It works by making asynchronous calls to target processes and runs them against resultchecker functions. These checks can be user-defined. 

The monitor process configuration script has been updated to include this new feature. Documentation has been added under Monitor in Processes.md.


code/monitor/checkmonitor.q
```diff
@@ -0,0 +1,262 @@
...
+ // read in config CSV (actually pipe delimited)
+ .lg.o["reading monitoring config from ",string file:hsym file];
+ // read in csv file, trap error
+ c:.[0:;(("SSS***NN";enlist"|");hsym file);{.lg.e["failed to load monitoring configuration file: ",x]; exit 1}];
+ // attempt to parse the params value
+ p:{@[value;x;{[x;y;e] .lg.e["failed to parse param value from config file at row ",(string y)," with definition ",x,": ",e];exit 2}[x;y]]}'[c`params;til count c]; 
+ // check each params value is a dictionary
+ if[not all 99h=type each p;  
+  .lg.e["all param values must have type dictionary. Values at rows ",(.Q.s1 where not 99h=type each p)," do not"];  
+  exit 3];
```
ThomasSmyth: Can double space indents be used instead of single space throughout the file? Should make things a bit easier to read.

code/monitor/checkmonitor.q
```diff
@@ -0,0 +1,262 @@
...
+ c:update params:p from c;
+ // add the config
+ addconfig c;
```
ThomasSmyth: Combine with 61?

code/monitor/checkmonitor.q
```diff
@@ -0,0 +1,262 @@
...
+ metric:`symbol$(); 		// specific check
+ process:`symbol$(); 		// process it was run on 
+ lastrun:`timestamp$();		// last time it was run
+ nextrun:`timestamp$();		// next time it will be run
+ status:`short$();		// status 
+ executiontime:`timespan$();	// time the execution took
+ totaltime:`timespan$();	// total time- including the network transfer time+queue time on target
+ timerstatus:`short$();		// whether the check run in the correct amount of time
+ running:`short$();		// whether the check is currently running
+ result:())			// error message
```
jonathonmcmurray: comments on these could be aligned? (same for the other schemas below)

code/monitor/checkmonitor.q
```diff
@@ -0,0 +1,262 @@
...
+`checktracker insert (0Ni;0Np;0Np;0Nn;0Ni;0Nh;());
+
+// initialise the runid to 0
+runid:0i
+ 
+readmonitoringconfig:{[file]
+ // read in config CSV (actually pipe delimited)
+ .lg.o["reading monitoring config from ",string file:hsym file];
+ // read in csv file, trap error
+ c:.[0:;(("SSS***NN";enlist"|");hsym file);{.lg.e["failed to load monitoring configuration file: ",x]; exit 1}];
```
jonathonmcmurray: is `exit 1` required? I think `.lg.e` exits if not running with `-debug` on cmd line. If the goal here is to exit with a specific exit code, I don't think that will work unless in debug mode; `.lg.e` will already have exitted with code 3 I believe

code/monitor/checkmonitor.q
```diff
@@ -0,0 +1,262 @@
...
+ // check type
+ if[not type[newval]=type current paramkey;
+  '"supplied value type ",(string type newval)," doesn't match current type ",string type current paramkey];
+ // crack on
+ .[`checkconfig;(checkid;`params;paramkey);:;newval]; 
+ }
+
+forceconfig:{[checkid;newconfig]
+ // force config over the top of current
+ // don't check for existance, types etc.
```
jonathonmcmurray: is this comment correct? seems to check for existence on next two lines?

code/monitor/checkmonitor.q
```diff
@@ -0,0 +1,262 @@
...
+
+// RESULT HANDLERS
+// some example result handler functions here 
+// these should always return `status`result!(status; message)
+// they should take two params- p (dictionary parameters) and r (resultrow)
+
+checkcount:{[p;r]
+ if[`morethan=p`cond;
+  if[p[`count]<r`result; :`status`result!(1h;"")];
+   :`status`result!(0h;"variable ",(string p`varname)," has count of ",(string r`result)," but requires ",string p`count)];
```
jonathonmcmurray: think it's more clear to have the ending `]` for a multiline `if` on a newline

code/processes/monitor.q
```diff
@@ -1,3 +1,10 @@
+//TorQ Monitor Process
+
+//configurable parameters for check monitoring
+.monitor.configcsv:@[value;`.monitor.configcsv;first .proc.getconfigfile["monitorconfig.csv"]];                           //name of config csv to load in
+.monitor.configstored:@[value;`.monitor.configstored;first .proc.getconfigfile["monitorconfig"]];                        //name of stored table for save and reload
+.monit.checkinterval:@[value;`.monit.checkinterval;0D00:00:05];                                                      //interval to run checks
```
jonathonmcmurray: align comments in this block

jonathonmcmurray:
On a more general note, is it not potentially a bit confusing to have both a binary & plaintext version of the config file, where editing the plaintext version will have no effect once the binary version has been generated? Has it been decided that the potential for confusion is worth the flexibility of what can be done with kdb data structures? 

Would an alternative plaintext format be more suited to the data structure here (which I think is a dictionary inside a table) e.g. json, which allows for more nested structure than csv (although possibly a bit more limited in terms of datatypes)?

code/monitor/checkmonitor.q
```diff
@@ -0,0 +1,271 @@
...
+  // override the status and error message as appropriate
+  toinsert[`status]:toinsert[`status] and r`status;
+  toinsert[`result]:r`result; 
+  ];
+ 
+ // if the query has failed, add in the error
+  if[not x`status; toinsert[`result]:"request failed on remote server: ",x`result];
+ 
+  // insert the record into checkstatus
+  `checkstatus upsert ((enlist `checkid)!enlist toinsert`checkid),
```
ThomasSmyth: Might be the same as either of these:
```
enlist[`checkid]#toinsert;
(1#`checkid)#toinsert;
```

code/monitor/checkmonitor.q
```diff
@@ -0,0 +1,271 @@
...
+// RESULT HANDLERS
+// some example result handler functions here 
+// these should always return `status`result!(status; message)
+// they should take two params- p (dictionary parameters) and r (resultrow)
+
+checkcount:{[p;r]
+  if[`morethan=p`cond;
+  if[p[`count]<r`result; :`status`result!(1h;"")];
+  :`status`result!(0h;"variable ",(string p`varname)," has count of ",(string r`result)," but requires ",string p`count)
+   ];
```
ThomasSmyth: I think this should be:
```
if[`morethan=p`cond;
  if[p[`count]<r`result; :`status`result!(1h;"")];
  :`status`result!(0h;"variable ",(string p`varname)," has count of ",(string r`result)," but requires ",string p`count)
];
```

code/monitor/checkmonitor.q
ruelaidler: so the specification is redundant? if so, can i just take it out?

jonathonmcmurray: I think so 👍 it won't get to the `exit` command unless you're in debug mode, and if you're in debug mode, you probably don't want to exit (in the general case, want to be able to poke around and figure out what caused the error)

code/monitor/checkmonitor.q
ruelaidler: just tried these, they don't seem to upsert the time data

ruelaidler:
> On a more general note, is it not potentially a bit confusing to have both a binary & plaintext version of the config file, where editing the plaintext version will have no effect once the binary version has been generated? Has it been decided that the potential for confusion is worth

This is something Jonny and I came up with...another alternative I guess would be to override the CSV? I do think saving down the flat table does maintain the cool flexibility of kdb+ tables.

ruelaidler:
hi, can this be reviewed again please? think all the comments have now been addressed.

code/monitor/checkmonitor.q
```diff
@@ -0,0 +1,271 @@
...
+`checktracker insert (0Ni;0Np;0Np;0Nn;0Ni;0Nh;());
+
+// initialise the runid to 0
+runid:0i
+ 
+readmonitoringconfig:{[file]
+  // read in config CSV (actually pipe delimited)
+  .lg.o["reading monitoring config from ",string file:hsym file];
+  // read in csv file, trap error
+  c:.[0:;(("SSS***NN";enlist"|");hsym file);{.lg.e["failed to load monitoring configuration file: ",x]}];
```
ThomasSmyth: Minor but `hsym` is run on file on line 52 so doesn't need run here.

code/monitor/checkmonitor.q
```diff
@@ -46,17 +46,26 @@ checktracker:(
...
+ };
+
 readmonitoringconfig:{[file]
   // read in config CSV (actually pipe delimited)
   .lg.o["reading monitoring config from ",string file:hsym file];
   // read in csv file, trap error
-  c:.[0:;(("SSS***NN";enlist"|");hsym file);{.lg.e["failed to load monitoring configuration file: ",x]}];
+  c:.[0:;(("SS****NN";enlist"|");file);{.lg.e["failed to load monitoring configuration file: ",x]}];
+  //ungroup checks and make new row for each process
+  c:duplicateconfig[update `$";"vs/:process from c];
```
jonnypress: I think you would be better to duplicate the config after the next set of checks. Rationale is that the error messages for the subsequent parsing checks are based on row numbers - if you expand the duplicates in advance of these checks, then the row numbers will not match those in the raw file and it will be confusing

code/monitor/checkmonitor.q
```diff
@@ -123,6 +132,16 @@ addconfig:{
...
 
+copyconfig:{[checkid;newproc]
+  //function to copy config from checkconfig table and reinsert with new target process
+  //check if supplied checkid exists
+  if[not checkid in exec checkid from checkconfig;
+    '"supplied checkid doesn't exist in checkconfig table"
+  ];
+  newcheckid: 1i+exec max checkid from checkconfig;
+  `checkconfig upsert update checkid:newcheckid,process:newproc from 0!select from checkconfig where checkid=checkid
+ }
```
jonnypress: If this is to duplicate a check from one process to another, I think you should pass it to addcheck or addconfig to make sure that it gets implemented properly and actually run (I think if you just insert it directly to checkconfig then that will not happen)

code/monitor/checkmonitor.q
```diff
@@ -249,6 +268,15 @@ statusbyfam:{[f]
   `status`timerstatus xasc select from checkstatus where family in f
  }
 
+
+//Delete rows from Checkstatus 
+deleterows:{[checkids]
+ if[all null checkids; :delete from `checkstatus];
+ delete from `checkstatus where checkid in checkids
+ }
```
jonnypress: What's this one for? I don't think we should be deleting out of checkstatus - it will likely screw things up

jonnypress: This might be mis-communication - was this the thing I asked for to periodically remove rows over a certain age from the checktracker table?

ruelaidler: ah i have in my notes to delete rows from checkstatus. i can change it to be that ^

code/processes/monitor.q
```diff
@@ -1,3 +1,10 @@
+//TorQ Monitor Process
+
+//configurable parameters for check monitoring
+.monitor.configcsv:@[value;`.monitor.configcsv;first .proc.getconfigfile["monitorconfig.csv"]];                      //name of config csv to load in
+.monitor.configstored:@[value;`.monitor.configstored;first .proc.getconfigfile["monitorconfig"]];                    //name of stored table for save and reload
+.monit.checkinterval:@[value;`.monit.checkinterval;0D00:00:05];                                                      //interval to run checks 
+.monit.checktimeinterval:@[value;`.monit.checktimeinterval;0D00:00:05];                                              //interval to make sure checks are not lagging
```
jonnypress: Should these be .monitor?

code/processes/monitor.q
```diff
@@ -52,10 +59,27 @@ bucketlmchartdata:{[x] x:`minute$$[x=0;1;x];0!select errcount:count i by (0D00:0
...
+// capture any prior definition
+.z.exit:{[x;y] saveconfig[.monitor.configstored;checkconfig];x@y}[@[value;`.z.exit;{{[x]}}]]
+
+//initialise monitor checks
+initcheck[]
+
+//Timers
+.timer.repeat[.z.p;0Wp;0D00:00:05;(`runnow;`);"run the monitoring checks"]
+.timer.repeat[.z.p;0Wp;0D00:00:05;(`checkruntime;0D00:01:00.000000000);"update status if running slow"]
+.timer.repeat[.z.p;0Wp;0D00:00:05;(`cleartracker;0D00:01:00.000000000);"delete rows if over certain age"]
```
jonnypress: Should these be using checkinterval/checktimeinterval?

config/settings/monitor.q
```diff
@@ -1,5 +1,15 @@
 // Default configuration for the monitor process
 
+\d .monitor 
+configcsv: first .proc.getconfigfile["monitorconfig.csv"];
```
jonnypress: Can you add a comment in here for what each of these variables are i.e. what they control

PR #144, opened 2019-02-20
loaddir added to startline params

samuelmcginn:
This change is extending the functionality of a previous commit that enabled `loaddir` to be added manually to the startline for TorQ processes.

Adding the <loaddir> param to the <startline> function in <torq.sh> means that a start line containing a <loaddir> parameter can be output from the function, pulling from the process.csv file, rather than having to add `-loaddir` to a printed start line manually

### DIFF
--- a/torq.sh
+++ b/torq.sh
@@ -52,7 +52,7 @@ findproc() {
 startline() {
   procno=$(awk '/,'$1',/{print NR}' "$CSVPATH")                                                     # get line number for file
   proctype=$(getfield "$procno" "proctype")                                                         # get proctype for process 
-  params="U localtime g T w load"                                                                   # list of params to read from config 
+  params="U localtime g T w loaddir load"                                                           # list of params to read from config 
   sline="${TORQHOME}/torq.q -stackid ${KDBBASEPORT} -proctype $proctype -procname $1"               # base part of startup line
   for p in $params; do                                                                              # iterate over params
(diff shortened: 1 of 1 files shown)


samuelmcginn:
For this change it may be sensible to add a loaddir column to the process csv in the TorQ Finance Starter Pack also - I can make this PR

samuelmcginn:
@jonathonmcmurray no, if the loaddir column isn't included in process.csv, the startline is assembled without a -loaddir param

frenzy21:
As long as it works I don’t mind.

In fairness though you could argue the reverse as well. What I really care about is not picking up the swap files.

James

On 20 Feb 2019, at 15:39, Jonny Press <notifications@github.com<mailto:notifications@github.com>> wrote:


@jonnypress commented on this pull request.

Can -loaddir not just be added in the extras field? What's the issue with that? (or I guess turning the question around, if we are adding this one, why aren't we adding the rest of the torq command line options?)

samuelmcginn:
Yes putting the loaddir param into extras as jonny says does ignore the swp files - so if that method's ok with you James we can just not use this change?

frenzy21:
Yes that’s fine

On 21 Feb 2019, at 09:27, samuelmcginn <notifications@github.com<mailto:notifications@github.com>> wrote:


Yes putting the loaddir param into extras as jonny says does ignore the swp files - so if that method's ok with you James we can just not use this change?

PR #145, opened 2019-02-22
restart function added

samuelmcginn:
This adds a restart option to the torq start line that will stop and start a process with one command

The code is set out in the same structure as the regular `stop `and `start` options

The `restart `option will check, first of all, if the process is up - if this is true, it will stop the process before starting it up again; if this is false, it will simply notify the user that the process was already down, then start the process up again

This has been tested and found to be compatible with using single and multiple processes as well as the `all` flag, for a mix of up and down processes

### DIFF
--- a/torq.sh
+++ b/torq.sh
@@ -118,6 +118,19 @@ stop() {
   fi
  }
 
+restart() {
+  if [[ -z $(findproc "$1") ]]; then                                                                # check process not running
+    echo "$(date '+%H:%M:%S') | $1 is not currently running - starting process..."                  # if not already up, process will start up anyway
+  else
+    echo "$(date '+%H:%M:%S') | Restarting $1..."                                                   # else, process will be stopped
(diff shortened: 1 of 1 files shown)


jonathonmcmurray:
Could this not just be:

```bash
restart() {
  stop $1;
  start $1;
}
```

or something like that?

samuelmcginn:
@jonathonmcmurray It could yeah - I thought it better to add a completely separate function that displays a different message also - could do either though

samuelmcginn:
Hi all - would someone like to weigh in on this and I can close it off?

jonnypress:
I think it would be better to do as Jonathon suggested.  You could put a "restarting PROC" line at the top and then just invoke start and stop. It will be less maintenance.

samuelmcginn:
OK done. Now it just echoes what's being restarted and uses start and stop on all arguments

PR #149, opened 2019-04-11
execute torq.q with q

vbmithr:
I needed this fix to make it start from my Linux install. I don't see how launching a nonexecutable script could work anywhere, but I'm probably missing something.

### DIFF
--- a/torq.sh
+++ b/torq.sh
@@ -59,6 +59,9 @@ startline() {
     sline="$sline$a";                                                                               # append to startup line
   done
   qcmd=$(getfield "$procno" "qcmd")
+  if [ -z $qcmd ]; then
+  qcmd="q"
+  fi
   sline="$qcmd $sline $(getfield "$procno" extras) -procfile $CSVPATH $EXTRAS"                      # append csv file and extra arguments to startup line
   echo "$sline"
  }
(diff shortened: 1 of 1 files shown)


ThomasSmyth:
The q executable is added to the start line `$sline` on line 62 of `torq.sh` based on the contents of `config/process.csv`. In you case it seems that `qcmd` in `process.csv` is not populated for the specific process you are trying to start and there is no fall back to a default behaviour. A fix for this could be to revert your change then add the following between lines 61 + 62:
```
if [ -z $qcmd ]; then
  qcmd = "q"
fi
```
This will default to `q` if no `qcmd` is specified in `config/process.csv`.

Can you try the above and let us know if you have any success?

vbmithr:
Done. Indeed I did not specify `qcmd` in `process.csv`. I added the default you suggested. Now I also needed a fix for the kill command, my second patch. Maybe something was wrong here, but in my case the additional $ causes a problem.

ThomasSmyth:
I see you undid the second change with the additional `$`, was this issue resolved?

vbmithr:
Yes, when using your config file template in the form `localhost,{KDBBASEPORT}+2...` but those scripts are fragile.

PR #151, opened 2019-04-25
Add proctype and proctype to proc start lines in Processes.md

ThomasSmyth:
In `Processes.md` most of the start lines for procs don't include `proctype` and `procname` causing them to fail if run. I have updated these start lines to include the flags so they don't become a point of confusion for those new to TorQ.

The alternative to this may be to set a default `proctype`/`procname` for TorQ? Although that could cause issues if the relevant process code is not loaded and the user does not get an error to notify them accordingly.

### DIFF
--- a/docs/Processes.md
+++ b/docs/Processes.md
@@ -52,7 +52,7 @@ listed as discovery.
 
 To run the discovery service, use a start line such as:
 
-    aquaq $ q torq.q -load code/processes/discovery.q -p 9995
+    aquaq $ q torq.q -load code/processes/discovery.q -p 9995 -proctype discovery -procname discovery1
 
 Modify the configuration as required.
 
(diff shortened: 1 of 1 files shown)


jonathonmcmurray:
So I think the reason the example commands _didn't_ have `procname` & `proctype` is that these can be inferred from port number. So whatever port you have configured in `process.csv` as discovery, if you start a process on that port, it'll automatically set `procname` & `proctype` accordingly

The alternative is to define `procname` & `proctype`, and TorQ will infer and set the port (if applicable)

ThomasSmyth:
Interesting, perhaps this change is not needed *but* `process.csv` being empty by default and users using alternative ports can still lead to errors from following the examples.

jonathonmcmurray:
I agree that specifying `procname` & `proctype` is a clearer example

PR #159, opened 2019-07-30
Processes with no ports

michaelturkington:
Issue from redmine 
http://smithers.aquaq.co.uk/issues/2488

* This update means that if a process's port is not found in process.csv, then the process will no longer start.

### DIFF
--- a/code/handlers/trackservers.q
+++ b/code/handlers/trackservers.q
@@ -118,13 +118,13 @@ cleanup:{if[count w0:exec w from`.servers.SERVERS where not .dotz.livehn w;
 
 / add a new server for current session 
 addnthawc:{[name;proctype;hpup;attributes;W;checkhandle]
-    if[ not 0="J"$last ":"vs string hpup ;
+    if[ 0="J"$last ":"vs string hpup ;:()];
     if[checkhandle and not isalive:.dotz.liveh W;'"invalid handle"];
     cleanup[];
     $[not hpup in (exec hpup from .servers.SERVERS) inter (exec hpup from .servers.nontorqprocesstab);
         `.servers.SERVERS insert(name;proctype;lower hpup;W;0i;$[isalive;.proc.cp[];0Np];.proc.cp[];0Np;attributes); 
         .lg.o[`conn;"Removed double entries: name->", string[name],", proctype->",string[proctype],", hpup->\"",string[hpup],"\""]];
-    W ]
+    W
     }
 
 addh:{[hpuP]
--- a/code/handlers/trackservers.q
+++ b/code/handlers/trackservers.q
@@ -118,7 +118,6 @@ cleanup:{if[count w0:exec w from`.servers.SERVERS where not .dotz.livehn w;
 
 / add a new server for current session 
 addnthawc:{[name;proctype;hpup;attributes;W;checkhandle]
-    if[ 0="J"$last ":"vs string hpup ;:()];
     if[checkhandle and not isalive:.dotz.liveh W;'"invalid handle"];
     cleanup[];
     $[not hpup in (exec hpup from .servers.SERVERS) inter (exec hpup from .servers.nontorqprocesstab);
--- a/code/handlers/trackservers.q
+++ b/code/handlers/trackservers.q
@@ -118,12 +118,13 @@ cleanup:{if[count w0:exec w from`.servers.SERVERS where not .dotz.livehn w;
 
 / add a new server for current session 
 addnthawc:{[name;proctype;hpup;attributes;W;checkhandle]
+    if[ not 0="J"$last ":"vs string hpup ;
     if[checkhandle and not isalive:.dotz.liveh W;'"invalid handle"];
     cleanup[];
     $[not hpup in (exec hpup from .servers.SERVERS) inter (exec hpup from .servers.nontorqprocesstab);
         `.servers.SERVERS insert(name;proctype;lower hpup;W;0i;$[isalive;.proc.cp[];0Np];.proc.cp[];0Np;attributes); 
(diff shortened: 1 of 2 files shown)


code/handlers/trackservers.q
```diff
@@ -118,12 +118,13 @@ cleanup:{if[count w0:exec w from`.servers.SERVERS where not .dotz.livehn w;
 
 / add a new server for current session 
 addnthawc:{[name;proctype;hpup;attributes;W;checkhandle]
+    if[ not 0="J"$last ":"vs string hpup ;
```
jonathonmcmurray: Could you invert the condition here & just return early?
```suggestion
    if[0="J"$last ":"vs string hpup ;:()];
```

michaelturkington: Yep, can do. Is there a benefit of doing this instead?

jonathonmcmurray: I think it's easier to follow - as a general rule, I prefer shorter conditionals. With a long multi-line conditional it's easier to miss where the conditional ends & get confused about what code is executed conditionally.

jonnypress:
Ok I don't think this change is correct. We don't want processes to start up without a host and port. If the user supplies proctype+procname but no port, then it will attempt to read the port from the process.csv file. The change we need here is that if the user supplies proctype+procname and no port can be found in process.csv (based on proctype+procname+host) then the process should fail to start.

tszwailam:
Updated description as we changed the attempt on fixing the issue.

torq.q
```diff
@@ -400,6 +400,7 @@ readprocfile:{[file]
 		// map hostnames in res to order of preference, select most preferred
 		first res iasc prefs?res[`host];
 		first res];
+	if[null output`port;.err.ex[`readprocfile;"Port not found in process.csv. Parameters include host: ", string output`host, ", procname: ", string output`procname, ", proctype: ",string output`proctype;1]];
```
ThomasSmyth: Just needs some squares to avoid joining strings and symbols (and proctype first just for neatness):
```suggestion
	if[null output`port;.err.ex[`readprocfile;"Port not found in process.csv. Parameters include host: ", string[output`host], ", proctype: ", string[output`proctype], ", procname: ",string output`procname;1]]; 
```

jonnypress:
Can you document some test cases on here with test evidence? 
So 
1. if proctype+procname specified, it needs to use it's host to work out which port it should be on from process .csv and set it to that port
2. if as (1) but can't find the port using proctype+procname+host then it fails to start
3. If proctype+procname+port specified then it just starts using those
4. if port specified, it looks up using hostname to find it's proctype+procname. If it finds them, it sets them. If it doesn't, it fails.

ThomasSmyth:
I've just tested (1) by renaming the specified host in `process.csv` and the process starts anyway, will need to address that.

torq.q
```diff
@@ -400,6 +400,9 @@ readprocfile:{[file]
 		// map hostnames in res to order of preference, select most preferred
 		first res iasc prefs?res[`host];
 		first res];
+	// exit if no port passed via command line or specified in process.csv
+	if[null[output`port]&0i=system"p";
+		.err.ex[`readprocfile;"No port passed via -p flag or found in process.csv. Parameters are host: ", string[output`host], ", proctype: ", string[output`proctype], ", procname: ",string output`procname;1]];
```
ThomasSmyth: Use file name as passed to this function
```suggestion
		.err.ex[`readprocfile;"No port passed via -p flag or found in ",string[file],". Parameters are host: ", string[output`host], ", proctype: ", string[output`proctype], ", procname: ",string output`procname;1]]; 
```

torq.q
```diff
@@ -400,6 +400,9 @@ readprocfile:{[file]
 		// map hostnames in res to order of preference, select most preferred
 		first res iasc prefs?res[`host];
 		first res];
+	// exit if no port passed via command line or specified in process.csv
```
ThomasSmyth: Config file is configurable, can give it a generic name:
```suggestion
	// exit if no port passed via command line or specified in config
```

torq.q
```diff
@@ -400,6 +400,9 @@ readprocfile:{[file]
...
+	if[null[output`port]&0i=system"p";
```
ThomasSmyth: Under the strict condition of requiring hostname to match this additional check may be enough:
```suggestion
	if[(null[output`port]|not output[`host]in prefs)&0i=system"p";
```
At least enough for (1) + (2) I reckon.

ThomasSmyth: Under the strict condition of requiring hostname to match this additional check may be enough:
```suggestion
	if[(null[output`port]|not output[`host]in prefs)&0i=system"p";
```
At least enough for (1) + (2) I reckon.

torq.q
```diff
@@ -400,6 +400,9 @@ readprocfile:{[file]
...
+		.err.ex[`readprocfile;"No port passed via -p flag or found in process.csv. Parameters are host: ", string[output`host], ", proctype: ", string[output`proctype], ", procname: ",string output`procname;1]]; 
 	if[not output[`port] = system"p";
```
ThomasSmyth: May address (3):
```suggestion
	if[not[output[`port]=system"p"]&0i=system"p";
```

19CraigDN:
All the tests pass except for 2 when the hostname is empty. It does work in the case where hostname is not empty but does not match host.

ThomasSmyth:
@jonnypress is that acceptable behaviour for (2)?

jonnypress:
@ThomasSmyth I think that's ok (sorry only just going through this). So if the hostname is empty in the config file the process just starts up? I don't think that's necessarily bad

Issue #163, opened 2019-09-25
Housekeeping#find Windows inconsistency


SMLaursen:
Hi,

Looking at TorQ-Finance-Starter-Pack and the https://aquaqanalytics.github.io/TorQ/Processes/#housekeeping I got the impression that the exclude field (housekeeping.csv) can be left empty. Unfortunately on Windows this leads to everything being matched as exclusions therefore not carrying out the specified function. On the other hand an 'os error is thrown if the exclude search doesn't find anything, so as a workaround I currently place a dummy file in the directory for the exclude search to find. 

I havn't tested this on Linux but I assume find "" here exhibits the inverse and desired behavior.

Win10, KDB 3.6 (32bit), TorQ 3.5.0

SMLaursen:
On Windows the following command returns **everything** in the directory
`dir "" /s`
Whereas on Linux the following command returns an empty list
`/usr/bin/find -maxdepth 1 -type f -name ""`
Which leads to the discrepancy. Unfortunatly, on Windows, an empty search result is represented by the String "File Not Found" which I believe causes the error when I put in a dummy exclude rules with no matches.

PR #166, opened 2019-09-26
Fix for windows housekeeping inconsistency

tszwailam:
Attempted fix for housekeeping for Win OS - following up on issue #163 

Now the windows find function follows the unix find logic.

added an early exit to return files as an empty list if match is empty (targeting the issue from 163)

### DIFF
--- a/code/processes/housekeeping.q
+++ b/code/processes/housekeeping.q
@@ -122,7 +122,7 @@ find:{[path;match;age;agemin]
 //defines full list of files based on Windows OS version
 winfilelist:{[path;match]
 	$[version in `w7`w8`w10;-2_(5_system "dir ",path,match);-5_(5_system "dir ",path,match, " /s")]
-}
+ }
 
 
 //removes files
--- a/code/processes/housekeeping.q
+++ b/code/processes/housekeeping.q
@@ -96,6 +96,12 @@ zip:{[FILE]
 
 //-locates files with path, matching string and age
 find:{[path;match;age;agemin]
+	//returns empty list for files if match is an empty string
+        if[""~match;
+          .lg.o[`housekeeping;"No matching files located"];
+          :();
+        ];
+
 	//renames the path to a windows readable format
 	PATH:ssr[path;"/";"\\"];
 	//searches for files and refines return to usable format
--- a/code/processes/housekeeping.q
+++ b/code/processes/housekeeping.q
@@ -108,15 +108,21 @@ find:{[path;match;age;agemin]
         //searches for files and refines return to usable format
 	$[0=agemin;
 	files:.[{[PATH;match;age].lg.o[`housekeeping;"Searching for: ", match];
-		system "z 1";fulllist:-2_(5_system "dir ",PATH,match);
+		system "z 1";fulllist:winfilelist[PATH;match];
 		removelist:fulllist where ("D"${10#x} each fulllist)<.proc.cd[]-age; system "z 0";
 		{[path;x]path,last " " vs x} [PATH;] each removelist};(PATH;match;age);err];
 	files:.[{[PATH;match;age].lg.o[`housekeeping;"Searching for: ", match];
-		system "z 1";fulllist:-2_(5_system "dir ",PATH,match);
+		system "z 1";fulllist:winfilelist[PATH;match];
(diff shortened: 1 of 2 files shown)


code/processes/housekeeping.q
```diff
@@ -96,20 +96,25 @@ zip:{[FILE]
...
 	//renames the path to a windows readable format
 	PATH:ssr[path;"/";"\\"];
-	//searches for files and refines return to usable format
+	//error and info for find function 
+	err:{.lg.e[`housekeeping;"Find function failed: ", x]; ()};
+        //searches for files and refines return to usable format
 	$[0=agemin;
 	files:.[{[PATH;match;age].lg.o[`housekeeping;"Searching for: ", match];
-		system "z 1";fulllist:-5_(5_system "dir ",PATH,match, " /s");
+		system "z 1";fulllist:-2_(5_system "dir ",PATH,match);
```
jonnypress: So I think the crux of this problem is that the functionality of these system commands differs between windows versions (windows 7, windows 10, windows server 20XX etc.). I believe we've changed this exact section before. So what we need to do is extract this functionality into a separate function, which we can extend based on the OS. 

If this bit is what we need now: 

-2_(5_system "dir ",PATH,match)

then that should be the option for whatever version of windows this works for. I'm pretty sure this would've worked in the past too: 
-5_(5_system "dir ",PATH,match, " /s")

So can you create a separate function to encapsulate this and have an OS switch e.g. 

winfind:{$[.windows.version=`w10; .... ; .windows.version=`w7; ... } 

etc. We will have to pick an appropriate default value. 

This should make it more extendible in the future. For now .windows.version will just have to be manually set by the user in config I think unless we can work out a way of working out the version from the cmd prompt (though this itself my then be dependent on the version :-S... thanks Windows!).  kdb will only tell us w32 or w64 which isn't enough

ThomasSmyth: Fair enough, will need this documented. `-2_5_` and no `/s` is what I needed on Windows 8 FWIW.

ThomasSmyth: Windows 7 is the same as Windows 8, with `/s` meaning search subdirs.

code/processes/housekeeping.q
```diff
@@ -108,16 +108,22 @@ find:{[path;match;age;agemin]
...
-		system "z 1";fulllist:-2_(5_system "dir ",PATH,match);
+		system "z 1";fulllist:winfilelist[PATH;match];
 		removelist:fulllist where ({ts:("  " vs 17#x);("D"$ts[0]) + value ts[1]} each fulllist)<.proc.cp[]-`minute$age; system "z 0";
 		{[path;x]path,last " " vs x} [PATH;] each removelist};(PATH;match;age);err]];
 	$[(count files)=0;
 	[.lg.o[`housekeeping;"No matching files located"];files]; files]}
 
+//defines full list of files based on Windows OS version
+winfilelist:{[path;match]
+	$[.windows.version in `w7`w8`w10;-2_(5_system "dir ",PATH,match);-5_(5_system "dir ",PATH,match, " /s")]
```
ThomasSmyth: Local variable is `path` but this references `PATH`. Just noticed this is under `\d .win` so this can become:
```
$[version in`w7`w8`w10;-2_5_system"dir ",path,match;-5_5_system "dir ",path,match," /s"]
```

config/settings/housekeeping.q
```diff
@@ -4,3 +4,7 @@
 inputcsv:first .proc.getconfigfile["housekeeping.csv"]
 runtimes:02:00:00
 runnow:0b
+
+\d .windows
+
+version:`w10
```
ThomasSmyth: Need to add this to the start of `code/processes/housekeeping.q` (with the other variables):
```
.win.version:@[value;`.win.version;`w7]
```
Not the changed namespace, just had a thought that `.win` already exists so lets continue to use it!

ThomasSmyth:
Can find OS name via this command but it's pretty slow:
```
systeminfo | findstr /B /C:"OS Name"
```

ThomasSmyth:
Any test cases for this?

19CraigDN:
We ran the find function with an empty exclude field on both windows 10 and linux to ensure they produced the same result of no files found. They both successfully found no files.
WinNoFilesFound
WinDirectory
LinuxNoFilesFound
LinuxDirectory

PR #168, opened 2019-10-04
introduce reval for gw func calls

### LINKED ISSUE #162: Possible to violate .readonly property via gateway functions
We're in the process of amending our existing KDB setup with TorQ. Initially we just want to prevent accidental deletions and modifications to existing data where **writeaccess.q** appears more suitable than using the fullblown permission setup

I've enabled the readonly parameter on **default.q** (i.e. every process) and as expected i get a \`noupdate error when I try to modify the tables from either the gateway or rdb. But I've noticed that connecting to the gateway and issuing e.g. h(\`.gw.syncexec;"delete from orders";\`rdb) executes thereby violating the readonly setting.

Win10, KDB 3.6 (32bit) and TorQ 3.5.0

jwrno1:
Fixes #162 
Added new eval/value functions for both .pm and .gw namespaces which prevent the readonly function being violated by calling a .gw function.

### DIFF
--- a/code/common/execas.q
+++ b/code/common/execas.q
@@ -1,4 +1,4 @@
 \d .pm
 
-execas:{[f;u;g]
- $[`.pm.requ ~ key `.pm.requ;.pm.requ[u;f];`.pm.valp ~ key `.pm.valp;.pm.valp f; g f]}
+execas:{[f;u]
+ $[`.pm.requ ~ key `.pm.requ;.pm.requ[u;f]; value f]}
--- a/code/common/execas.q
+++ b/code/common/execas.q
@@ -1,4 +1,4 @@
 \d .pm
 
 execas:{[f;u]
- $[`.pm.requ ~ key `.pm.requ;.pm.requ[u;f]; value f]}
+ $[`.pm.requ ~ key `.pm.requ;.pm.requ[u;f]; valp f]}
--- a/code/common/execas.q
+++ b/code/common/execas.q
@@ -1,4 +1,4 @@
 \d .pm
 
 execas:{[f;u]
- $[`.pm.requ ~ key `.pm.requ;.pm.requ[u;f]; valp f]}
+ $[`.pm.requ ~ key `.pm.requ;.pm.requ[u;f];`.pm.valp ~ key `.pm.valp;.pm.valp f; value f]}
--- a/code/common/execas.q
+++ b/code/common/execas.q
@@ -1,4 +1,4 @@
 \d .pm
 
-execas:{[f;u]
- $[`.pm.requ ~ key `.pm.requ;.pm.requ[u;f];`.pm.valp ~ key `.pm.valp;.pm.valp f; value f]}
+execas:{[f;u;g]
+ $[`.pm.requ ~ key `.pm.requ;.pm.requ[u;f];`.pm.valp ~ key `.pm.valp;.pm.valp f; g f]}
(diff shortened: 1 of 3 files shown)


code/common/execas.q
```diff
@@ -1,4 +1,4 @@
 \d .pm
 
 execas:{[f;u]
- $[`.pm.requ ~ key `.pm.requ;.pm.requ[u;f]; value f]}
+ $[`.pm.requ ~ key `.pm.requ;.pm.requ[u;f]; valp f]}
```
jonnypress: This now requires permissions.q to be loaded. Can you add a check to see if valp exists and only use it if it does

code/processes/gateway.q
```diff
@@ -446,7 +449,7 @@ asyncexecjpts:{[query;servertype;joinfunction;postback;timeout;sync]
    :();
    ];
   ];
- query:({[u;q]$[`.pm.execas ~ key `.pm.execas;value (`.pm.execas; q; u);value q]}; .z.u; query);
+ query:({[u;q]$[`.pm.execas ~ key `.pm.execas;value (`.pm.execas; q; u);valp q]}; .z.u; query);
```
jonnypress: you need an existence check in here for valp. We have people who use the TorQ gateway standalone, so wouldn't have loaded the permissions lib

code/processes/gateway.q
```diff
@@ -503,7 +506,7 @@ syncexecjpre36:{[query;servertype;joinfunction]
  handles:(exec serverid!handle from tab)first each (exec serverid from tab) inter/: serverids;
  setserverstate[handles;1b];
  start:.z.p;
- query:({[u;q]$[`.pm.execas ~ key `.pm.execas;value (`.pm.execas; q; u);value q]}; .z.u; query);
+ query:({[u;q]$[`.pm.execas ~ key `.pm.execas; value(`.pm.execas; q; u);valp q]}; .z.u; query);
```
jonnypress: and here - existence check

code/processes/gateway.q
jwrno1: This is a seperate but identical valp that's defined in the .gw namespace, should I change it to just use the .pm namespace valp and have an existence check?

jonnypress: So the "query" function is called on the remote process. It would need to have valp defined. But really it should be an additional conditional, something like 

query:({[u;q]$[`.pm.execas ~ key `.pm.execas;value (`.pm.execas; q; u); `.pm.valp ~ key `.pm.valp; .pm.valp q; value q]}; .z.u; query);

code/common/execas.q
```diff
@@ -1,4 +1,4 @@
 \d .pm
 
 execas:{[f;u]
- $[`.pm.requ ~ key `.pm.requ;.pm.requ[u;f]; value f]}
+ $[`.pm.requ ~ key `.pm.requ;.pm.requ[u;f];`.pm.valp ~ key `.pm.valp;.pm.valp f; value f]}
```
jwrno1: My only worry about the change is that this now makes the readonly property dependent on having the permissions library installed which I thought we wanted to avoid

code/processes/gateway.q
```diff
@@ -446,7 +450,7 @@ asyncexecjpts:{[query;servertype;joinfunction;postback;timeout;sync]
    :();
    ];
   ];
- query:({[u;q]$[`.pm.execas ~ key `.pm.execas;value (`.pm.execas; q; u);value q]}; .z.u; query);
+ query:({[u;q]$[`.pm.execas ~ key `.pm.execas;value (`.pm.execas; q; u);`.pm.valp ~ key `.pm.valp; .pm.valp q; @[value;q;{.lg.e[`gw;"Failed to execute query."]}]]}; .z.u; query;.gw.valp);
```
jonnypress: I don't think this will work as you are defining the function with 2 args but passing in 3 (and same for the "query" function definition later)
Also I don't think you want to have an error trap in this remote execution

PR #171, opened 2019-11-13
Adding error messages to gateway.q

CatherineHigg:


code/processes/gateway.q
```diff
@@ -454,8 +454,19 @@ asyncexecjpts:{[query;servertype;joinfunction;postback;timeout;sync]
    if[99h<>type servertype;
      // its a list of servertypes e.g. `rdb`hdb
      servertype:distinct servertype,();
-     missing:servertype except exec distinct servertype from .gw.servers where active;
-     if[count missing; errStr:.gw.errorprefix,"not all of the requested server types are available; missing "," " sv string missing];
+     //list of active servers
```
rossedens: Can this logic be moved to the `getserverids` function?

code/processes/gateway.q
```diff
@@ -454,8 +454,19 @@ asyncexecjpts:{[query;servertype;joinfunction;postback;timeout;sync]
...
+     activeservers:exec distinct servertype from .gw.servers where active;
+     //list of all servers
+     allservers:exec distinct servertype from .gw.servers;
+     //if any requested servers are missing then:
+       //if requested server does not exist, return error with list of available servers
+       //if requested server exists but is currently inactive, return error with list of available servers
+     if[count servertype except activeservers; 
+       $[max not servertype in allservers; 
+	errStr:.gw.errorprefix,"the following are not valid servers: ",(", " sv string servertype except allservers),". Available servers include: "," " sv string activeservers;
```
jonathonmcmurray: inconsistent indenting on this line which uses tab compared to surrounding which use spaces

CatherineHigg: Will fix that

code/processes/gateway.q
CatherineHigg: Am planning to put the logic in `getserverids` function for the sync queries (pre 3.6) but yeah might be an idea to have `asyncexecjpts` call that function too rather than duplicating code.

code/processes/gateway.q
```diff
@@ -388,7 +388,7 @@ getserverscross:{[req;att;besteffort]
 getserverids:{[att]
   if[99h<>type att;
    // its a list of servertypes e.g. `rdb`hdb
-   servertype:att,();
+   servertype:distinct att,();
```
rossedens: @jonnypress was there any edge case for this not using `distinct` before?

code/processes/gateway.q
```diff
@@ -454,8 +465,19 @@ asyncexecjpts:{[query;servertype;joinfunction;postback;timeout;sync]
...
+     //list of all servers
+     allservers:exec distinct servertype from .gw.servers;
+     //if any requested servers are missing then:
+     //if requested server does not exist, return error with list of available servers
+     //if requested server exists but is currently inactive, return error with list of available servers
+     if[count servertype except activeservers; 
+       $[max not servertype in allservers; 
+        errStr:.gw.errorprefix,"the following are not valid servers: ",(", " sv string servertype except allservers),". Available servers include: "," " sv string activeservers;
+        errStr:.gw.errorprefix,"the following requested servers are currently inactive: ",(", " sv string servertype except activeservers),". Available servers include: "," " sv string activeservers
+       ];
```
ThomasSmyth: A lot of duplication here can be moved outside the conditional:
```
errStr:.gw.errorprefix,"the following ",$[max not servertype in allservers; 
  "are not valid servers: ",", "sv string servertype except allservers;
  "requested servers are currently inactive: ",", "sv string servertype except activeservers
  ],". Available servers include: "," " sv string activeservers;
```

CatherineHigg: Changed now in getserverids.

PR #173, opened 2019-11-14
Add dqe process to TorQ

dmartinaquaq:
Added data quality engine process to TorQ.

Contains function called tableExists that will have a table name as a symbol fed into it.
It will then be able to check in other processes if they contain a table with this name.

### DIFF
--- a/code/processes/dqe.q
+++ b/code/processes/dqe.q
@@ -1,36 +1,23 @@
 \d .dqe
 
-/- default parameters
-subscribeto:@[value;`subscribeto;`];                          /-list of tables subscribed to
-subscribesyms:@[value;`subscribesyms;`];                      /-list of syms subscribed to
-schema:@[value;`schema;0b];                                   /-retrieve schema from tickerplant
-tickerplanttypes:@[value;`tickerplanttypes;`tickerplant];     /-list of tickerplant types to try and make a connection to
-replaylog:@[value;`replaylog;0b];                             /-replay the tickerplant log file
-tpconnsleepintv:@[value;`tpconnsleepintv;10];                 /-default wait time before trying to reconnect to discovery/TP
-/- end of default parameters
-
-subscribe:{[]
-  if[0=count s:.sub.getsubscriptionhandles[tickerplanttypes;();()!()];:()];
-  .lg.o[`subscribe;"found available tickerplant, attempting to subscribe"];
-  /-set the date that was returned by the subscription code i.e. the date for the tickerplant log file
-  /-and a list of the tables that the process is now subscribing for
-  /-setting subtables and tplogdate globals
-  .dqe,:.sub.subscribe[subscribeto;subscribesyms;schema;replaylog;first s];
-  }
-
-upd:{};
-
 init:{
-  .servers.CONNECTIONS:distinct .servers.CONNECTIONS, .dqe.tickerplanttypes;
+  .servers.CONNECTIONS:`;                                                                                       /- set to nothing so that is only connects to discovery
   .lg.o[`init;"searching for servers"];
-  .servers.startup[];
-  subscribe[];
+  .servers.startup[];                                                                                           /- Open connection to discovery
   }
 
-tableexists:{x in tables[]};                                  /-function to check for table, param is table name as a symbol
+tableexists:{x in tables[]};                                                                                    /- function to check for table, param is table name as a symbol
+
+runcheck:{[fn;vars;rs]                                                                                          /- function used to send other function to test processes
+  h:.servers.opencon each exec hpup from .servers.querydiscovery`ALL where (procname in rs)|(proctype in rs);   /- check with discovery and open handles to other processes using procname and/or proctype
+  if[0=count h;.lg.e[`handle;"cannot open handle to any given processes"];:()];                                 /- check if any handles exist, if not exit function
+  /if[0=(count h)=(count rs);.lg.e[`handle;"not all supplied processes have handles"];                          /- check that number of handles equal input, Needs fixed due to proctype input!
+  ans:{[func;vrs;hand]hand(func;vrs)}[fn;vars]'[h];                                                             /- send function with variables down handle
+  hclose each h;                                                                                                /- close handles as they are no longer needed
+  ans
+  }
 
 \d .
 
-upd:.dqe.upd;
-
 .dqe.init[]
+
--- a/code/processes/dqe.q
+++ b/code/processes/dqe.q
@@ -5,15 +5,19 @@ init:{
   .servers.startup[];                                                                                           /- Open connection to discovery
   }
 
+gethandles:{exec procname,proctype,w from .servers.SERVERS where (procname in x) | (proctype in x)};
+
 tableexists:{x in tables[]};                                                                                    /- function to check for table, param is table name as a symbol
 
 runcheck:{[fn;vars;rs]                                                                                          /- function used to send other function to test processes
   rs:(),rs;                                                                                                     /- set rs to a list
-  h:exec w from .servers.SERVERS where (procname in rs) | (proctype in rs);                                     /- check if processes exist and are valid
+  h:.dqe.gethandles[rs];                                                                                        /- check if processes exist and are valid
   if[0=count h;.lg.e[`handle;"cannot open handle to any given processes"];:()];                                 /- check if any handles exist, if not exit function
-  ans:{[func;vrs;hand]hand(func;vrs)}[fn;vars]'[h];                                                             /- send function with variables down handle
+  ans:{[funct;vars;hand].async.postback[hand;(funct;vars);`.dqe.showresult]}[value fn;vars]'[h[`w]];            /- send function with variables down handle
   ans
   }
+
+showresult:{show x};
 
 \d .
 
--- a/code/processes/dqe.q
+++ b/code/processes/dqe.q
@@ -10,24 +10,23 @@ tpconnsleepintv:@[value;`tpconnsleepintv;10];                 /-default wait tim
 /- end of default parameters
 
 subscribe:{[]
-  if[count s:.sub.getsubscriptionhandles[tickerplanttypes;();()!()];
+  if[0=count s:.sub.getsubscriptionhandles[tickerplanttypes;();()!()];:()];
   .lg.o[`subscribe;"found available tickerplant, attempting to subscribe"];
   /-set the date that was returned by the subscription code i.e. the date for the tickerplant log file
   /-and a list of the tables that the process is now subscribing for
   subinfo: .sub.subscribe[subscribeto;subscribesyms;schema;replaylog;first s];
   /-setting subtables and tplogdate globals
-  .dqe,:subinfo]
+  .dqe,:subinfo
   }
 
 init:{
   .servers.CONNECTIONS:distinct .servers.CONNECTIONS, .dqe.tickerplanttypes;
   .lg.o[`init;"searching for servers"];
   .servers.startup[];
-  .servers.startupdependent[tickerplanttypes;tpconnsleepintv];
   subscribe[];
   }
 
-tableExists:{x in tables[]};                                  /-function to check for table, param is table name as a symbol
+tableexists:{x in tables[]};                                  /-function to check for table, param is table name as a symbol
 
 \d .
 
--- a/code/processes/dqe.q
+++ b/code/processes/dqe.q
@@ -1,7 +1,8 @@
 \d .dqe
 
+.servers.CONNECTIONS:`;                                                                                         /- set to nothing so that is only connects to discovery
+
 init:{
-  .servers.CONNECTIONS:`;                                                                                       /- set to nothing so that is only connects to discovery
   .lg.o[`init;"searching for servers"];
   .servers.startup[];                                                                                           /- Open connection to discovery
   }
@@ -11,7 +12,6 @@ tableexists:{x in tables[]};                                                    
 runcheck:{[fn;vars;rs]                                                                                          /- function used to send other function to test processes
   h:.servers.opencon each exec hpup from .servers.querydiscovery`ALL where (procname in rs)|(proctype in rs);   /- check with discovery and open handles to other processes using procname and/or proctype
   if[0=count h;.lg.e[`handle;"cannot open handle to any given processes"];:()];                                 /- check if any handles exist, if not exit function
-  /if[0=(count h)=(count rs);.lg.e[`handle;"not all supplied processes have handles"];                          /- check that number of handles equal input, Needs fixed due to proctype input!
   ans:{[func;vrs;hand]hand(func;vrs)}[fn;vars]'[h];                                                             /- send function with variables down handle
   hclose each h;                                                                                                /- close handles as they are no longer needed
   ans
--- a/code/processes/dqe.q
+++ b/code/processes/dqe.q
@@ -8,7 +8,9 @@ init:{
 tableexists:{x in tables[]};                                                                                    /- function to check for table, param is table name as a symbol
 
 runcheck:{[fn;vars;rs]                                                                                          /- function used to send other function to test processes
-  h:.servers.opencon each exec hpup from .servers.querydiscovery`ALL where (procname in rs)|(proctype in rs);   /- check with discovery and open handles to other processes using procname and/or proctype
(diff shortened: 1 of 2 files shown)


code/processes/dqe.q
```diff
@@ -0,0 +1,34 @@
...
+subscribeto:@[value;`subscribeto;`];                          /-list of tables subscribed to
+subscribesyms:@[value;`subscribesyms;`];                      /-list of syms subscribed to
+schema:@[value;`schema;0b];                                   /-retrieve schema from tickerplant
+tickerplanttypes:@[value;`tickerplanttypes;`tickerplant];     /-list of tickerplant types to try and make a connection to
+replaylog:@[value;`replaylog;0b];                             /-replay the tickerplant log file
+tpconnsleepintv:@[value;`tpconnsleepintv;10];                 /-default wait time before trying to reconnect to discovery/TP
+/- end of default parameters
+
+subscribe:{[]
+  if[count s:.sub.getsubscriptionhandles[tickerplanttypes;();()!()];
```
ThomasSmyth: No indentation passed this point? :cry:
I'm all for changing this up a bit to exit early here:
```suggestion
  if[0=count s:.sub.getsubscriptionhandles[tickerplanttypes;();()!()];:()];
```
(needs the closing bracket removed on line 19)

code/processes/dqe.q
```diff
@@ -0,0 +1,34 @@
...
+  subinfo: .sub.subscribe[subscribeto;subscribesyms;schema;replaylog;first s];
+  /-setting subtables and tplogdate globals
+  .dqe,:subinfo]
+  }
+
+init:{
+  .servers.CONNECTIONS:distinct .servers.CONNECTIONS, .dqe.tickerplanttypes;
+  .lg.o[`init;"searching for servers"];
+  .servers.startup[];
+  .servers.startupdependent[tickerplanttypes;tpconnsleepintv];
```
jonathonmcmurray: Are both these lines required? I assumed the second would mean the first wasn't needed?

code/processes/dqe.q
```diff
@@ -0,0 +1,34 @@
...
+  subscribe[];
+  }
+
+tableExists:{x in tables[]};                                  /-function to check for table, param is table name as a symbol
```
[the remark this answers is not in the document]
dmartinaquaq: If I am understanding right, this function will be sent over to other processes using a function in the next part .
Will get rid of the camel case. :)

code/processes/dqe.q
```diff
@@ -10,24 +10,23 @@ tpconnsleepintv:@[value;`tpconnsleepintv;10];                 /-default wait tim
...
 subscribe:{[]
-  if[count s:.sub.getsubscriptionhandles[tickerplanttypes;();()!()];
+  if[0=count s:.sub.getsubscriptionhandles[tickerplanttypes;();()!()];:()];
   .lg.o[`subscribe;"found available tickerplant, attempting to subscribe"];
   /-set the date that was returned by the subscription code i.e. the date for the tickerplant log file
   /-and a list of the tables that the process is now subscribing for
   subinfo: .sub.subscribe[subscribeto;subscribesyms;schema;replaylog;first s];
   /-setting subtables and tplogdate globals
-  .dqe,:subinfo]
+  .dqe,:subinfo
```
jonathonmcmurray: I think `subinfo` is assigned, then immediately copied to another var & discarded; could you just have `.dqe,:.sub.subscribe...` above?

code/processes/dqe.q
```diff
@@ -1,34 +1,23 @@
...
-  /-set the date that was returned by the subscription code i.e. the date for the tickerplant log file
-  /-and a list of the tables that the process is now subscribing for
-  subinfo: .sub.subscribe[subscribeto;subscribesyms;schema;replaylog;first s];
-  /-setting subtables and tplogdate globals
-  .dqe,:subinfo]
-  }
-
 init:{
-  .servers.CONNECTIONS:distinct .servers.CONNECTIONS, .dqe.tickerplanttypes;
+  .servers.CONNECTIONS:`;                                                                                       /- set to nothing so that is only connects to discovery
```
jonathonmcmurray: Not sure this should be set inside the init function? I'd set it at the top of the file

code/processes/dqe.q
```diff
@@ -1,34 +1,23 @@
...
+  .servers.startup[];                                                                                           /- Open connection to discovery
   }
 
-tableExists:{x in tables[]};                                  /-function to check for table, param is table name as a symbol
+tableexists:{x in tables[]};                                                                                    /- function to check for table, param is table name as a symbol
+
+runcheck:{[fn;vars;rs]                                                                                          /- function used to send other function to test processes
+  h:.servers.opencon each exec hpup from .servers.querydiscovery`ALL where (procname in rs)|(proctype in rs);   /- check with discovery and open handles to other processes using procname and/or proctype
+  if[0=count h;.lg.e[`handle;"cannot open handle to any given processes"];:()];                                 /- check if any handles exist, if not exit function
+  /if[0=(count h)=(count rs);.lg.e[`handle;"not all supplied processes have handles"];                          /- check that number of handles equal input, Needs fixed due to proctype input!
```
jonathonmcmurray: remove commented line

dmartinaquaq: Yep, will remove now.
Just had it in because it is a check that will need to be run but not sure how to get it working.
Need to check if any input comes back with null handles, so when you are looking at the results, it is known that an input was entered for a process that wasn't there.
Was maybe thinking of using a dictionary, will look into this.

code/processes/dqe.q
```diff
@@ -1,34 +1,25 @@
...
-  subscribe[];
+  .servers.startup[];                                                                                           /- Open connection to discovery
   }
 
-tableExists:{x in tables[]};                                  /-function to check for table, param is table name as a symbol
+tableexists:{x in tables[]};                                                                                    /- function to check for table, param is table name as a symbol
+
+runcheck:{[fn;vars;rs]                                                                                          /- function used to send other function to test processes
+  if[0<count s:{x where 0=x in raze exec procname,proctype from .servers.querydiscovery`ALL}[rs];
+    .lg.e[`process;string s," processes are not connectable"]];
```
jonathonmcmurray: Should we return early at this point? Seems like it this error msg gets output there's no point continuing with function

code/processes/dqe.q
```diff
@@ -1,34 +1,25 @@
...
+  h:.servers.opencon each exec hpup from .servers.querydiscovery`ALL where (procname in rs)|(proctype in rs);   /- check discovery and open handles to other processes using procname and/or proctype
```
jonathonmcmurray: do we need to query discovery twice? can we not just keep the result from the first time locally?

code/processes/dqe.q
dmartinaquaq: Thought it would be good to still get back results of processes that where entered correctly or would it be better to maybe set up an email alert to say that a config file has wrong input?

jonathonmcmurray: I don't follow... are the situations where you get no results & print an error here, but _do_ get handles in the next line?

dmartinaquaq: If an input process does not exist (or can't get connected too), a null will be returned when checking with discovery. So then no handle will be opened for it. As long as one of the inputs is correct, the function still runs and will return results for the correct processes. 
This was a problem when looking at the results as there was nothing to say that anything went wrong. We added this so that the incorrect inputs are returned so that the user knows that these are the ones that did not run.

jonathonmcmurray: > If an input process does not exist (or can't get connected too), a null will be returned when checking with discovery. So then no handle will be opened for it.

Yeah, so both this & the next line will fail? I'm still not really getting it, it seems to me like there are two checks for the same thing? Can you give me an example of something that would output the error on the first check, but _not_ error on the second?

jonathonmcmurray: Oh I think I see, this would be if you passed in e.g. `` `rdb1`rdb2`` but you only had `` `rdb1`` running, is that right?

code/processes/dqe.q
```diff
@@ -1,34 +1,27 @@
...
+  .servers.startup[];                                                                                           /- Open connection to discovery
   }
 
-tableExists:{x in tables[]};                                  /-function to check for table, param is table name as a symbol
+tableexists:{x in tables[]};                                                                                    /- function to check for table, param is table name as a symbol
+
+runcheck:{[fn;vars;rs]                                                                                          /- function used to send other function to test processes
+  discov:.servers.querydiscovery`ALL;
+  if[0<count s:{
+      $[1<count sf:x in raze exec procname,proctype from y;x where 0=sf;(enlist x) where 0=enlist sf]           /- check if processes exist and are valid, slight diff check if atom or list input
```
ThomasSmyth: Is this line just about working with lists? And if so will this go in its place?
```suggestion
      x where not sf:x in raze exec procname,proctype from y          /- check if processes exist and are valid, slight diff check if atom or list input
```

code/processes/dqe.q
```diff
@@ -1,34 +1,27 @@
...
-  subscribe[];
+  .servers.startup[];                                                                                           /- Open connection to discovery
   }
 
-tableExists:{x in tables[]};                                  /-function to check for table, param is table name as a symbol
+tableexists:{x in tables[]};                                                                                    /- function to check for table, param is table name as a symbol
+
+runcheck:{[fn;vars;rs]                                                                                          /- function used to send other function to test processes
+  discov:.servers.querydiscovery`ALL;
+  if[0<count s:{
```
ThomasSmyth:
```
  if[count s:rs where not rs in raze exec procname,proctype from discov;
    .lg.e[`process;(", "sv string s)," processes are not connectable"];
  ];
```

code/processes/dqe.q
dmartinaquaq: I has to work with atoms too as a single process can be passed in to be checked.
Example:
```
.dqe.runcheck[.dqe.tableexists;`quote;`rdb]
```
There is a problem with the where returning a type error when a list is not passed to it.

jonathonmcmurray: Can you just use `(),` to make it a list?

```q
q)(),`rdb
,`rdb
q)(),`rdb`hdb
`rdb`hdb
```

code/processes/dqe.q
```diff
@@ -1,34 +1,36 @@
...
+runcheck:{[fn;vars;rs]                                                                                          /- function used to send other function to test processes
+  fncheck:(` vs fn);
+  if [0b=b:fncheck[2] in key value .Q.dd[`;fncheck[1]];                                                         /- run check to make sure passed in function exists
+    .lg.e[`function;"Function ",(string fn)," doesn't exist"];
+    :()];
+
+  rs:(),rs;                                                                                                     /- set rs to a list
+  h:.dqe.gethandles[rs];                                                                                        /- check if processes exist and are valid
+  
+  missingproc:rs where not rs in h[`procname],h[`proctype];							/- check all process exist
```
jonathonmcmurray: tabs used on this line for comment alignment, every other line has space, so this displays inconsistently

code/processes/dqe.q
```diff
@@ -1,34 +1,36 @@
...
+  .servers.startup[];                                                                                           /- Open connection to discovery
   }
 
-tableExists:{x in tables[]};                                  /-function to check for table, param is table name as a symbol
+gethandles:{exec procname,proctype,w from .servers.SERVERS where (procname in x) | (proctype in x)};
+
+tableexists:{x in tables[]};                                                                                    /- function to check for table, param is table name as a symbol
+
+runcheck:{[fn;vars;rs]                                                                                          /- function used to send other function to test processes
+  fncheck:(` vs fn);
```
ThomasSmyth: Drop brackets.

code/processes/dqe.q
```diff
@@ -1,34 +1,36 @@
...
+  if [0b=b:fncheck[2] in key value .Q.dd[`;fncheck[1]];                                                         /- run check to make sure passed in function exists
```
ThomasSmyth: Can drop some square brackets:
```suggestion
  if[not b:fncheck[2] in key value .Q.dd[`;fncheck 1];                                                          /- run check to make sure passed in function exists
```

code/processes/dqe.q
```diff
@@ -1,34 +1,36 @@
...
+    .lg.e[`function;"Function ",(string fn)," doesn't exist"];
+    :()];
+
+  rs:(),rs;                                                                                                     /- set rs to a list
+  h:.dqe.gethandles[rs];                                                                                        /- check if processes exist and are valid
+  
+  missingproc:rs where not rs in h[`procname],h[`proctype];                                                     /- check all process exist
+  failedproc:{x,`failed}'[missingproc];
```
ThomasSmyth: This doesn't seem to do much.

dmartinaquaq: This stores the processes from the input that are not up or exist.
Right now it is only used for the error message but will be used in one of the next parts.

jonathonmcmurray: > This stores the processes from the input that are not up or exist

aren't those already stored in `missingproc` though?

dmartinaquaq: Get your point, this just adds failed to each for an upsert later.
I will remove it and replace it with a more useful expression later :)

code/processes/dqe.q
```diff
@@ -1,34 +1,35 @@
...
+    :()];
+
+  rs:(),rs;                                                                                                     /- set rs to a list
+  h:.dqe.gethandles[rs];                                                                                        /- check if processes exist and are valid
+  
+  missingproc:rs where not rs in h[`procname],h[`proctype];                                                     /- check all process exist
+  if[0<count missingproc;.lg.e[`process;(", "sv string missingproc)," processes are not connectable"]];
+
+  if[0=count h;.lg.e[`handle;"cannot open handle to any given processes"];:()];                                 /- check if any handles exist, if not exit function
+  ans:{[funct;vars;hand].async.postback[hand;(funct;vars);`.dqe.showresult]}[value fn;vars]'[h[`w]];            /- send function with variables down handle
```
ThomasSmyth: Return here instead, can skip declaring the variable `ans`

code/processes/dqe.q
```diff
@@ -1,34 +1,35 @@
...
+runcheck:{[fn;vars;rs]                                                                                          /- function used to send other function to test processes
+  fncheck:` vs fn;
+  if [not fncheck[2] in key value .Q.dd[`;fncheck 1];                                                           /- run check to make sure passed in function exists
+    .lg.e[`function;"Function ",(string fn)," doesn't exist"];
+    :()];
+
+  rs:(),rs;                                                                                                     /- set rs to a list
+  h:.dqe.gethandles[rs];                                                                                        /- check if processes exist and are valid
+  
+  missingproc:rs where not rs in h[`procname],h[`proctype];                                                     /- check all process exist
```
ThomasSmyth: Can grab both at the same time:
```suggestion
  missingproc:rs where not rs in h`procname`proctype;                                                           /- check all process exist
```
If that doesn't work may just need to use raze:
```
raze h`procname`proctype
```

code/processes/dqe.q
```diff
@@ -1,34 +1,34 @@
...
+    :()];
+
+  rs:(),rs;                                                                                                     /- set rs to a list
+  h:.dqe.gethandles[rs];                                                                                        /- check if processes exist and are valid
+  
+  missingproc:rs where not rs in raze h`procname`proctype;                                                      /- check all process exist
+  if[0<count missingproc;.lg.e[`process;(", "sv string missingproc)," processes are not connectable"]];
+
+  if[0=count h;.lg.e[`handle;"cannot open handle to any given processes"];:()];                                 /- check if any handles exist, if not exit function
+  ans:{[funct;vars;hand].async.postback[hand;(funct;vars);`.dqe.showresult]}[value fn;vars]'[h[`w]]             /- send function with variables down handle
```
ThomasSmyth: This may not actually return, can fix in next PR.

dmartinaquaq: I had tested it before and it seemed fine.
```
q).dqe.runcheck[`.dqe.tableexists;`quote;`rdb`hdb1`bbo`hfd]
2019.11.22D14:45:07.558743000|homer|dqe|dqe1|ERR|process|bbo, hfd processes are not connectable
,1b
,1b
q)1b
1b
```

PR #178, opened 2019-11-25
Add configuration for loading checks

tszwailam:
Added a loadcsv function for loading the csv
added a dqeconfig.csv for a sample csv

### DIFF
--- a/code/processes/dqe.q
+++ b/code/processes/dqe.q
@@ -14,6 +14,13 @@ readdqeconfig:{[file]
   c:.[0:;(("S****NNN";enlist",");file);{.lg.e["failed to load dqe configuration file: ",x]}];
  }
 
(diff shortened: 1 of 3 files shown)


code/processes/dqe.q
```diff
@@ -1,10 +1,39 @@
...
 init:{
   .lg.o[`init;"searching for servers"];
   .servers.startup[];                                                                                           /- Open connection to discovery
   }
 
+configtable:(
+
+
+  action:`symbol$();              // type of check
+
```
jonathonmcmurray: there's a lot of empty lines in this schema, is that necessary?

more debatable, I think most of these column names are reasonably self-explanatory, could this just be a one-liner schema perhaps?

svetsolo: thanks, going to delete empty lines and make it one line

PR #183, opened 2019-12-12
Adding check for a null server

CatherineHigg:


code/processes/gateway.q
```diff
@@ -393,14 +393,17 @@ getserverids:{[att]
    activeservers:exec distinct servertype from .gw.servers where active;
    //list of all servers
    allservers:exec distinct servertype from .gw.servers;
+   activeserversmsg:". Available servers include: "," " sv string activeservers;
```
ThomasSmyth: I think a comma should be added for clarity, just like on line 405:
```suggestion
   activeserversmsg:". Available servers include: ",", " sv string activeservers;
```

ThomasSmyth:
Not merging to master so no need to wait for approval I think :+1:

PR #184, opened 2019-12-12
call extended logging functionality on log messages

EllenMooney95:
Any extended logging functionality defined in .lg.ext should be executed as part of logging

### DIFF
--- a/torq.q
+++ b/torq.q
@@ -202,9 +202,9 @@ pubmap:@[value;`pubmap;`ERROR`ERR`INF`WARN!1 1 0 1]
 // Log a message
 l:{[loglevel;proctype;proc;id;message;dict]
 	if[0 < redir:`int$(0w 1 `onelog in key .proc.params)&0^outmap[loglevel];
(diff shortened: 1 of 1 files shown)


torq.q
```diff
@@ -201,7 +201,7 @@ pubmap:@[value;`pubmap;`ERROR`ERR`INF`WARN!1 1 0 1]
 
 // Log a message
 l:{[loglevel;proctype;proc;id;message;dict]
-	$[0 < redir:`int$(0w 1 `onelog in key .proc.params)&0^outmap[loglevel];
+	if[0 < redir:`int$(0w 1 `onelog in key .proc.params)&0^outmap[loglevel];
 		neg[redir] .lg.format[loglevel;proctype;proc;id;message];
 		ext[loglevel;proctype;proc;id;message;dict]];
```
jonnypress: I think you want to call the ext function in either case - if 0<redir or not. So can you change that, and fix up the indenting

EllenMooney95: so if 0<redir we want to run both of the above, if not, we just want to run the extended logging?

PR #187, opened 2019-12-13
Dqe function dir added and loaded into dqe process

dmartinaquaq:
Add dqe dir to `code/dqe`
All check functions added as separate files to new dir
Remove check functions from dqe code
Load check function dir into dqe process

### DIFF
--- a/code/dqe/chkslowsub.q
+++ b/code/dqe/chkslowsub.q
@@ -1,4 +1,5 @@
-.dqe.chkslowsub:{[threshold]                                                                                    /- function to check for slow subscribers
+\d .dqe
+chkslowsub:{[threshold]                                                                                         /- function to check for slow subscribers
   .lg.o[`dqe;"Checking for slow subscribers"];
   overlimit:(key .z.W) where threshold<sum each value .z.W;
   $[0=count overlimit;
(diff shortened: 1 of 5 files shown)


code/dqe/chkslowsub.q
```diff
@@ -0,0 +1,7 @@
+.dqe.chkslowsub:{[threshold]                                                                                    /- function to check for slow subscribers
```
jonathonmcmurray: in TorQ we generally use `\d .dqe` etc. rather than fully qualified names

code/processes/dqe.q
```diff
@@ -5,6 +5,7 @@ configcsv:@[value;`.dqe.configcsv;first .proc.getconfigfile["dqeconfig.csv"]];
 init:{
   .lg.o[`init;"searching for servers"];
   .servers.startup[];                                                                                           /- Open connection to discovery
+  .proc.loaddir[(getenv `TORQHOME),"/code/dqe"];                                                                /- load in dqe function code directory
```
jonathonmcmurray: as long as configuration is right (I think `.proc.loadprocesscode` needs set to `1b`), this should happen automatically

PR #188, opened 2019-12-16
Checks for correct attribute parameter types

nicolew02:
Inform user of the correct parameter types if a wrong type is passed in.
Checks are placed in the `getserverids` function and include checks for:
- If servertype is given as a symbol list 
- If servertype attributes are of the correct types compared to a correctAttTypes dictionary

### DIFF
--- a/code/processes/gateway.q
+++ b/code/processes/gateway.q
@@ -313,7 +313,7 @@ getserversinitial:{[req;att]
  att:(where all each (key req) in/: key each att)#att;
 
  if[not count att;'"getservers: no servers report all requested attributes"];
-
+ 
  /- calculate where each of the requirements is in each of the attribute sets
  s:update serverid:key att from value req in'/: (key req)#/:att;
 
@@ -388,6 +388,10 @@ getserverscross:{[req;att;besteffort]
 getserverids:{[att]
   if[99h<>type att;
    // its a list of servertypes e.g. `rdb`hdb
+   // check if user attributes are a symbol list
+   if[11h<>type att;
+   '" Servertype should be given as either a dictionary(type 99h) or a symbol list (11h)"
+   ];
    servertype:distinct att,();
    //list of active servers
    activeservers:exec distinct servertype from .gw.servers where active;
@@ -409,6 +413,12 @@ getserverids:{[att]
   ];
 
   // its a dictionary of attributes
+  // check if attribute types are correct types
+  correctAttTypes:`dates`servertypes`tables!(14h;11h;11h);
+  w:(correctAttTypes [key att]=type each att);
+  if[not all{[x;y] y each key x}[att;w];
+  '("Wrong function parameter types provided.", " ",(raze " " sv string (where not w))," parameter(s) need type(s) ",raze " " sv string correctAttTypes where not w),"h"
+  ];
 
   serverids:$[`servertype in key att; 
   raze getserveridstype[delete servertype from att] each (),att`servertype; 
--- a/code/processes/gateway.q
+++ b/code/processes/gateway.q
@@ -313,7 +313,7 @@ getserversinitial:{[req;att]
  att:(where all each (key req) in/: key each att)#att;
 
  if[not count att;'"getservers: no servers report all requested attributes"];
- 
+
  /- calculate where each of the requirements is in each of the attribute sets
  s:update serverid:key att from value req in'/: (key req)#/:att;
 
@@ -413,12 +413,12 @@ getserverids:{[att]
   ];
 
   // its a dictionary of attributes
-  // check if attribute types are correct types
-  correctAttTypes:`dates`servertypes`tables!(14h;11h;11h);
-  w:(correctAttTypes [key att]=type each att);
-  if[not all{[x;y] y each key x}[att;w];
-  '("Wrong function parameter types provided.", " ",(raze " " sv string (where not w))," parameter(s) need type(s) ",raze " " sv string correctAttTypes where not w),"h"
-  ];
+    // check if attribute types are correct types
+    correctAttTypes:`dates`servertypes`tables!14 11 11h;
+    w:correctAttTypes[key att]=type each att;
+    if[not all{[x;y] y each key x}[att;w];
+    '("Wrong function parameter types provided.", " ",(raze " " sv string (where not w))," parameter(s) need type(s) ",raze " " sv string correctAttTypes where not w),"h"
+    ];
 
   serverids:$[`servertype in key att; 
   raze getserveridstype[delete servertype from att] each (),att`servertype; 
--- a/code/processes/gateway.q
+++ b/code/processes/gateway.q
@@ -388,10 +388,10 @@ getserverscross:{[req;att;besteffort]
 getserverids:{[att]
   if[99h<>type att;
    // its a list of servertypes e.g. `rdb`hdb
-     // check if user attributes are a symbol list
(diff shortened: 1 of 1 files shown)


code/processes/gateway.q
```diff
@@ -313,7 +313,7 @@ getserversinitial:{[req;att]
  att:(where all each (key req) in/: key each att)#att;
 
  if[not count att;'"getservers: no servers report all requested attributes"];
-
+
```
ThomasSmyth: :eyes:

code/processes/gateway.q
```diff
@@ -388,6 +388,10 @@ getserverscross:{[req;att;besteffort]
 getserverids:{[att]
   if[99h<>type att;
    // its a list of servertypes e.g. `rdb`hdb
+   // check if user attributes are a symbol list
+   if[11h<>type att;
+   '" Servertype should be given as either a dictionary(type 99h) or a symbol list (11h)"
```
ThomasSmyth: Fix indent (+2 spaces).

code/processes/gateway.q
```diff
@@ -409,6 +413,12 @@ getserverids:{[att]
   ];
 
   // its a dictionary of attributes
+  // check if attribute types are correct types
+  correctAttTypes:`dates`servertypes`tables!(14h;11h;11h);
+  w:(correctAttTypes [key att]=type each att);
+  if[not all{[x;y] y each key x}[att;w];
+  '("Wrong function parameter types provided.", " ",(raze " " sv string (where not w))," parameter(s) need type(s) ",raze " " sv string correctAttTypes where not w),"h"
```
ThomasSmyth: Fix indent (+2 spaces). Should the first `where not w` be used in indexing? Seems like a stringed Boolean list will be returned here.

code/processes/gateway.q
```diff
@@ -409,6 +413,12 @@ getserverids:{[att]
   ];
 
   // its a dictionary of attributes
+  // check if attribute types are correct types
+  correctAttTypes:`dates`servertypes`tables!(14h;11h;11h);
+  w:(correctAttTypes [key att]=type each att);
```
ThomasSmyth: Remove brackets + don't use space before args.
```suggestion
  w:correctAttTypes[key att]=type each att;
```

code/processes/gateway.q
```diff
@@ -409,6 +413,12 @@ getserverids:{[att]
   ];
 
   // its a dictionary of attributes
+  // check if attribute types are correct types
+  correctAttTypes:`dates`servertypes`tables!(14h;11h;11h);
```
ThomasSmyth:
```suggestion
  correctAttTypes:`dates`servertypes`tables!14 11 11h;
```

code/processes/gateway.q
nicolew02: `where not w` should return
`,`dates`

code/processes/gateway.q
nicolew02: oops

code/processes/gateway.q
```diff
@@ -409,6 +413,12 @@ getserverids:{[att]
...
+  w:(correctAttTypes [key att]=type each att);
+  if[not all{[x;y] y each key x}[att;w];
```
ThomasSmyth: Lambda not needed here and pretty sure `each` not required either:
```suggestion
  if[not all w key att;
```

code/processes/gateway.q
```diff
@@ -409,6 +413,12 @@ getserverids:{[att]
   ];
 
   // its a dictionary of attributes
+    // check if attribute types are correct types
+    correctAttTypes:`dates`servertypes`tables!14 11 11h;
+    w:correctAttTypes[key att]=type each att;
+    if[not all{[x;y] y each key x}[att;w];
+    '("Wrong function parameter types provided.", " ",(raze " " sv string (where not w))," parameter(s) need type(s) ",raze " " sv string correctAttTypes where not w),"h"
```
ThomasSmyth: Great, run through with some dummy args and that checks out. Last fix for this line, need to fix the indenting and should be able to drop `raze` by moving some brackets around. Can also avoid manually adding `"h"` by using `.Q.s1`.
```suggestion
      '"Wrong function parameter types provided.", " ",(" "sv string where not w)," parameter(s) need type(s) ",.Q.s1 correctAttTypes where not w
```

code/processes/gateway.q
claire-sloan: Can combine:
```if[type[att]in 99 11h;'"Error..."];```

File: code/processes/gateway.q [hunk 8]
```diff
@@ -413,12 +413,12 @@ getserverids:{[att]
...
 
   // its a dictionary of attributes
-  // check if attribute types are correct types
-  correctAttTypes:`dates`servertypes`tables!(14h;11h;11h);
-  w:(correctAttTypes [key att]=type each att);
-  if[not all{[x;y] y each key x}[att;w];
-  '("Wrong function parameter types provided.", " ",(raze " " sv string (where not w))," parameter(s) need type(s) ",raze " " sv string correctAttTypes where not w),"h"
-  ];
+    // check if attribute types are correct types
+    correctAttTypes:`dates`servertypes`tables!14 11 11h;
```
CatherineHigg: The dictionary passed in will have `servertype rather than `servertypes, in the key so just drop the 's' here so it indexes correctly. Same for `dates I think.

code/processes/gateway.q
```diff
@@ -409,6 +413,12 @@ getserverids:{[att]
   ];
 
   // its a dictionary of attributes
+    // check if attribute types are correct types
+    correctAttTypes:`dates`servertypes`tables!14 11 11h;
```
CatherineHigg: Concerned this won't recognise attributes passed in as atoms so might want to adjust the dictionary to allow for this.

ThomasSmyth: Use `abs` in here somewhere to account for lists and atoms?

code/processes/gateway.q
```diff
@@ -409,6 +413,12 @@ getserverids:{[att]
   ];
 
   // its a dictionary of attributes
+    // check if attribute types are correct types
+    correctAttTypes:`date`servertype`tables!14 11 11h;
+    w:correctAttTypes[key att]=abs type each att;
+    if[not all w key att;
+    '"Wrong function parameter types provided.", " ",(" "sv string where not w)," parameter(s) need type(s) ",.Q.s1 correctAttTypes where not w
```
ThomasSmyth: Where's the extra indent here and on line 393? :cry:

nicolew02: Sorry I thought you meant the earlier if check, will get it done now!

code/processes/gateway.q
```diff
@@ -388,6 +388,10 @@ getserverscross:{[req;att;besteffort]
 getserverids:{[att]
   if[99h<>type att;
    // its a list of servertypes e.g. `rdb`hdb
+   // check if user attributes are a symbol list
+   if[not type[att]in 99 11h;
```
ThomasSmyth: @claire-sloan is this how you wanted these combined? Second check for 99h is redundant in this scenario.

CatherineHigg: The function still needs a way to differentiate between `att` being a list of symbols and it being a dictionary so I think we may need to keep them separate.

ThomasSmyth: Yeah, I think this check should just be `if[not type[att]=11h;`

PR #189, opened 2019-12-16
Remove double running of unit tests

jonathonmcmurray:
The TorQ version of k4unit has a modification which stores the results of tests that are run. This means that tests are run twice, once to measure time/memory usage & once to store the output; after discussing with @ThomasSmyth we think this is a bug & it can cause unexpected behaviour due to tests running twice

@jonnypress do you know why this was added, and if it's needed? This PR simply removes, but perhaps we can rework it to do both in one run if necessary


ThomasSmyth:
If it was possible to run once, time it and keep the result that may be worthwhile. But otherwise this should be removed as the behaviour is unexpected IMO.

jonathonmcmurray:
I think this is actually already (slightly) broken in the case `repeat<>1` - in that case, k4unit injects a `do[...]` loop meaning there would be no return

jonathonmcmurray:
I have a modification that does allow storing of the result as well as returning, however I realised another issue with this - if the return of first test is, e.g. an int, the `KUrunresults` table gets `res` column typed as `int` after first insert. If next test returns something else, we get a `'type` on the upsert.

I'm struggling to see the use case for manually inspecting the return of unit tests, I think we should just get rid of this.

jonathonmcmurray:
@jonnypress @ThomasSmyth any objects to merging this? as mentioned above, think there are multiple bugs with current code

PR #190, opened 2019-12-18
Added kill -9 to stop function to ensure processes are stopped

jlogan540:
Improve the stop function in torq.sh to ensure that a process stops. Check if the function is still up after kill -15, if it is sleep, check again and perform kill -9 on active process.

### DIFF
--- a/torq.sh
+++ b/torq.sh
@@ -119,6 +119,13 @@ stop() {
     port=$(($(eval echo \$"$(getfield "$procno" port)")))
     eval "kill -15 `lsof -i :$port -sTCP:LISTEN | awk '{if(NR>1)print $2}'`"
   fi
+  if [[ ! -z $(findproc "$1") ]]; then                                                              # check process still running
+    sleep 1
+    if [[ ! -z $(findproc "$1") ]]; then
+      echo "$(date '+%H:%M:%S') | $1 did not shut down correctly. Killing process with -9..."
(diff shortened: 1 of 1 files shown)


frenzy21:
Needs unit tests to be added for this btw.

jonnypress:
Can we get this one closed out with Thomas's suggestion please?

torq.sh
```diff
@@ -139,7 +144,9 @@ checkinput() {
   PROCS=$(awk -F, '{if(NR>1) print $4}' "$CSVPATH")                                                 # get all process names from csv
   avail=()
   for i in $input; do
-    if [[ $(echo "$PROCS" | grep -w "$i") ]]; then                                                  # check input process is valid
+    if [[ $i == "-force" ]]; then                                                                   # check for force flag
+      echo "Force stop has been set"
```
ThomasSmyth: As per offline convo:
```suggestion
      :
```

torq.sh
```diff
@@ -111,13 +111,18 @@ summary() {
  }
 
 stop() {
+  if [[ $(echo ${BASH_ARGV[*]} | grep -e -force) ]]; then                                           # check for force flag
+    cmd="kill -9"
```
ThomasSmyth: Add line below:
```
echo "Force stop has been set"
```

PR #193, opened 2019-12-19
Datadog Integration

PeterSmiley:
Integration of Datadog to run checks on TorQ processes and visual monitoring via a dashboard.
• setupdatadog.sh file for user to configure datadog daemon files and to set the datadog port using command line argument.
• datachecks.q which configures the checks so they can be sent to datadog
• Example_TorQ_Monitoring_Dashboard.json - a JSON file which can be used to import an example monitor
• Monitors folder containing example monitors.
• Datadog.q in code/common which defines the checks in process.


code/common/datadog.q
```diff
@@ -0,0 +1,36 @@
...
+sendMetric:{[metric_name;metric_value] system"bash -c \"echo  -n '",metric_name,":",(string metric_value),"|g|#shell' > /dev/udp/127.0.0.1/",(string dogstatsd_port),"\"";}
+
+.dg.sendEvent:{[event_title;event_text;tags;alert_type] system"event_title=",event_title,"; event_text=","\"",event_text,"\"","; tags=","\"",$[0h=type tags;"#",("," sv tags);"#",tags],"\"",";alert_type=",alert_type,"; ","echo \"_e{${#event_title},${#event_text}}:$event_title|$event_text|#$tags|t:$alert_type\" |nc -4u -w0 127.0.0.1 ",string dogstatsd_port;}
+
+\d .
+
+//override default .lg.ext functionality to send error and warning events to datadog
+
+.lg.ext:{[loglevel;proctype;proc;id;message;dict]
+ if[loglevel in `ERR`WARN;.dg.sendEvent[string proc;message;string proctype;]$[loglevel=`ERR;"error";"warning"]]}
```
jonathonmcmurray: Should we perhaps only overwrite if this is enabled? Maybe some `.dg.enabled` config variable (probably off by default too)

Also, what if someone already has a `.lg.ext` defined?

datadog/Example_TorQ_Monitoring_Dashboard.json
```diff
@@ -0,0 +1 @@
...mats":[{"comparator":">","value":0,"palette":"white_on_green"},{"comparator":"<","value":1,"palette":"white_on_red"}]}],"title":"Data Check - feed1","title_size":"16","title_align":"left","autoscale":true,"precision":2},"layout":{"x":65,"y":0,"width":17,"height":8}},{"id":11,"definition":{"type":"query_value","requests":[{"q":"default_zero(avg:process_hdb2_check{*})","aggregator":"last","conditional_formats":[{"comparator":">","value":0,"palette":"white_on_green"},{"comparator":"<","value":1,"palette":"white_on_red"}]}],"title":"Data Check - hdb2","title_size":"16","title_align":"left","autoscale":true,"precision":2},"layout":{"x":65,"y":45,"width":17,"height":8}},{"id":12,"definition":{"type":"query_value","requests":[{"q":"default_zero(avg:process_gateway1_check{*})","aggregator":"last","conditional_formats":[{"comparator":">","value":0,"palette":"white_on_green"},{"comparator":"<","value":1,"palette":"white_on_red"}]}],"title":"Data Check - gateway1","title_size":"16","title_align":"left","autoscale":true,"precision":2},"layout":{"x":65,"y":54,"width":17,"height":8}}],"template_variables":[],"layout_type":"free","is_read_only":false,"notify_list":[],"dashboard_id":"era-ztf-vye"}
```
jonathonmcmurray: is this file meant to be read/edited manually? if so, can prettify the layout with some JSON formatter

datadog/runchecks.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+echo `date`;
+. ~/.bashrc;
+~/q/l32/q ./datadog/datachecks.q
```
jonathonmcmurray: these hardcoded paths seem like they could be problematic

code/common/datadog.q
```diff
@@ -0,0 +1,36 @@
...
+$[`ddconfig.txt in key hsym `$getenv[`KDBAPPCONFIG];value each read0 ddconfigfile;dogstatsd_port:8125]
+
+//Functions are set to return 1b or 0b based on the health of the service.
+//Default check returns 1b from each process to indicate process is up and can be queried.
+//These checks run every time crontab/timer is run.
+
+handlers:(`symbol$())!()
+
+is_ok:{[x]
+        f:$[@[{[x].proc.proctype in key .dg.handlers};`;0b];
```
ThomasSmyth: This indent is extremely agressive, please tone down to 2 spaces.

I also don't really understand why an error trap is need here, I don't see a scenario where `.proc.proctype` and `.dg.handlers` won't exist at this point. Please correct me if I'm wrong. Can also make this a one liner now that it's slimmed up.
```suggestion
        f:$[.proc.proctype in key handlers;.dg.handlers .proc.proctype;.dg.default_is_ok];
```
I'm also inclined to say that `default_is_ok` doesn't need to be defined and you can use a lambda in its place.

code/common/datadog.q
```diff
@@ -0,0 +1,36 @@
...
+//functions to send metrics and events to datadog from TorQ processes
+sendMetric:{[metric_name;metric_value] system"bash -c \"echo  -n '",metric_name,":",(string metric_value),"|g|#shell' > /dev/udp/127.0.0.1/",(string dogstatsd_port),"\"";}
+
+.dg.sendEvent:{[event_title;event_text;tags;alert_type] system"event_title=",event_title,"; event_text=","\"",event_text,"\"","; tags=","\"",$[0h=type tags;"#",("," sv tags);"#",tags],"\"",";alert_type=",alert_type,"; ","echo \"_e{${#event_title},${#event_text}}:$event_title|$event_text|#$tags|t:$alert_type\" |nc -4u -w0 127.0.0.1 ",string dogstatsd_port;}
+
+\d .
+
+//override default .lg.ext functionality to send error and warning events to datadog
+
+.lg.ext:{[loglevel;proctype;proc;id;message;dict]
```
ThomasSmyth: I really don't think the `.lg` namespace should be updated here, was this in the requirements?

code/common/datadog.q
```diff
@@ -0,0 +1,36 @@
...
+handlers:(`symbol$())!()
+
+is_ok:{[x]
+        f:$[@[{[x].proc.proctype in key .dg.handlers};`;0b];
+                .dg.handlers[.proc.proctype];
+                .dg.default_is_ok];
+        @[f;`;0b]
+        }
+
+default_is_ok:{[x]1b}
```
ThomasSmyth: If there is no intention of having this overwritten then it should just be a lambda in `is_ok`.

code/common/datadog.q
```diff
@@ -0,0 +1,36 @@
...
+
+//functions to send metrics and events to datadog from TorQ processes
+sendMetric:{[metric_name;metric_value] system"bash -c \"echo  -n '",metric_name,":",(string metric_value),"|g|#shell' > /dev/udp/127.0.0.1/",(string dogstatsd_port),"\"";}
```
ThomasSmyth: Was opening a handle to the file/location ever tested as an alternative?

code/common/datadog.q
```diff
@@ -0,0 +1,36 @@
...
+
+.dg.sendEvent:{[event_title;event_text;tags;alert_type] system"event_title=",event_title,"; event_text=","\"",event_text,"\"","; tags=","\"",$[0h=type tags;"#",("," sv tags);"#",tags],"\"",";alert_type=",alert_type,"; ","echo \"_e{${#event_title},${#event_text}}:$event_title|$event_text|#$tags|t:$alert_type\" |nc -4u -w0 127.0.0.1 ",string dogstatsd_port;}
```
ThomasSmyth: This is one of the biggest lines I've seen in a while, should be broken up for readability.
```suggestion
.dg.sendEvent:{[event_title;event_text;tags;alert_type]
  system"event_title=",event_title,"; event_text=","\"",event_text,"\"","; tags=","\"#",$[0h=type tags;","sv tags;tags],"\"",";alert_type=",alert_type,"; ","echo \"_e{${#event_title},${#event_text}}:$event_title|$event_text|#$tags|t:$alert_type\" |nc -4u -w0 127.0.0.1 ",string dogstatsd_port;
 }
```

datadog/datachecks.q
```diff
@@ -0,0 +1,51 @@
+//DATADOG CHECKS
+//Script for running checks manually/via crontab
+//Makes dictionary of defaults and uses .Q.opt to refer to command values passed by its key.
+o:.Q.def[`user`pass`timeout`init`noexit!(`admin;`admin;100;1b;0b);.Q.opt[.z.x]]
```
ThomasSmyth: :eyes:
This not making use of TorQ input handling?

datadog/datachecks.q
```diff
@@ -0,0 +1,51 @@
...
+
+ddconfigfile:hsym `$getenv[`KDBAPPCONFIG],"/ddconfig.txt"
+
+//sets dogstatsd_port to the port defined by ddconfigfile
+$[`ddconfig.txt in key hsym `$getenv[`KDBAPPCONFIG];value each read0 ddconfigfile;dogstatsd_port:8125]
```
ThomasSmyth: Same as before, I think this should reference `ddconfigfile`.

datadog/datachecks.q
```diff
@@ -0,0 +1,51 @@
+//DATADOG CHECKS
+//Script for running checks manually/via crontab
+//Makes dictionary of defaults and uses .Q.opt to refer to command values passed by its key.
+o:.Q.def[`user`pass`timeout`init`noexit!(`admin;`admin;100;1b;0b);.Q.opt[.z.x]]
+
+ddconfigfile:hsym `$getenv[`KDBAPPCONFIG],"/ddconfig.txt"
```
ThomasSmyth: Use TorQ config finder if available.

datadog/datachecks.q
```diff
@@ -0,0 +1,51 @@
...
+
+//sets dogstatsd_port to the port defined by ddconfigfile
+$[`ddconfig.txt in key hsym `$getenv[`KDBAPPCONFIG];value each read0 ddconfigfile;dogstatsd_port:8125]
+
+// Send To Datadog function - takes a non string value and a stringed name.
+//Will send the value received (1 or 0) and the process name (hdb etc)
+.datadog.sendMetric:{[metric_name;metric_value] system"bash -c \"echo  -n '",metric_name,":",(string metric_value),"|g|#shell' > /dev/udp/127.0.0.1/",(string dogstatsd_port),"\"";}
+
+.datadog.sendEvent:{[event_title;event_text;tags;alert_type] system"event_title=",event_title,"; event_text=","\"",event_text,"\"","; tags=","\"",$[0h=type tags;"#",("," sv tags);"#",tags],"\"",";alert_type=",alert_type,"; ","echo \"_e{${#event_title},${#event_text}}:$event_title|$event_text|#$tags|t:$alert_type\" |nc -4u -w0 127.0.0.1 ",string dogstatsd_port;}
```
ThomasSmyth: Is this a duplicated? And if so why? Can we define these functions in one place only?

datadog/datachecks.q
```diff
@@ -0,0 +1,51 @@
...
+
+
+//Creates the torq summary table without the pipes
+.datadog.getprocess:{[x]
+        {[x]flip (((`TIME`PROCESS`STATUS`PID`PORT!"TSSII")key[x]))$x} {[x] {[x](`$x[;0])! flip 1_ flip[x]} trim ("*****"; "|")0:x} system "./torq.sh summary"
```
ThomasSmyth: Lots of brackets! I think this can probably be simplified + it needs more comments to explain what's going on.

datadog/datachecks.q
```diff
@@ -0,0 +1,51 @@
...
+ }
+
+//Names of processes to be monitored to be edited depending on monitoring needs
+.datadog.monitorprocess:()
+//.datadog.monitorprocess:`tickerplant`wdb`rdb`hdb
```
ThomasSmyth: Remove if not in use.

datadog/datachecks.q
```diff
@@ -0,0 +1,51 @@
...
+//state either 1b or 0b (with error trap) calling sendMetric.
+.datadog.check:{[x;o]state:.[.datadog.sendcheck;(o;x);{[x]0b}];.datadog.sendMetric[.datadog.createeventname[x];state]}
+
+//Checks each of the processes on the monitor list
+.datadog.checkall:{[x;o].datadog.check[;o] each x}
+
+//Retrieves the processes to be monitored with summary data. 
+.datadog.init:{[x]
+        .datadog.allprocesslist:.datadog.getprocess[];
+	.datadog.monitoredprocesslist:select from .datadog.allprocesslist where any PROCESS like/:{[x]string[x],"*"}'[.datadog.monitorprocess];
```
ThomasSmyth: Inconsistent indenting.

datadog/datachecks.q
```diff
@@ -0,0 +1,51 @@
...
+.datadog.check:{[x;o]state:.[.datadog.sendcheck;(o;x);{[x]0b}];.datadog.sendMetric[.datadog.createeventname[x];state]}
+
+//Checks each of the processes on the monitor list
+.datadog.checkall:{[x;o].datadog.check[;o] each x}
+
+//Retrieves the processes to be monitored with summary data. 
+.datadog.init:{[x]
+        .datadog.allprocesslist:.datadog.getprocess[];
+	.datadog.monitoredprocesslist:select from .datadog.allprocesslist where any PROCESS like/:{[x]string[x],"*"}'[.datadog.monitorprocess];	
+        $[count .datadog.monitorprocess;.datadog.checkall[.datadog.monitoredprocesslist;o];.datadog.checkall[.datadog.allprocesslist;o]]
```
ThomasSmyth: Can remove a lot of duplication here by use the conditional to pick between the first args only.
```suggestion
        .datadog.checkall[.datadog $[count .datadog.monitorprocess;`monitoredprocesslist;`allprocesslist]]o
```

datadog/datachecks.q
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+//if init is set to true run init
+if[o[`init];@[.datadog.init;`;1]]
```
ThomasSmyth: Should this be a Boolean?
```suggestion
if[o`init;@[.datadog.init;`;1b]]
```
Also `o` is a fairly undescriptive name if that is a global.

datadog/runchecks.sh
ThomasSmyth: What is this file? The version and location of q should not be hardcoded as it will differ between hosts.

datadog/setupdatadog.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+#Set the enviroment for TORQHOME
+cd ..
```
ThomasSmyth: Do not change directory, use environment variables to find full paths.

datadog/setupdatadog.sh
```diff
@@ -0,0 +1,68 @@
...
+source setenv.sh
```
ThomasSmyth: Probably this:
```suggestion
source $TORQHOME/setenv.sh
```

datadog/setupdatadog.sh
```diff
@@ -0,0 +1,68 @@
...
+
+if [ -z $1 ]; then
+  echo "no argument provided; defaulting dogstatsd_port to 8125"
+  dogstatsd_port=8125
```
ThomasSmyth: Should this port be defined in `setenv.sh`?

datadog/setupdatadog.sh
```diff
@@ -0,0 +1,68 @@
...
+else
+  dogstatsd_port=$1
+fi
+
+ddconfigfile=${KDBAPPCONFIG}/ddconfig.txt
+echo "dogstatsd_port:"$dogstatsd_port > $ddconfigfile
+
+#File path of the process.yaml file
+filename1=/etc/datadog-agent/conf.d/process.d/process.yaml
```
ThomasSmyth: `filename1` is not a very descriptive variable name, should be renamed to something more appropriate.

datadog/setupdatadog.sh
```diff
@@ -0,0 +1,68 @@
...
+
+#The process.csv is used to check what processes are to be added to the process.yaml file
+input=${KDBAPPCONFIG}/process.csv
+
+#If file doesn't exist create it and remove all restriction. Add first line "instances:". If datadog column exists in process.csv take processes marked for monitoring
+if [[ ! -f $filename1 ]];then
+  sudo touch $filename1
+  sudo chmod 777 $filename1
+  echo "instances:" >> $filename1
+  echo "Process.yaml file created"
```
ThomasSmyth: The file name should not be capitalised.

datadog/setupdatadog.sh
```diff
@@ -0,0 +1,68 @@
...
+      printf -- "- name: `echo "$line" | tr a-z A-Z`\n  search_string: ['$line']\n  exact_match: False\n" >> $filename1
+    done
+    echo "All processes added to process.yaml from $input"  
+  fi
+else
+  echo "$input already exists, no changes were made to avoid duplication."
+fi
+
+#File path of the datadog.yaml file
+filename2=/etc/datadog-agent/datadog.yaml
```
ThomasSmyth: Again rename to something more appropriate.

datadog/setupdatadog.sh
```diff
@@ -0,0 +1,68 @@
...
+sudo touch $filename2
+sudo chmod 777 $filename2
+if grep -qFx "use_dogstatsd: true" $filename2 ; then
+  echo "Datadog.yaml file has port enabled, no changes were made to avoid duplication."
+else
+  sudo echo "use_dogstatsd: true
+dogstatsd_port: $dogstatsd_port
+
+process_config:
+ enabled: \"true\"" >> $filename2
```
ThomasSmyth: Inconsistent indent.

datadog/setupdatadog.sh
```diff
@@ -0,0 +1,68 @@
...
+  echo "Datadog.yaml file is now enabled to send data through port $dogstatsd_port"
+fi
+
+#Create the crontab to run runchecks.sh at specified intervals.
+echo "Creating crontab if it doesn't already exist"
+crontab -l | grep -q 'runchecks'  && echo 'Crontab already exists.' || (crontab -l 2>/dev/null; echo " * * * * * cd $TORQHOME/; . $TORQHOME/datadog/runchecks.sh") | crontab -
```
ThomasSmyth: Is there a command that can crate `crontab` automatically when you need to add to it and it doesn't exist? Worth looking up.

code/common/datadog.q
jonathonmcmurray: alternatively might be possible to use netcat (`nc` command) to make something a bit cleaner?

code/common/datadog.q
EllenMooney95: I think we can wrap this inside some kind of .dg.init function, so it only defines this if the user calls it?

datadog/setupdatadog.sh
jonathonmcmurray: I'd also ask the user before modifying their crontab - there are other schedulers available etc., so I think it might not always be the right option depending on the system set up (although I wouldn't worry about adding support for any other schedulers)

code/common/datadog.q
EllenMooney95: netcat doesn't work well for the sendMetric functionality. Agree about disabling for other platforms though

datadog/Example_TorQ_Monitoring_Dashboard.json
PeterSmiley: The contents are just pasted in, or the file is uploaded to the website,  and then any changes to the dash board can be done on the website.

datadog/setupdatadog.sh
PeterSmiley: The layout in the datadog.yaml file is important here for it to work. WIthout the indent  before enabled the process_config is not set to "true". Is there another way to ensure this indent is here if indenting is removed here?

ThomasSmyth: If you update the indent in this bash file how does that have a knock on effect to the yaml file? To be clear this is the change I am suggesting:
```suggestion
  enabled: \"true\"" >> $filename2
```

PeterSmiley: In the .yaml file the "enabled:"true" has to be on a new line and indented like:
process_config:
 enabled: "true"

If I remove the space in the setupdatadog.sh file echo it becomes:
process_config:
enabled: "true"

Which causes problem for the datadog daemon.

ThomasSmyth: Is everything from line 58 to 61 sent to `$filename2`? If so then that is very unclear. At the very least that should be wrapped in quotes to make it obvious but the bigger issue is why is the config file modified like this? This seems ripe for being a major pain going forward.

datadog/setupdatadog.sh
```diff
@@ -0,0 +1,68 @@
...
+  fi
+else
+  echo "$input already exists, no changes were made to avoid duplication."
+fi
+
+#File path of the datadog.yaml file
+filename2=/etc/datadog-agent/datadog.yaml
+
+#If file doesn't exist make it and remove restrictions. Add necessary lines and save file.
+sudo touch $filename2
```
ThomasSmyth: `sudo` :eyes:
Perhaps this should be removed.

code/common/datadog.q
```diff
@@ -0,0 +1,36 @@
+// DATADOG CHECKS
+
+//Create datadog namespace
+\d .dg
+
+ddconfigfile:@[value;`ddconfigfile;hsym `$getenv[`KDBAPPCONFIG],"/ddconfig.txt"]
```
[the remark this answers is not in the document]
EllenMooney95: @PeterSmiley I think you can use .proc.getconfigfile or something for this (Thomas correct me if I'm wrong)

code/common/datadog.q
```diff
@@ -0,0 +1,36 @@
...
+
+//sets dogstatsd_port to the port defined by ddconfigfile
+$[`ddconfig.txt in key hsym `$getenv[`KDBAPPCONFIG];value each read0 ddconfigfile;dogstatsd_port:8125]
```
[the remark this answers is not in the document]
PeterSmiley: Sorry not quite sure what you mean - get rid of the hardcoded `ddconfig.txt and use ddconfigfile instead? Assuming ddconfigfile is the full path to the file, will this work in the above?

code/common/datadog.q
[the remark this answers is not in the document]
ThomasSmyth: That's it.
```suggestion
ddconfigfile:@[value;`ddconfigfile;hsym first .proc.getconfigfile"ddconfig.txt"]
```

code/common/datadog.q
PeterSmiley: Yes the function was more complicated previously which was why there was an error trap. Will change.

code/common/datadog.q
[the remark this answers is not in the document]
ThomasSmyth: A user can specify a custom path to the config file. For example mine might be `"datadog.txt"`. In this scenario your code may find the config doesn't exist and overwrite the port.

datadog/datachecks.q
PeterSmiley: At the moment this is run through a crontab not with a TorQ process.

datadog/setupdatadog.sh
PeterSmiley: If this is run before the user has started their TorQ process and the environment need set for the first time, is there a nicer way to source setenv.sh correctly so that TORQHOME will be the correct path?

ThomasSmyth: It shouldn't be, running `setenv.sh` is explicitly the first instruction before doing anything with TorQ and it should be the same here.

datadog/setupdatadog.sh
PeterSmiley: I think maybe you need sudo to write the changes to the datadog.yaml file. I checked with datadog support and the location (/etc/datadog-agent/) can't be changed. Is there another way to edit a read only yaml file without sudo?

datadog/runchecks.sh
PeterSmiley: Removed hard coding by adding PATH to crontab in setup file and removed .bashrc as this is not needed.

code/common/datadog.q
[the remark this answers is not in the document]
jzabroski: Why is the datadog namespace dg but the function call to the config file ddconfigfile?  Shouldn't the namespace be dd?  Or is there already a commonly accepted namespace for dd?  I'm a kdb hobbyist.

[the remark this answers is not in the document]
PeterSmiley: You are correct, they should be consistent. The configfile will be changed to dg.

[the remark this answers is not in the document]
jzabroski: Uh... but why not change hte namespace to dd?  after all, **D**ata**D**og...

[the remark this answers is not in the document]
jzabroski: @PeterSmiley  Just wondering why dg vs dd.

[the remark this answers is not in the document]
PeterSmiley: The dg was taken from **d**atado**g**. It made sense at the time of writing the code but the main thing is that the namespace is consistent throughout the files.

code/common/datadog.q
```diff
@@ -0,0 +1,41 @@
+// DATADOG CHECKS
+
+//Create datadog namespace
+\d .dg
+
+//Find the config file created during set up containing the port set by user in setenv. If it doesn't exist default to 8125.
+dgconfigfile:@[value;`dgconfigfile;hsym first .proc.getconfigfile"dgconfig.txt"]
+//sets dogstatsd_port to the port defined by dgconfigfile
+$[`dgconfig.txt in key hsym `$getenv[`KDBAPPCONFIG];value each read0 dgconfigfile;dogstatsd_port:8125]
```
ThomasSmyth: What does the config file look like? This makes it seem that it is a text file containing q commands?
```
value each read0 dgconfigfile
```
This line still suffers from the same issue from before where users can rename the config file but this stills expects `dgconfig.txt` I think you want this:
```suggestion
$[not()~key dgconfigfile;value each read0 dgconfigfile;dogstatsd_port:8125]
```
I would say the variable `dgconfigfile` should be the full path (not relative).

code/common/datadog.q
```diff
@@ -0,0 +1,41 @@
...
+handlers:(`$())!()
+
+is_ok:{
+  f:$[.proc.proctype in key .dg.handlers;.dg.handlers .proc.proctype;1b]
+ }
+
+// Send To Datadog function - takes a non string value and a stringed name.
+//Will send the value received (1 or 0) and the process name (hdb etc)
+//functions to send metrics and events to datadog from TorQ processes, error check for systems other than linux
+//$[any `l32`l64 in .z.o;sendMetric:{[metric_name;metric_value] system"bash -c \"echo  -n '",metric_name,":",(string metric_value),"|g|#shell' > /dev/udp/127.0.0.1/",(string dogstatsd_port),"\"";};"Currently only linux operating systems are supported to send metrics"]
```
ThomasSmyth: Should this comment be removed? Do you have a linux restriction in place somewhere else?

datadog/datachecks.q
```diff
@@ -0,0 +1,46 @@
...
+//Makes dictionary of defaults and uses .Q.opt to refer to command values passed by its key.
+o:.Q.def[`user`pass`timeout`init`noexit!(`admin;`admin;100;1b;0b);.Q.opt[.z.x]]
+
+// Send To Datadog function - takes a non string value and a stringed name.
+//Will send the value received (1 or 0) and the process name (hdb etc)
+//functions to send metrics and events to datadog from TorQ processes, error check for systems other than linux
+.dg.sendMetric:$[any `l32`l64 in .z.o;{[metric_name;metric_value] system"bash -c \"echo  -n '",metric_name,":",(string metric_value),"|g|#shell' > /dev/udp/127.0.0.1/",(string dogstatsd_port),"\"";};"Currently only linux operating systems are supported to send metrics"]
+
+.dg.sendEvent:$[any `l32`l64 in .z.o;{[event_title;event_text;tags;alert_type] system"event_title=",event_title,"; event_text=","\"",event_text,"\"","; tags=","\"#",$[0h=type tags;","sv tags;tags],"\"",";alert_type=",alert_type,"; ","echo \"_e{${#event_title},${#event_text}}:$event_title|$event_text|#$tags|t:$alert_type\" |nc -4u -w0 127.0.0.1 ",string dogstatsd_port; };"Currently only linux operating systems are supported to send events"]
```
ThomasSmyth: Either of these:
```
.z.o like"l*"
.z.o in`32`l64
```

datadog/datachecks.q
```diff
@@ -0,0 +1,46 @@
...
+
+//Creates the torq summary table without the pipes
+.dg.getprocess:{[x]
```
ThomasSmyth: Might be clearer to call this `getsummary` or `getprocessstatus`.

Note that above this line camel case is used for function names and below lower case is used. Use a consistent naming convention.

datadog/datachecks.q
```diff
@@ -0,0 +1,46 @@
...
+  {[x]flip (((`TIME`PROCESS`STATUS`PID`PORT!"TSSII")key[x]))$x} {[x] {[x](`$x[;0])! flip 1_ flip[x]} trim ("*****"; "|")0:x} system "./torq.sh summary"
```
ThomasSmyth: First issue here is that it uses a relative path to `torq.sh` that assumes it is in the same directory, can you make use of environment variables instead?
Second issue is that this seems incredibly complicated as the line below seems to do the same thing:
```suggestion
  ("TSSII";1#"|")0:system"$TORQHOME/torq.sh summary"
```

datadog/datachecks.q
```diff
@@ -0,0 +1,46 @@
...
+//Open port to process and sends check for each is_ok function
+.dg.sendcheck:{[o;x]
+  h:hopen[(hsym `$":" sv string[(`localhost;x[`PORT];o[`user];o[`pass])];o[`timeout])];
+  :h(`.dg.is_ok;`)
+ }
+//Creates name for the event (how it is found on datadog metric)
+.dg.createeventname:{[x]"_" sv string (`process;x[`PROCESS];`check)}
+
+//state either 1b or 0b (with error trap) calling sendMetric.
+.dg.check:{[x;o]state:.[.dg.sendcheck;(o;x);{[x]0b}];.dg.sendMetric[.dg.createeventname[x];state]}
```
ThomasSmyth: Can skip creation of `state` variable:
```suggestion
.dg.check:{[x;o].dg.sendMetric[.dg.createeventname x;.[.dg.sendcheck;(o;x);{[x]0b}]]}
```

datadog/datachecks.q
```diff
@@ -0,0 +1,46 @@
...
+
+//Checks each of the processes on the monitor list
+.dg.checkall:{[x;o].dg.check[;o] each x}
+
+//Retrieves the processes to be monitored with summary data. 
+.dg.init:{[x]
+  .dg.allprocesslist:.dg.getprocess[];
+  .dg.monitoredprocesslist:select from .dg.allprocesslist where any PROCESS like/:{[x]string[x],"*"}'[.dg.monitorprocess];
```
ThomasSmyth: Can do without lambda:
```suggestion
  .dg.monitoredprocesslist:select from .dg.allprocesslist where any PROCESS like/:string[.dg.monitorprocess],'"*";	
```

datadog/runchecks.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+echo `date`;
+q ./datadog/datachecks.q
```
ThomasSmyth: Use full path.

datadog/setupdatadog.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+echo dogstatsd_port is set to $DOGSTATSD_PORT in setenv.sh.
```
ThomasSmyth:
```suggestion
echo"dogstatsd_port is set to $DOGSTATSD_PORT in setenv.sh"
```

code/common/datadog.q
PeterSmiley: Sorry these functions have been moved to the datachecks.q file in datadog folder. I'm not sure how they ended up still being in this file, I thought I'd removed them

code/common/datadog.q
```diff
@@ -0,0 +1,22 @@
...
+\d .dg
+
+//Functions are set to return 1b or 0b based on the health of the service.
+//Default check returns 1b from each process to indicate process is up and can be queried.
+//These checks run every time crontab/timer is run.
+
+handlers:(`$())!()
+
+is_ok:{
+  f:$[.proc.proctype in key .dg.handlers;.dg.handlers .proc.proctype;1b]
```
ThomasSmyth: Can skip creating variable `f` + make one line + this naming convention is inconsistent, use `isok` or `isOK` (as appropriate): 
```suggestion
isok:{$[.proc.proctype in key .dg.handlers;.dg.handlers .proc.proctype;1b]}
```

datadog/setupdatadog.sh
```diff
@@ -0,0 +1,71 @@
...
+dogstatsd_port: $DOGSTATSD_PORT
+
+process_config:
+ enabled: \"true\"" >> $datadogfile
+  echo "Datadog.yaml file is now enabled to send data through port $DOGSTATSD_PORT"
+fi
+
+#Create the crontab to run runchecks.sh at specified intervals.
+while true;do
+  read -p "Do you wish to install a crontab? [y/N]" yn
```
ThomasSmyth: I think this should say something along the lines of "modifying crontab" rather than installing it.

code/common/datadog.q
```diff
@@ -0,0 +1,20 @@
...
+//These checks run every time crontab/timer is run.
+
+handlers:(`$())!()
+
+isok:{$[.proc.proctype in key .dg.handlers;.dg.handlers .proc.proctype;1b]}
+
+//Option to override default .lg.ext functionality to send error and warning events to datadog
+init:{[]
+  .lg.ext:{[loglevel;proctype;proc;id;message;dict]
+  if[loglevel in `ERR`WARN;.dg.sendEvent[string proc;message;string proctype;]$[loglevel=`ERR;"error";"warning"]]}
```
jonnypress: You should modify this to wrap any pre-existing condition - e.g. if someone has already modified .lg.ext and then wants to also add this, they have to manually modify. Which might be fine, but better to have separate e.g. 

.lg.ext:{[olddef;loglevel;proctype;proc;id;message;dict]
  olddef[loglevel;proctype;proc;id;message;dict]; 
  if[loglevel in `ERR`WARN;.dg.sendEvent[string proc;message;string proctype;]$[loglevel=`ERR;"error";"warning"]]}[@[value;`.lg.ext;{{[loglevel;proctype;proc;id;message;dict]}}]]

datadog/datachecks.q
```diff
@@ -0,0 +1,43 @@
...
+
+//Creates the torq summary table without the pipes
+.dg.getprocessstatus:("TSSII";1#"|")0:system"$TORQHOME/torq.sh summary"
+
+//Names of processes to be monitored to be edited depending on monitoring needs (must be input as `rdb1, `hdb2 etc.)
+.dg.monitorprocess:()
+
+//Open port to process and sends check for each isok function
+.dg.sendcheck:{[o;x]
+  h:hopen[(hsym `$":" sv string[(`localhost;x[`PORT];o[`user];o[`pass])];o[`timeout])];
```
jonnypress: Should this be hardcoded to localhost?

PR #194, opened 2019-12-19
Data persistence to save down the results historically

tszwailam:
added savedata and modified .u.end for the dqe process to savedown by end of day.

working on saving down check configurations

### DIFF
--- a/code/dqe/savedata.q
+++ b/code/dqe/savedata.q
@@ -1,11 +1,11 @@
 \d .dqe
 savedata:{[dir;pt;tabname]                                                                                       /- function to save data persistently
   .lg.o[`dqe;"Saving ",(string tabname)," data to ", string dir];
-.[
+  .[
     upsert;
     (` sv .Q.par[dir;pt;tabname],`;.Q.en[dir;r:0!.save.manipulate[tabname;`. tabname]]);                         /- upsert data to partition
     {[e] .lg.e[`savedata;"Failed to save dqe data to disk : ",e];`e}
-];
+  ];
   .lg.o[`delete;"deleting ",(string tabname)," data from in-memory table"];                                      /- empty the table from memory
   @[`.;tabname;0#];
   }
--- a/code/dqe/savedata.q
+++ b/code/dqe/savedata.q
@@ -1,11 +1,9 @@
 \d .dqe
 savedata:{[dir;pt;tabname]                                                                                       /- function to save data persistently
   .lg.o[`dqe;"Saving ",(string tabname)," data to ",.os.pth dir];
-  .[
-    upsert;
-    (` sv .Q.par[dir;pt;tabname],`;.Q.en[dir;r:0!.save.manipulate[tabname;`. tabname]]);                         /- upsert data to partition
-    {[e] .lg.e[`savedata;"Failed to save dqe data to disk : ",e];'e}
-  ];
+  pth:` sv .Q.par[dir;pt;tabname],`;
+  err:{[e].lg.e[`savedata;"Failed to save dqe data to disk : ",e];'e}
+  .[upsert;(pth;.Q.en[dir;r:0!.save.manipulate[tabname;`. tabname]]);err];
   .lg.o[`delete;"deleting ",(string tabname)," data from in-memory table"];                                      /- empty the table from memory
   @[`.;tabname;0#];
   };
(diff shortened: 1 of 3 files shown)


code/dqe/savedata.q
```diff
@@ -0,0 +1,18 @@
+\d .dqe
+savedata:{[dir;pt;tabname]                                                                                       /- function to save data persistently
+  .lg.o[`dqe;"Saving ",(string tabname)," data to ", string dir];
```
ThomasSmyth: There's an inbuilt TorQ function for stringing paths.
```suggestion
  .lg.o[`dqe;"Saving ",(string tabname)," data to ",.os.pth dir];
```

code/dqe/savedata.q
```diff
@@ -0,0 +1,18 @@
...
+    {[e] .lg.e[`savedata;"Failed to save dqe data to disk : ",e];`e}
+  ];
+  .lg.o[`delete;"deleting ",(string tabname)," data from in-memory table"];                                      /- empty the table from memory
+  @[`.;tabname;0#];
+  };
+
+endofday:{[pt]
+  .lg.o[`eod;"end of day message received - ",string pt];
+  savedata[.dqe.dqedbdir;pt;`.dqe.results];
+  savedata[.dqe.dqedbdir;pt;`.dqe.configtable];
```
ThomasSmyth: Combine these 2 bad boys:
```
savedata[.dqe.dqedbdir;pt]each `.dqe.results`.dqe.configtable;
```

code/processes/dqe.q
```diff
@@ -1,7 +1,13 @@
 \d .dqe
 
 configcsv:@[value;`.dqe.configcsv;first .proc.getconfigfile["dqeconfig.csv"]];
+dqedbdir:@[value;`dqedbdir;`:dqedb];
 detailcsv:@[value;`.dqe.detailcsv;first .proc.getconfigfile["dqedetailtab.csv"]];
+gmttime:@[value;`gmttime;1b];
+partitiontype:@[value;`partitiontype;`date];
+getpartition:@[value;`getpartition;
+	{{@[value;`.dqe.currentpartition;
```
ThomasSmyth: Should this be a nested function which will return a function when called? Looks like you want this to return a date.

code/processes/dqe.q
```diff
@@ -1,7 +1,13 @@
 \d .dqe
 
 configcsv:@[value;`.dqe.configcsv;first .proc.getconfigfile["dqeconfig.csv"]];
+dqedbdir:@[value;`dqedbdir;`:dqedb];
```
ThomasSmyth: I think this should be driven by an environment variable rather than being a relative path.

code/dqe/savedata.q
```diff
@@ -0,0 +1,18 @@
+\d .dqe
+savedata:{[dir;pt;tabname]                                                                                       /- function to save data persistently
+  .lg.o[`dqe;"Saving ",(string tabname)," data to ", string dir];
+  .[
+    upsert;
+    (` sv .Q.par[dir;pt;tabname],`;.Q.en[dir;r:0!.save.manipulate[tabname;`. tabname]]);                         /- upsert data to partition
+    {[e] .lg.e[`savedata;"Failed to save dqe data to disk : ",e];`e}
```
ThomasSmyth: This will return the symbol e and not the error message as a symbol. But I think this is supposed to throw an error? If so:
```suggestion
    {[e] .lg.e[`savedata;"Failed to save dqe data to disk : ",e];'e}
```
Else use `` `$`` to convert to symbol.

code/dqe/savedata.q
```diff
@@ -0,0 +1,18 @@
...
+  ];
```
ThomasSmyth: I'm really not a fan of this whole section, I would probably go for something like:
```
pth:` sv .Q.par[dir;pt;tabname],`;
err:{[e].lg.e[`savedata;"Failed to save dqe data to disk : ",e];'e}
.[upsert;(pth;.Q.en[dir;r:0!.save.manipulate[tabname;`. tabname]]);err];
```

code/processes/dqe.q
tszwailam: I made the change in setenv in torq-fsp in the branch upd_setting_dqe

PR #199, opened 2020-01-02
Attribute Checker

tszwailam:
Function to check if attributes on a table match expectations


code/dqe/attrcheck.q
```diff
@@ -0,0 +1,8 @@
+\d .dqe
+attrcheck:{[tab;attribute;col]
+  metaoftab:select c,a from meta tab where c in col;
+  checktab:([]c:col;a:attribute);
+  $[metaoftab~checktab;
```
ThomasSmyth: Can get this section down to a single line by removing some duplication:
```
(c;"attribute of ",(","sv string(),col)," ",$[c:metaoftab~checktab;"matched";"did not match"]," expectation")
```

code/dqe/attrcheck.q
```diff
@@ -0,0 +1,8 @@
+\d .dqe
+attrcheck:{[tab;attribute;col]
```
ThomasSmyth: Add a comment to explain this function.

code/dqe/attrcheck.q
```diff
@@ -0,0 +1,8 @@
+\d .dqe
```
ThomasSmyth: Add a space after this line.

code/dqe/attrcheck.q
ThomasSmyth: Should the passed attrs + cols be a dict? Would help prevent users passing 2 lists of lengths.

PR #200, opened 2020-01-07
Adding unit tests for getserverids in gateway

CatherineHigg:
Unit tests to ensure changes made to gateway.q haven't affected .gw.getserverids. Checks .gw.getserverids runs successfully or errors when expected.


code/processes/gwtests.q
```diff
@@ -0,0 +1,102 @@
+//test process for performing unit tests on functions within gateway.q
+//changes were made to .gw.getserverids
+
+//mock servers table
+.gw.servers:([serverid:`u#(3;4;5i)]handle:(6;7;8i); servertype:(`rdb;`hdb;`hdb); inuse:(0b;0b;0b);active:(0b;1b;1b);querycount:(0i;0i;0i);lastquery:(0Np;0Np;0Np);usage:(0D;0D;0D);attributes:(`date`tables!(2015.01.07 2015.01.08 2019.09.26 2019.09.27;`heartbeat`logmsg`quote`quote_iex`trade`trade_iex);`date`tables!(2015.01.07 2015.01.08 2019.09.26 2019.09.27;`heartbeat`logmsg`quote`quote_iex`trade`trade_iex);`date`tables!(2015.01.07 2015.01.08 2019.09.26 2019.09.27;`heartbeat`logmsg`quote`quote_iex`trade`trade_iex));disconnecttime:(0Np;0Np;0Np))
+
+.gw.getserverids:{[att]
```
ThomasSmyth: This function already exists in the gateway, why is it duplicated here? Issues with loading the gateway code without it trying to start?

CatherineHigg: We just wanted to isolate getserverids (and the functions it calls) and test it against a mock table to save loading in all of the gateway functions and variables, but you think this isn't the way to go?

rossedens: Could the GW not just load this file to avoid duplication? Call it something like gwlib.q or something and keep a lot of the generic stuff here.

ThomasSmyth: If `getserverids` ever needs modified then that needs to happen in 2 places or else this test falls behind. This is essentially a completely separate function and does not necessarily reflect what you wish to test.

jonathonmcmurray: if you place the function in a file within `code/gateway/`, it'll automatically get loaded by any process with proctype of `gateway`

so then you can run tests like

`q torq.q -proctype gateway -procname gatewaytest -tests {path to tests file}`

with no `-load` flag so you're not loading the main `gateway.q` file & starting an actual gateway

CatherineHigg: Okay thanks folks. I'll try moving gwtests.q and removing those functions from gateway.q

PR #201, opened 2020-01-07
if upd function is already defined, use that definition. This will po…

jonnypress:
…tentially override the settings for batch publishing and writing to the log file


code/processes/chainedtp.q
```diff
@@ -38,7 +38,7 @@ openlog:{[lgfile]
     .[set;(lgfile;());{[lgf;err] .lg.e[`openlog;"cannot create new log file : ",string[lgf]," : ", err]}[lgfile]]];
 
   /- backup upd & redefine for counting
-  updold:upd;
+  updold:`. `upd;
```
ThomasSmyth: Does it matter if `upd` is undefined and this returns nothing?

jonnypress: I think that would be problematic, but not worse than current situation (i.e. that would be a development bug). At this point, upd pretty much must be defined.

PR #202, opened 2020-01-08
Schema checker for DQE

tszwailam:
a check function that checks whether a table matches proposed schema


code/dqe/schemacheck.q
```diff
@@ -0,0 +1,6 @@
+\d .dqe
+schemacheck:{[tab;schema;colname]
+  origschema:select t,c from meta tab;
+  checkschema:([]t:schema;c:colname);
+  (c;"attribute of ",(","sv string(),col)," ",$[c:origschema~checkschema;"matched";"did not match"]," proposed schema")
```
ThomasSmyth: Type rather than attribute?

code/dqe/schemacheck.q
```diff
@@ -0,0 +1,6 @@
+\d .dqe
+schemacheck:{[tab;schema;colname]
```
ThomasSmyth: I would say "schema" is the names, types and attributes rather than just types :eyes: my 2 cents

tszwailam: There is an attribute check already, and this schemacheck function checks both col name and types. Would you say combining both attribute check and schemacheck makes sense?

PR #203, opened 2020-01-09
Nullcheck

svetsolo:
Function to analyse columns within a table and determine if there are any nulls and if so how many.  Configuration includes a list of column(s)  and outputs % of nulls for each column comparing it with a given threshold.


code/processes/dqe.q
```diff
@@ -50,6 +50,13 @@ failchk:{[runtype;idnum;error;proc]
   `.dqe.results set update chkstatus:`failed,output:0b,descp:c#enlist error,chkruntype:runtype from .dqe.results where id=idnum, procschk=proc,chkstatus=`started;
   }
 
+nullchk:{[t;colslist;thres]                                                                                     /- function to check percentage of nulls in each column from colslist of a table t
+  tt:$[(type colslist)<0;(enlist colslist)#t;colslist#t];
```
ThomasSmyth:
```suggestion
  tt:((),colslist)#t;
```

code/processes/dqe.q
```diff
@@ -50,6 +50,13 @@ failchk:{[runtype;idnum;error;proc]
...
+  d:(sum null tt)*100%count tt;
```
ThomasSmyth: Could combine with the line above
```suggestion
  d:sum[null tt]*100%count tt:((),colslist)#t;
```

code/processes/dqe.q
```diff
@@ -50,6 +50,13 @@ failchk:{[runtype;idnum;error;proc]
...
+  res:([] colsnames:key d; nullspercentage:value d);
+  res:update thresholdfail:nullspercentage>thres from res                                                       /- compare each column's nulls percentage with threshold thres
```
ThomasSmyth: No need to update res before returning
```suggestion
  update thresholdfail:nullspercentage>thres from res                                                           /- compare each column's nulls percentage with threshold thres
```

code/processes/dqe.q
```diff
@@ -50,6 +50,12 @@ failchk:{[runtype;idnum;error;proc]
   `.dqe.results set update chkstatus:`failed,output:0b,descp:c#enlist error,chkruntype:runtype from .dqe.results where id=idnum, procschk=proc,chkstatus=`started;
   }
 
+nullchk:{[t;colslist;thres]                                                                                     /- function to check percentage of nulls in each column from colslist of a table t
+  d:sum[null tt]*100%count tt:((),colslist)#t;
```
jonathonmcmurray: Does this work as expected for columns with nested lists e.g. string columns?

```q
q)sum null ([] a:1 2 0N 0N;b:1.1 2.2 0n 0n;c:("abc";"def";"ghi";"jkl"))
a| 2i
b| 2i
c| 0 0 0i
```

svetsolo: Thanks, the current version does not support nested lists, to include this case we can substitute "sum null t" by "{sum$[0h=type x;0=count@'x;null x]}each flip t" which will sum nulls inside nested lists if they appear in t.
For your example, it returns:
a| 2
b| 2
c| 0

PR #204, opened 2020-01-10
Gw documentation improvement

nicolew02:
- [x] Added some client call examples
- [x] Fixed some spelling and grammar 
- [x] Added some commenting in detailed client call example


docs/Processes.md
```diff
@@ -221,8 +229,63 @@ the gateway api. Use .api.p“.gw.\*” for more details.
...
+```
+For asynchronous calls
+```
+q) neg[h](`.gw.asyncexec;"`$last .z.x";`hdb);h[]
+```
+
+#### Calls to the HDB and RDB
+For synchronous calls
+```
+q) h(`.gw.syncexec;"$[{`date in cols x}[trade]; select from trade where date within (.z.d-2;.z.d-1); select from trade]";`rdb`hdb)
```
rossedens: - Should we use `.proc.proctype` to mimic other examples on the doc?

- If hardcoding the table here I don't think you need the lamba for the conditional.

docs/Processes.md
```diff
@@ -221,8 +229,63 @@ the gateway api. Use .api.p“.gw.\*” for more details.
...
+```
+For asynchronous calls
+```
+q) neg[h](`.gw.asyncexec;"$[{`date in cols x}[trade]; select from trade where date within (.z.d-2;.z.d-1); select from trade]";`rdb`hdb);h[]
```
rossedens: Same as above

docs/Processes.md
```diff
@@ -221,8 +229,63 @@ the gateway api. Use .api.p“.gw.\*” for more details.
...
+`q) h(".gw.synccallsallowed:0b")`.
+
+#### Calls to the RDB only
+For synchronous calls
+```
+// To return the avg price per sym for the day so far
+q) h(`.gw.syncexec;"select avp:avg price by sym from trade where time.date=.z.d";`rdb)
+
+// Using the hloc function
+q) h(`.gw.syncexec;"hloc[2020.01.08;2020.01.08;10]";`rdb)
```
rossedens: For calling functions defined on the service, can you just do:

```h(`.gw.syncexec;(`hloc;2020.01.08;2020.01.08;10);`rdb)```

File: docs/Processes.md [hunk 4]
```diff
@@ -221,8 +229,63 @@ the gateway api. Use .api.p“.gw.\*” for more details.
...
 
-For the purposes of demonstration, assume that the queries must be run
-across an RDB and HDB process, and the gateway has one RDB and two HDB
+### Client Call Examples
+
+Here are a couple of examples for using client calls via a handle to the gateway process. 
+It is important to note if asynchronous calls are to be made, the .gw.synccallsallowed 
+variable needs to be set to 0b before running any async calls.
+To do this use
+`q) h(".gw.synccallsallowed:0b")`.
```
rossedens: Don't think we should encourage a client to do this. This variable should be controlled by config on startup. I think it is enough to mention that it will need to be set to false to run async.

docs/Processes.md
```diff
@@ -221,8 +229,63 @@ the gateway api. Use .api.p“.gw.\*” for more details.
...
+
+#### Calls to the RDB only
+For synchronous calls
+```
+// To return the avg price per sym for the day so far
+q) h(`.gw.syncexec;"select avp:avg price by sym from trade where time.date=.z.d";`rdb)
+
+// Using the hloc function - change query for appropriate date
```
rossedens: Maybe show the definition for hloc function?

PR #205, opened 2020-01-14
Crossed Market Alert

tszwailam:
- This check should be implemented to look specifically at if and where the bid has exceeded the ask in market data.
- Description should return number of times it has crossed and when.


code/dqe/xmarketalert.q
```diff
@@ -0,0 +1,7 @@
+\d .dqe 
+xmarketalert:{[tab]                                     // alerts user when bid has exceeded the ask in market data
+  data:select from tab where bid>ask;
+  $[0=count data;
+    (1b;"bid has not exceeded the ask in market data");
+    (0b;(raze/) ("bid has exceeded the ask ";(string count data);" times and they have occured at: ";($[1=count exec time from data;string exec time from data;"," sv string exec time from data])))]
```
jonathonmcmurray:
```suggestion
    (0b;"bid has exceeded the ask ",string[count data]," times and they have occured at: ",$[1=count exec time from data;string exec time from data;"," sv string exec time from data])]
```

I think this should work & seems a bit cleaner than using a `raze/` ?

code/dqe/xmarketalert.q
```diff
@@ -0,0 +1,7 @@
+\d .dqe 
+xmarketalert:{[tab]                                     // alerts user when bid has exceeded the ask in market data
+  data:select from tab where bid>ask;
```
jonathonmcmurray: should we have a check here that these columns exist? or are we relying on error catching at a higher level in DQE?

e.g. what happens if a user passes `trade` table as argument to this function instead of `quote`?

config/dqeconfig.csv
```diff
@@ -4,3 +4,4 @@ chkslowsub,1000000,tickerplant,tickerplant1,repeat,09:00:00.000000000,19:00:00.0
 tableticking,(`quote;5;`minute),rdb,rdb1,single,12:00:00.000000000,0N,0N
 constructcheck,(`quote;`table),rdb,rdb1,repeat,09:00:00.000000000,0Wn,0D00:02:00
 attrcheck,(`quote;`s`g;`time`sym),rdb,rdb1,preat,09:00:00.000000000,0Wn,0D00:02:00
```
jonathonmcmurray: not changed in this PR, but noticing that this line appears to have a typo

```suggestion
attrcheck,(`quote;`s`g;`time`sym),rdb,rdb1,repeat,09:00:00.000000000,0Wn,0D00:02:00
```

code/dqe/xmarketalert.q
```diff
@@ -0,0 +1,9 @@
+\d .dqe 
+xmarketalert:{[tab]                                     // alerts user when bid has exceeded the ask in market data
+  $[0b=all `bid`ask in cols tab;
+    (0b;"columns bid and ask do not exist in provided table");
+    (data:select from tab where bid>ask;
+    $[0=count data;
```
ThomasSmyth: This second condition should probably be wrapped in square brackets to make it a series of statements rather than a list but it can be made into a single statement instead:
```suggestion
    $[0=count data:select from tab where bid>ask
```

tszwailam: I was testing it out the other day and it turns out the error handling of DQE handles the error mentioned by Jonathon earlier and no further conditions were needed

code/dqe/xmarketalert.q
jonathonmcmurray: @tszwailam can you answer this question?

tszwailam: @jonathonmcmurray Sorry I answered in the bottom and forgot to answer here - 
Error handling takes care of this and it doesn't require a conditional here

PR #206, opened 2020-01-14
DQE anomalycheck

svetsolo:
Anomaly checker - function to analyse columns within a table and determine if there are any anomalies (0w) and if so how many. Configuration includes column(s) from a columns list and % of anomalies that is compared to a threshold value.

### DIFF
--- a/code/processes/dqe.q
+++ b/code/processes/dqe.q
@@ -57,7 +57,7 @@ nullchk:{[t;colslist;thres]                                                     
   }
 
 anomalychk:{[t;colslist;thres]                                                                                  /-function to check percentage of anomalies in each column from colslist of a table t
-  d:({sum[{(x~0w) or (x~-0w) or (x~0W) or (x~-0W)}@'x]}each flip tt)*100%count tt:((),colslist)#t;
+  d:({sum{any x~'(0w;-0w;0W;-0W)}'[x]}@'x]}each flip tt)*100%count tt:((),colslist)#t;
   res:([] colsnames:key d; anomalypercentage:value d);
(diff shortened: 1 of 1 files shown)


code/processes/dqe.q
```diff
@@ -56,6 +56,12 @@ nullchk:{[t;colslist;thres]
   update thresholdfail:nullspercentage>thres from res                                                           /- compare each column's nulls percentage with threshold thres
   }
 
+anomalychk:{[t;colslist;thres]                                                                                  /-function to check percentage of anomalies in each column from colslist of a table t
+  d:({sum[{x~0w}@'x]}each flip tt)*100%count tt:((),colslist)#t;
```
jonathonmcmurray: what if one of the cols passed is a different type? with the `~` operator, this will only find *long* infinities, is that the desired behaviour?

svetsolo: Thanks, we were considering specifically 0w, so I am going to add other infinity types here

code/processes/dqe.q
```diff
@@ -56,6 +56,12 @@ nullchk:{[t;colslist;thres]
   update thresholdfail:nullspercentage>thres from res                                                           /- compare each column's nulls percentage with threshold thres
   }
 
+anomalychk:{[t;colslist;thres]                                                                                  /-function to check percentage of anomalies in each column from colslist of a table t
+  d:({sum[{(x~0w) or (x~-0w) or (x~0W) or (x~-0W)}@'x]}each flip tt)*100%count tt:((),colslist)#t;
```
ThomasSmyth: Can slim this up a little by using a list:
```suggestion
  d:({sum{any x~'(0w;-0w;0W;-0W)}'[x]}each flip tt)*100%count tt:((),colslist)#t;
```

PR #207, opened 2020-01-16
Gw imprv

nicolew02:
## Issue Overview
Improvements made to the Gateway to improve user usage.

## Checklist
 - [x] Added a check for active/inactive servers
 - [x] Added a check for null servertypes
 - [x] Added a check for incorrect parameter types
 - [x] Added unit tests for getserversid
 - [x] Improved Gateway documentation


code/gateway/gatewaylib.q
```diff
@@ -0,0 +1,105 @@
...
+     "are not valid servers: ",", " sv string servertype except allservers;
+     "requested servers are currently inactive: ",", " sv string servertype except activeservers
+    ],activeserversmsg;
+   ];
+   :(exec serverid by servertype from .gw.servers where active)[servertype];
+  ];
+
+  // its a dictionary of attributes
+  // check if attribute types are correct types
+  correctAttTypes:`date`servertype`tables!14 11 11h;
```
jonnypress: The rest of this file is all lower case for function names - can you stick to that

code/processes/gateway.q
```diff
@@ -458,8 +381,8 @@ asyncexecjpts:{[query;servertype;joinfunction;postback;timeout;sync]
    if[99h<>type servertype;
      // its a list of servertypes e.g. `rdb`hdb
      servertype:distinct servertype,();
-     missing:servertype except exec distinct servertype from .gw.servers where active;
-     if[count missing; errStr:.gw.errorprefix,"not all of the requested server types are available; missing "," " sv string missing];
+     errcheck:@[getserverids;servertype;{.gw.errorprefix,x}];
+     if[10h=type errcheck; errStr:errcheck];
```
jonnypress: don't mix in camel case here (errStr)... I know that's the way it is in the old code but better to remove that now

docs/Processes.md
```diff
@@ -154,14 +154,18 @@ not completed in the alotted time, and return a timeout error to the
 client. If the query is already running on any backend servers then they
 cannot be timed out other than by using the standard -T flag.
 
+As a default, the gateway is set to only take synchronous calls with 
+`.gw.synccallsallowed`. To allow asynchronous calls, edit the 
+gateway.q file so that .gw.synccallsallowed is set to false.
```
jonnypress: Is this still the case? I think if it can detect kdb 3.6 then it should just allow them

docs/Processes.md
```diff
@@ -221,8 +229,71 @@ the gateway api. Use .api.p“.gw.\*” for more details.
...
+variable needs to be set to 0b before running any async calls.
+
+#### Calls to the RDB only
+For synchronous calls
+```
+// To return the avg price per sym for the day so far
+q) h(`.gw.syncexec;"select avp:avg price by sym from trade where time.date=.z.d";`rdb)
+
+// hloc function in RDB process
+q) h(`.gw.syncexec;(`hloc);`rdb)
```
jonnypress: Don't need the brackets around (`hloc)

docs/Processes.md
jonathonmcmurray: is this description correct? if so, it seems counter-intuitive to me that `.gw.synccallsallowed` defines whether or not _async_ calls are allowed

code/gateway/gatewaylib.q
```diff
@@ -0,0 +1,105 @@
...
+ /- build the cross product of requirements
+ reqcross:buildcross[req];
+
+ /- calculate the cross product of data contributed by each source
+ /- and drop it from the list of stuff that is required
+ util:flip `remaining`found!flip ({[x;y;z] (y[0] except found; y[0] inter found:$[0=count y[0];y[0];buildcross x@'where each z])}[req]\)[(reqcross;());value s];
+
+ /- check if everything is done
+ if[(count last util`remaining) and not besteffort;
+  '"getserverscross: cannot satisfy query as the cross product of all attributes can't be matched"];
```
ThomasSmyth: Missing a space in your indent :eyes:
```suggestion
   '"getserverscross: cannot satisfy query as the cross product of all attributes can't be matched"];
```

CatherineHigg:
> There's a lot of code moved into code/gateway/gatewaylib.q. Can you point out the sections (functions) within that which have changed?
We've made changes to getserverids only. The other functions haven't been changed

tests/gwtests.csv
```diff
@@ -0,0 +1,14 @@
+action,ms,bytes,lang,code,repeat,minver,comment
+before,0,0,q,".proc.loadf getenv[`KDBCODE],"/processes/gwtests.q"",1,,
```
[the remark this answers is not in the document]
CatherineHigg: Sorry no, that should be `/gateway/gatewaylib.q`. I mustn't have pushed the updated csv. Will do that now

CatherineHigg:
Will make these changes now

docs/Processes.md
nicolew02: Am I right in saying if kdb v3.6 is being used, async calls are allowed and if not the user will have to set `.gw.synccallsallowed` to 0b?
Should I remove this paragraph completely or edit it to explain that with kdb versions older than 3.6, the user will potentially need to set .gw.synccallsallowed?

 I'll look into the code now, but when I was running the gateway in debug mode, I couldn't use async calls from a client unless I set `.gw.synccallsallowed` to 0b in the gateway.

tests/gwtests.csv
```diff
@@ -0,0 +1,15 @@
+action,ms,bytes,lang,code,repeat,minver,comment
+before,0,0,q,".proc.loadf getenv[`KDBCODE],"/gateway/gatewaylib.q"",1,,
```
jonathonmcmurray: I think as long as you use `-proctype gateway` when running the tests this should be automatically loaded, no need to load explicitly IMO

docs/Processes.md
jonathonmcmurray: I'd edit it rather than remove it. Definitely could be that you have to disable sync calls to allow async calls, just seems a little non-obvious to me (but that's more to do with the code itself than this documentation)

jonnypress: So 3.6 should allow both at the same time since kx put in that new deferred response mechanism. Pre 3.6 the gateways had to allow one or the other, otherwise life got difficult (e.g. someone could come in with an sync request and just ruin everyone else's good async work)

docs/Processes.md
```diff
@@ -221,8 +228,69 @@ the gateway api. Use .api.p“.gw.\*” for more details.
 | .gw.asyncexecjpt\[query; servertypes; joinfunction; postback; timeout\] | Execute the specified query against the required list of servers. Use the specified join function to aggregate the results. If the postback function is not set, the client must block and wait for the results. If it is set, the result will be wrapped in the specified postback function and returned asynchronously to the client. The query will be timed out if the timeout value is exceeded. |
 
 
-For the purposes of demonstration, assume that the queries must be run
-across an RDB and HDB process, and the gateway has one RDB and two HDB
+### Client Call Examples
+
+Here are a couple of examples for using client calls via a handle to the gateway process.
```
jonnypress: There are more than a couple 
I think it should be re-emphasized here that the async calls are only relevant for users on kdb < 3.6. For 3.6+ just use sync

PR #209, opened 2020-01-21
DQE Rangecheck

svetsolo:
Check if values are within range

### DIFF
--- a/code/processes/dqe.q
+++ b/code/processes/dqe.q
@@ -58,12 +58,13 @@ nullchk:{[t;colslist;thres]                                                     
 
 rangechk:{t;colslist;tlower;tupper;thres]                                                                       /- check that values are within the range defined by tlower and tupper tables
   colslist:((),colslist) except exec c from meta t where t in "csSC ";                                          /- exclude columns that do not have pre-defined limits
-  tupper:((),colslist)#tupper;
-  tlower:((),colslist)#tlower;
-  d:(sum (tt within (tlower;tupper))*100%count tt:((),colslist)#t;                                              /- dictionary with results by columns
+  tupper:colslist#tupper;
+  tlower:colslist#tlower;
+  d:(sum (tt within (tlower;tupper))*100%count tt:colslist#t;                                                   /- dictionary with results by columns
   res:([] colsnames:key d; inrangepercentage:value d);
   update thresholdfail:inrangepercentage<thres from res                                                         /- check if within range percentage is higher than threshold
   }
+
 postback:{[runtype;idnum;proc;result]                                                                           /- function that updates the results table with the check result
   $["e"=first result;                                                                                           /- checks if error returned from server side
     .dqe.failchk[runtype;idnum;result;proc];
--- a/code/processes/dqe.q
+++ b/code/processes/dqe.q
@@ -56,13 +56,13 @@ nullchk:{[t;colslist;thres]                                                     
   update thresholdfail:nullspercentage>thres from res                                                           /- compare each column's nulls percentage with threshold thres
   }
 
-rangechk:{t;colslist;tlower;tupper;thres]                                                                       /- function to check percentage of values within tlower and tupper range
-  colslist:colslist except exec c from meta t where t in "csSC";                                                /- exclude columns that do not have pre-defined limits
+rangechk:{t;colslist;tlower;tupper;thres]                                                                       /- check that values are within the range defined by tlower and tupper tables
+  colslist:((),colslist) except exec c from meta t where t in "csSC ";                                          /- exclude columns that do not have pre-defined limits
   tupper:((),colslist)#tupper;
   tlower:((),colslist)#tlower;
-  d:(sum (tt>=tlower)&(tt<=tupper))*100%count tt:((),colslist)#t;                                               /- dictionary with results by columns
+  d:(sum (tt within (tlower;tupper))*100%count tt:((),colslist)#t;                                              /- dictionary with results by columns
   res:([] colsnames:key d; inrangepercentage:value d);
-  update thresholdfail:inrangepercentage<thres from res                                                         /- check if within range percantage is higher than threshold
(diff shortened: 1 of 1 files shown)


code/processes/dqe.q
```diff
@@ -56,6 +56,14 @@ nullchk:{[t;colslist;thres]
   update thresholdfail:nullspercentage>thres from res                                                           /- compare each column's nulls percentage with threshold thres
   }
 
+rangechk:{t;colslist;tlower;tupper;thres]                                                                       /- function to check percentage of values within tlower and tupper range
+  colslist:colslist except exec c from meta t where t in "csSC";                                                /- exclude columns that do not have pre-defined limits
```
jonathonmcmurray:
```suggestion
  colslist:(),colslist except exec c from meta t where t in "csSC";                                                /- exclude columns that do not have pre-defined limits
```

then can remove the need for the `(),` on all the subsequent lines & remove some brackets

Also there may be some other datatypes that should be excluded - should probably exclude `" "` for mixed columns, and I think these checks probably won't work on nested cols e.g. `"FJI"` etc.

code/processes/dqe.q
```diff
@@ -56,6 +56,14 @@ nullchk:{[t;colslist;thres]
...
+  tupper:((),colslist)#tupper;
+  tlower:((),colslist)#tlower;
+  d:(sum (tt>=tlower)&(tt<=tupper))*100%count tt:((),colslist)#t;                                               /- dictionary with results by columns
```
jonathonmcmurray: can you use `within` here instead?

Also I may be misinterpreting the data structures here, but does this actually work?

```q
q)colslist:`a`b
q)t:([] a:til 30;b:til 30;d:til 30)
q)tt:colslist#t
q)tlower:`a`b`c!10 15 20
q)tlower:colslist#tlower
q)tt>=tlower
'length
  [0]  tt>=tlower
         ^
q)
```

code/processes/dqe.q
```diff
@@ -56,6 +56,14 @@ nullchk:{[t;colslist;thres]
...
+  res:([] colsnames:key d; inrangepercentage:value d);
+  update thresholdfail:inrangepercentage<thres from res                                                         /- check if within range percantage is higher than threshold
```
jonathonmcmurray:
```suggestion
  update thresholdfail:inrangepercentage<thres from res                                                         /- check if within range percentage is higher than threshold
```

typo in comment

code/processes/dqe.q
ThomasSmyth: If `collist` can be an atom then I think you need this:
```
  colslist:((),colslist)except exec c from meta t where t in "csSC";
```
`except` expects first arg to be a list, if `collist` is always a list then this doesn't apply.

code/processes/dqe.q
svetsolo: Thanks, tlower and tupper are tables not dictionaries, so for your example it can be
 q)tlower:([] a:30#10;b:30#15;c:30#20). Then after selecting columns in colslist, we have tlower:colslist#tlower and then q)tt>=tlower returns table we are looking for: 
q)tt>=tlower
a b
---
0 0
0 0
0 0
0 0
0 0
0 0
0 0
0 0
0 0
0 0
1 0
1 0
1 0
1 0
1 0
1 1
1 1
1 1
1 1
1 1
..

Would it be better to use another data type for tlower/tupper?

code/processes/dqe.q
```diff
@@ -56,6 +56,14 @@ nullchk:{[t;colslist;thres]
   update thresholdfail:nullspercentage>thres from res                                                           /- compare each column's nulls percentage with threshold thres
   }
 
+rangechk:{t;colslist;tlower;tupper;thres]                                                                       /- check that values are within the range defined by tlower and tupper tables
+  colslist:((),colslist) except exec c from meta t where t in "csSC ";                                          /- exclude columns that do not have pre-defined limits
+  tupper:((),colslist)#tupper;
+  tlower:((),colslist)#tlower;
+  d:(sum (tt within (tlower;tupper))*100%count tt:((),colslist)#t;                                              /- dictionary with results by columns
```
jonathonmcmurray: Don't need the `((),colslist)` in these 3 lines, can just use `colslist` as it's already guaranteed to be a list from first line in function

code/processes/dqe.q
jonathonmcmurray: a table seems a bit non-obvious to me, I'd have expected a dictionary given there's presumably only going to be one lower limit value per column. at the very least, this needs documented somewhere etc.

svetsolo: My reasoning was to have functionality of possible multiple upper and lower bounds for each column (such as different limits for different timepoints).
I've added comments about tlower and tupper being tables but can change to a dictionary as a better option if we assume one limit per column?

code/processes/dqe.q
svetsolo: Thanks! Fixed

code/processes/dqe.q
```diff
@@ -56,6 +56,15 @@ nullchk:{[t;colslist;thres]
   update thresholdfail:nullspercentage>thres from res                                                           /- compare each column's nulls percentage with threshold thres
   }
 
+rangechk:{t;colslist;tlower;tupper;thres]                                                                       /- check that values are within the range defined by tlower and tupper tables
```
jemmaborland: missing a `[`

code/processes/dqe.q
```diff
@@ -56,6 +56,15 @@ nullchk:{[t;colslist;thres]
...
+  colslist:((),colslist) except exec c from meta t where t in "csSC ";                                          /- exclude columns that do not have pre-defined limits
+  tupper:colslist#tupper;
+  tlower:colslist#tlower;
+  d:(sum (tt within (tlower;tupper))*100%count tt:colslist#t;                                                   /- dictionary with results by columns
```
jemmaborland: There's one too many brackets here - I'm counting 3 `(` against 2 `)`
``sum[tt within (tlower;tupper)]*100%count tt:colslist#t``?

PR #210, opened 2020-01-22
Allow current checks of DQE to run in a HDB

dmartinaquaq:
Create a feature that lets the checks be run in a HDB
Will need to attach where clause to anywhere that calls a table in the HDB.

### DIFF
--- a/code/common/dqe.q
+++ b/code/common/dqe.q
@@ -1,5 +1,6 @@
 .dqe.evaluate:{[table]
-  if[count where (),(11h=abs type table) & (table in .Q.pt);                                                    /- checks if any variable for check function is type symbol
-    .lg.o[`evaluate;(string table)," has been changed to a functional select with a where clause"];
-    ?[table;enlist (=;.Q.pf;last .Q.PV);0b;()]]
+  $[count where (),(11h=abs type table) & (table in .Q.pt);                                                    /- checks if any variable for check function is type symbol
+    [.lg.o[`evaluate;(string table)," has been changed to a functional select with a where clause"];
+    ?[table;enlist (=;.Q.pf;last .Q.PV);0b;()]];
+    value table]
   }
--- a/code/common/dqe.q
+++ b/code/common/dqe.q
@@ -1,6 +1,5 @@
 .dqe.evaluate:{[table]
-  $[count where (),(11h=abs type table) & (table in .Q.pt);                                                    /- checks if any variable for check function is type symbol
-    [.lg.o[`evaluate;(string table)," has been changed to a functional select with a where clause"];
-    ?[table;enlist (=;.Q.pf;last .Q.PV);0b;()]];
-    value table]
-  }
+  if[0=count where(11h=abs type table)&(table in .Q.pt);:value table];                                          /- checks if any variable for check function is type symbol
+    .lg.o[`evaluate;(string table)," has been changed to a functional select with a where clause"];
+    ?[table;enlist (=;.Q.pf;last .Q.PV);0b;()]
+  } 
--- a/code/common/dqe.q
+++ b/code/common/dqe.q
@@ -1,0 +1,5 @@
+.dqe.evaluate:{[table]
+  if[count where (),(11h=abs type table) & (table in .Q.pt);                                                    /- checks if any variable for check function is type symbol
+    .lg.o[`evaluate;(string table)," has been changed to a functional select with a where clause"];
+    ?[table;enlist (=;.Q.pf;last .Q.PV);0b;()]]
+  }
--- a/code/dqe/freeform.q
+++ b/code/dqe/freeform.q
@@ -3,7 +3,6 @@
     :(0b;"error: query must be sent as type string")];
   if[11h=type a:@[value;query;{`error}];
     c:not `error=a;
-    (c;query,$[c;" passed";" failed"]);
-    :()];
+    :(c;query,$[c;" passed";" failed"])];
   (1b;query," passed")
-  }
+  } 
--- a/code/dqe/freeform.q
+++ b/code/dqe/freeform.q
@@ -1,0 +1,9 @@
+.dqe.freeform:{[query]
+  if[not 10h=type query;
+    :(0b;"error: query must be sent as type string")];
+  if[11h=type a:@[value;query;{`error}];
+    c:not `error=a;
+    (c;query,$[c;" passed";" failed"]);
+    :()];
+  (1b;query," passed")
+  }
--- a/code/processes/dqe.q
+++ b/code/processes/dqe.q
@@ -45,7 +45,10 @@ initstatusupd:{[runtype;idnum;funct;params;rs]                                  
   if[idnum in exec id from .dqe.compcounter;delete from `.dqe.compcounter where id=idnum;];
   .lg.o[`initstatus;"setting up initial record(s) for id ",(string idnum)];
   .dqe.dupchk[runtype;idnum;params]'[rs];                                                                       /- calls dupchk function to check if last runs chkstatus is still started
-  parprint:`$("," sv string (raze/) (),params[`vars] params`fnpar),$[params`comp;",comp(",(string params[`compproc]),",",(string params`compallow),")";""];
+  vars:params`vars;
+  updvars:(key params[`vars]) b:where (),10h=type each value params`vars;
+  if[count updvars;(vars updvars):`$params[`vars] updvars];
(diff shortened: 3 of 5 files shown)


code/processes/dqe.q
```diff
@@ -113,8 +113,24 @@ postback:{[runtype;idnum;proc;params;result]
...
+containerfn:{[funct;names;vars]
+  if[any b:where 11h=abs type each vars;
+    startvars:vars b;
+    names:names b;
+    finvars:startvars b2:where startvars in tables[];
+    names:names b2;
+    if[count finvars;
+      finvars:finvars b3:where `date in ' cols each finvars;
+      names:names b3;
+      names set ' value each ("select from " ,/: (string finvars)),\:" where date=.z.p-1"];
```
jonathonmcmurray:
```suggestion
      names set' {select from x where date=.z.p-1}'[finvars];
```

should work here I think?

also should that be `.z.d` perhaps? what about if this is run on Mondays, should it be looking back to Friday?

code/processes/dqe.q
```diff
@@ -113,8 +113,24 @@ postback:{[runtype;idnum;proc;params;result]
...
   }
 
+containerfn:{[funct;names;vars]
+  if[any b:where 11h=abs type each vars;
+    startvars:vars b;
+    names:names b;
+    finvars:startvars b2:where startvars in tables[];
+    names:names b2;
+    if[count finvars;
+      finvars:finvars b3:where `date in ' cols each finvars;
```
jonathonmcmurray: if the goal here is to get partitioned tables, could you just compare against `.Q.pt` instead of `tables[]` above? Also has the advantage that would work for other partition schemes

code/processes/dqe.q
```diff
@@ -113,8 +113,24 @@ postback:{[runtype;idnum;proc;params;result]
     .dqe.updresultstab[runtype;idnum;.z.p;first result;result[1];`complete;params;proc]];
   }
 
+containerfn:{[funct;names;vars]
```
jonathonmcmurray: I don't really understand the goal this function, can you add some more comments etc.? is it just to make copies of all hdb tables with most recent partition or something?

code/processes/dqe.q
dmartinaquaq: I am going to replace `.z.d-1` with the most recent date in the date variable stored in the HDB.

code/processes/dqe.q
dmartinaquaq: This function is to check if the passed over function is being run on a table in a HDB. Then if it is, a where clause will be added to stop a full HDB table being called.
I will add more comments to make it clear :)

code/processes/dqe.q
dmartinaquaq: Just realised that I typed `.z.p` instead of `.z.d`. 
As said though it will be replaced soon enough :)

code/processes/dqe.q
jonathonmcmurray: I think this & next line can be removed now that you are comparing against `.Q.pt`

code/processes/dqe.q
```diff
@@ -113,8 +113,22 @@ postback:{[runtype;idnum;proc;params;result]
...
+    names:names b2;
+    if[count finvars;
+      names set ' value each ("select from " ,/: (string finvars)),\:" where date=.z.p-1"];                     /- replace table names with select statments when run in HDB
+  ];
+   value funct,vars;                                                                                            /- run check function
+  }
+
 getresult:{[runtype;funct;params;idnum;proc;hand]
   .lg.o[`getresults;raze"send function over to prcess: ",string proc];
+  namesfunct:raze -1#2#value funct;
```
ThomasSmyth: Need to explain the magic numbers.

code/processes/dqe.q
```diff
@@ -113,8 +113,22 @@ postback:{[runtype;idnum;proc;params;result]
...
   }
 
+containerfn:{[funct;names;vars]                                                                                 /- used to check if table is being called in HDB
+  if[any b:where 11h=abs type each vars;                                                                        /- checks if any variable for check function is type symbol
+    startvars:vars b;
+    names:names b;
+    finvars:startvars b2:where startvars in .Q.pt;                                                              /- check if the variables are partitioned tables
+    names:names b2;
+    if[count finvars;
+      names set ' value each ("select from " ,/: (string finvars)),\:" where date=.z.p-1"];                     /- replace table names with select statments when run in HDB
```
ThomasSmyth: Not sure about creating strings here, can functional form be used? Seems like it could be a good fit.

code/processes/dqe.q
```diff
@@ -113,9 +113,23 @@ postback:{[runtype;idnum;proc;params;result]
...
 
-getresult:{[runtype;funct;params;idnum;proc;hand]
+containerfn:{[funct;names;vars]                                                                                 /- used to check if table is being called in HDB
+  if[any b:where (),11h=abs type each vars;                                                                     /- checks if any variable for check function is type symbol
+    startvars:vars b;
+    names:names b;
+    finvars:startvars b2:where (),startvars in .Q.pt;                                                           /- check if the variables are partitioned tables
+    names:names b2;
+    if[count finvars;
+      vars[b2]:({?[x;enlist (=;.Q.pf;last .Q.PV);0b;()]}'[finvars b2])];                                        /- replace table names with select statments when run in HDB
```
ThomasSmyth: Don't think brackets are needed:
```suggestion
      vars[b2]:{?[x;enlist (=;.Q.pf;last .Q.PV);0b;()]}'[finvars b2]];                                          /- replace table names with select statments when run in HDB
```

code/common/dqe.q
```diff
@@ -0,0 +1,6 @@
+.dqe.evaluate:{[table]
+  $[count where (),(11h=abs type table) & (table in .Q.pt);                                                    /- checks if any variable for check function is type symbol
```
ThomasSmyth: Not sure the listing is necessary + simplify to `if`:
```suggestion
  if[0=count where(11h=abs type table)&table in .Q.pt;:value table];                                           /- checks if any variable for check function is type symbol
```

code/dqe/freeform.q
```diff
@@ -0,0 +1,9 @@
+.dqe.freeform:{[query]
+  if[not 10h=type query;
+    :(0b;"error: query must be sent as type string")];
+  if[11h=type a:@[value;query;{`error}];
+    c:not `error=a;
+    (c;query,$[c;" passed";" failed"]);
```
ThomasSmyth: Should this `if` return on this line instead?

code/processes/dqe.q
```diff
@@ -45,7 +45,11 @@ initstatusupd:{[runtype;idnum;funct;params;rs]
   if[idnum in exec id from .dqe.compcounter;delete from `.dqe.compcounter where id=idnum;];
   .lg.o[`initstatus;"setting up initial record(s) for id ",(string idnum)];
   .dqe.dupchk[runtype;idnum;params]'[rs];                                                                       /- calls dupchk function to check if last runs chkstatus is still started
-  `.dqe.results insert (idnum;funct;`$"," sv string (raze/) (),4#params;rs[0];rs[1];.z.p;0Np;0b;"";`started;runtype);
+  vars:params`vars;
+  updvars:(key params[`vars]) b:where (),10h=type each value params`vars;
+  if[count updvars;(vars updvars):`$params[`vars] updvars];
```
ThomasSmyth: Use squares instead for clarity:
```suggestion
  if[count updvars;vars[updvars]:`$params[`vars] updvars];
```

code/processes/dqe.q
```diff
@@ -64,13 +68,16 @@ chkcompare:{[runtype;idnum;params]
   a:d[`results] where not d[`procs]=params`compproc;                                                            /- obtain all the check returns
   procsforcomp:d[`procs] except params`compproc;
   b:d[`results] where d[`procs]=params`compproc;                                                                /- obtain the check to compare the others to
-  
-  if[all 0W=first b;                                                                                            /- if error in compare proc then fail check
+
+  if[@[{all 0W=x};first b;0b];                                                                                  /- if error in compare proc then fail check
     .dqe.updresultstab[runtype;idnum;.z.p;0b;"error: error on comparison process";`failed;params;`];:()];
-  errorprocs:d[`procs] where all each 0W=d`results;
+  errorprocs:d[`procs] where (),all each @[{0W=x};d`results;0b];
```
ThomasSmyth: Listing needed?

dmartinaquaq: The listing needs to be there because of the where getting a type error when the error trap returns 0b. 
Was there originally for if only one proc returns an error but think that might of been fixed elsewhere.

PR #214, opened 2020-01-24
Sym file checker

tszwailam:
Check to determine whether sym files exists at the top level


code/dqe/symfilecheck.q
```diff
@@ -0,0 +1,8 @@
+\d .dqe
+hdbdir:@[value;`hdbdir;`hdb]
+symfilecheck:{[directory;filename]                     // if sym file exists
+  fullpath:.Q.dd[directory]filename;
+  $[not ()~key hsym fullpath;
```
jonathonmcmurray: I think you can use `.os.Fex` here https://github.com/AquaQAnalytics/TorQ/blob/master/code/common/os.q#L4

```suggestion
  $[.os.Fex fullpath;
```

(could probably also lose the `fullpath` intermediate variable)

code/dqe/symfilecheck.q
```diff
@@ -0,0 +1,8 @@
+\d .dqe
+hdbdir:@[value;`hdbdir;`hdb]
+symfilecheck:{[directory;filename]                     // if sym file exists
+  (c;"sym file named ",(","sv string(), filename)," ",$[c:.os.Fex .Q.dd[directory]filename;"exists";"doesn't exists"])
+//  $[.os.Fex .Q.dd[directory]filename;
```
ThomasSmyth: Remove commented code.

code/dqe/symfilecheck.q
```diff
@@ -0,0 +1,8 @@
+\d .dqe
+hdbdir:@[value;`hdbdir;`hdb]
+symfilecheck:{[directory;filename]                     // if sym file exists
+  (c;"sym file named ",(","sv string(), filename)," ",$[c:.os.Fex .Q.dd[directory]filename;"exists";"doesn't exists"])
```
ThomasSmyth: Why is string stitching needed here? My assumption would be that filename is an atom.

code/dqe/symfilecheck.q
```diff
@@ -0,0 +1,5 @@
+\d .dqe
+hdbdir:@[value;`hdbdir;`hdb]
+symfilecheck:{[directory;filename]                     // if sym file exists
+  (c;"sym file named ",(string filename)," ",$[c:.os.Fex .Q.dd[directory]filename;"exists";"doesn't exists"])
```
ThomasSmyth: Typo:
```suggestion
  (c;"sym file named ",(string filename)," ",$[c:.os.Fex .Q.dd[directory]filename;"exists";"doesn't exist"])
```

PR #216, opened 2020-01-27
DQE d file check

svetsolo:
Function to check that the .d file in latest date matches previous date values.

### DIFF
--- a/code/processes/dqe.q
+++ b/code/processes/dqe.q
@@ -94,6 +94,14 @@ anomalychk:{[t;colslist;thres]                                                  
   d:({sum{any x~'(0w;-0w;0W;-0W)}'[x]}each flip tt)*100%count tt:((),colslist)#t;
   res:([] colsnames:key d; anomalypercentage:value d);
   update thresholdfail:anomalypercentage>thres from res                                                         /- compare each column's anomalies percentage with threshold thres
+  }
+
+dchk:{[tname;dirname]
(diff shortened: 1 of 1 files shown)


code/processes/dqe.q
```diff
@@ -94,6 +94,11 @@ anomalychk:{[t;colslist;thres]
   d:({sum{any x~'(0w;-0w;0W;-0W)}'[x]}each flip tt)*100%count tt:((),colslist)#t;
   res:([] colsnames:key d; anomalypercentage:value d);
   update thresholdfail:anomalypercentage>thres from res                                                         /- compare each column's anomalies percentage with threshold thres
+  }
+
+dfilechk:{[tname;dirname]                                                                                       /- function to check .d file. Sample use: .dqe.dfilechk[`trade;getenv `KDBHDB]
+  system"l ",dirname;
+  (get .Q.dd[`:.](`$string first .Q.PV;tname;`.d))~get .Q.dd[`:.](`$string first .Q.PV except first .Q.PV;tname;`.d)
```
ThomasSmyth: > latest date matches previous date values

Should this use last instead? This is a pretty neat way to do it:
```
(~). get each` sv'.Q.par'[`:.;-2#.Q.PV;tname],'`.d;
```
Two points to consider:
1. I think if you are comparing multiple partitions then there needs to be a check that more than 1 exists
2. You pass `dirname` but then use current location in the function, can use `dirname` again.

PR #217, opened 2020-01-31
DQE date check

svetsolo:
Function to check date vector contains latest date in an hdb

### DIFF
--- a/code/processes/dqe.q
+++ b/code/processes/dqe.q
@@ -118,10 +118,12 @@ dfilechk:{[tname;dirname]                                                       
     [.lg.o[`dfilechk;"Two partitions are available but there are no two .d files for the given table to compare"]; 0b]]
   }
 
-datechk:{[dirname] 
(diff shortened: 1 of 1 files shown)


code/processes/dqe.q
```diff
@@ -118,6 +118,15 @@ dfilechk:{[tname;dirname]
     [.lg.o[`dfilechk;"Two partitions are available but there are no two .d files for the given table to compare"]; 0b]]
   }
 
+datechk:{[dirname]                                                                                              /- function to check date vector contains latest date in an hdb 
+  system"l ",dirname;
+  if[not `PV in key`.Q;
+    .lg.o[`datechk;"The directory is not partitioned"]; :0b];
+  if[not `date in .Q.pf;
+    .lg.o[`datechk;"date is not a partition field value"]; :0b];
+  last date=.z.d
```
jonathonmcmurray: surely hdb typically doesn't contain today's data, so this check would usually return false? is flase the "expected" result for this check?

PR #219, opened 2020-02-04
Allow subscription when TP logs disabled

jonathonmcmurray:
In #215 ability to start tickerplant without TP logs was added

When subscriber connects (including rdb, wdb etc.) & attempts to replay log, this fails because `.u.L` can't be retrieved from tickerplant.

This PR changes this so that `.u.L` is always taken from tickerplant by indexing into `.u` (which gives generic null if variable is not defined) & displays a warning if `replaylog` was set to true but there is no log file provided by tickerplant.

Additionally, CTP defined `.u.L` regards of setting of `.ctp.createlog`; this meant that subscribing & replaying log while CTP log is disabled would still be broken - change to define `.u.L` only if log is being created.

### DIFF
--- a/code/common/subscriptions.q
+++ b/code/common/subscriptions.q
@@ -55,9 +55,8 @@ subscribe:{[tabs;instrs;setschema;replaylog;proc]
 	/-if replaylog is false,dont try to get the log file details - they may not exist
 	/-set the function to send to the server based on this
 	.lg.o[`subscribe;"getting details from the server"];
-	df:$[replaylog;{(.u.sub\:[x;y];(.u.i;.u.L);(.u `icounts);(.u `d))};{(.u.sub\:[x;y];(.u `i;`);(.u `icounts);(.u `d))}];
-	details:@[proc`w;(df;subtabs;instrs);
-		{.lg.e[`subscribe;"failed to get the required details from server : ",x];()}];
+    df:{(.u.sub\:[x;y];(.u`i`L);(.u `icounts);(.u `d))};
+    details:@[proc`w;(df;subtabs;instrs);{.lg.e[`subscribe;"subscribe failed : ",x];()}];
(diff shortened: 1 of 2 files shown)


code/common/subscriptions.q
```diff
@@ -55,9 +55,8 @@ subscribe:{[tabs;instrs;setschema;replaylog;proc]
 	/-if replaylog is false,dont try to get the log file details - they may not exist
 	/-set the function to send to the server based on this
 	.lg.o[`subscribe;"getting details from the server"];
-	df:$[replaylog;{(.u.sub\:[x;y];(.u.i;.u.L);(.u `icounts);(.u `d))};{(.u.sub\:[x;y];(.u `i;`);(.u `icounts);(.u `d))}];
-	details:@[proc`w;(df;subtabs;instrs);
-		{.lg.e[`subscribe;"failed to get the required details from server : ",x];()}];
+    df:{(.u.sub\:[x;y];(.u`i`L);(.u `icounts);(.u `d))};
```
ThomasSmyth: Bring tabs back

code/common/subscriptions.q
```diff
@@ -82,6 +81,8 @@ subscribe:{[tabs;instrs;setschema;replaylog;proc]
 			@[{-11!x;};details 1;{.lg.e[`subscribe;"could not replay the log file: ", x]}];
 			/-reset the upd function back to original upd
 			@[`.;`upd;:;origupd]];
+		if[replaylog&null details[1;1];
+			.lg.w[`subscribe;"replaylog set to true but TP not using log file"]];
```
jonnypress: Could you just make this an error? Warnings tend to get ignored, with error the process will fail and then someone (a developer) will fix it whichever way they want it (to replay or not to replay)

PR #220, opened 2020-02-05
Update CTP subscribe function

mpotterq:
The `.sub.getsubscriptionhandles` function was updated to ensure that it returned correct subscriber information and also to handle `()` as a function argument. The CTP was using a dummy parameter when calling this function to return any handle with the correct process name, and the new function behaviour would cause this not to work and in fact return nothing. 

Instead, the dummy process type should be replaced with a null symbol or `()` to return processes of any type with the given process name, and that is what has been done with this change.

### DIFF
--- a/code/processes/chainedtp.q
+++ b/code/processes/chainedtp.q
@@ -55,7 +55,7 @@ openlog:{[lgfile]
 
 /- subscribe to tickerplant and refresh tickerplant settings
 subscribe:{[]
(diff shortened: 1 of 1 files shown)


jonathonmcmurray:
Can you add a brief description to PR of why this change is necessary? i.e. what bug this is fixing

Helpful to have a little more detail in the commit history for the future

PR #222, opened 2020-02-05
Schema checker comparing rdb and hdb

tszwailam:
The comparison does not include attribute for now. 

Also wondering if this is the best way to approach it using ports


code/dqe/rdbhdbcompare.q
```diff
@@ -0,0 +1,10 @@
+\d .dqe
+hdbport:@[value;`hdbport;6003]
+rdbhdbcompare:{[tab]                                    // this function should be ran in the rdb
+  h:hopen (`::56003:admin:admin);
+  hdbmeta:h(meta;tab);
+  removeddate:select c,t,f from delete from hdbmeta where c=`date;
```
jonathonmcmurray:
```suggestion
  removeddate:select c,t,f from hdbmeta where c<>`date;
```

code/dqe/rdbhdbcompare.q
```diff
@@ -0,0 +1,10 @@
+\d .dqe
+hdbport:@[value;`hdbport;6003]
+rdbhdbcompare:{[tab]                                    // this function should be ran in the rdb
+  h:hopen (`::56003:admin:admin);
```
jonathonmcmurray: even if line 2 was the right way to get the port in torq, you don't use it here :grin: 

I think you should be using built in TorQ stuff rather than hardcoding a port number - this is the whole reason discovery process exists :)

See this section of TorQ docs: http://aquaqanalytics.github.io/TorQ/conn/#retrieving-and-using-handles

code/dqe/rdbhdbcompare.q
```diff
@@ -0,0 +1,9 @@
+\d .dqe
+rdbhdbcompare:{[tab]
+  hdbmeta:(first exec w from .servers.SERVERS where procname=`hdb1)(meta;tab);
```
jonathonmcmurray: Rather than directly querying `.servers.SERVERS`, I think you should use the `.servers.getservers` function

code/dqe/rdbhdbcompare.q
```diff
@@ -0,0 +1,9 @@
+\d .dqe
+rdbhdbcompare:{[tab]
+  hdbmeta:(first exec w from .servers.getservers[`procname;`hdb1;()!();1b;0b])(meta;tab);
+  rdbmeta:(first exec w from .servers.getservers[`procname;`rdb1;()!();1b;0b])(meta;tab);
```
jonathonmcmurray: what if we want to check `rdb1` against `hdb2`? or `rdb2` against `hdb2`? etc. Maybe procnames should be passed as args here?

code/dqe/rdbhdbcompare.q
```diff
@@ -0,0 +1,9 @@
+\d .dqe
+rdbhdbcompare:{[tab;hdbproc;rdbproc]
```
ThomasSmyth: Add a comment for this function.

PR #231, opened 2020-02-19
DQE engine distinct symbols count

svetsolo:
DQE side distinct symbols - a function on the Engine side that outputs the count of distinct symbols in a pre-specified column and a pre-specified table

### DIFF
--- a/code/dqe/symcount.q
+++ b/code/dqe/symcount.q
@@ -1,4 +1,4 @@
 \d .dqe
-symcount:{[t;col]
-  (enlist t)!(enlist count {?[x; enlist(=;.Q.pf;last .Q.PV); 1b; (enlist y)!(enlist y)]}[t;col])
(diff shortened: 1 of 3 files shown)


code/dqe/symcount.q
```diff
@@ -0,0 +1,4 @@
+\d .dqe
+symcount:{[t;col]
+  (enlist t)!(enlist count {?[x; enlist(=;.Q.pf;last .Q.PV); 1b; (enlist y)!(enlist y)]}[t;col])
```
jonathonmcmurray: Don't think the lambda is doing anything useful here, can get rid of it

```suggestion
  (enlist t)!(enlist count ?[t; enlist(=;.Q.pf;last .Q.PV); 1b; {x!x}enlist col])
```

Do we maybe need a comment or something somewhere to say that this only works on partitioned tables in an hdb too? This would fail if someone tried to run in rdb for example, because of the use of `.Q.pf` & `.Q.PV`

PR #232, opened 2020-02-19
Sym file sym count on the engine side

tszwailam:
Checking the number of syms in sym file - 
Also implemented a check regarding the null type to fit the schema of .dqe.resultstab

### DIFF
--- a/code/dqe/symfilecheck.q
+++ b/code/dqe/symfilecheck.q
@@ -2,5 +2,5 @@
 hdbdir:@[value;`hdbdir;`hdb]
 symfilecheck:{[directory;filename]
   .lg.o[`test;"Counting number of symbols in the symbol file each day"];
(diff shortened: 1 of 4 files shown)


code/dqe/symfilecheck.q
```diff
@@ -0,0 +1,6 @@
+\d .dqe
+hdbdir:@[value;`hdbdir;`hdb]
```
jonathonmcmurray: should this be an hsym i.e. `` `:hdb``? also should it be defaulting to a TorQ env var perhaps, e.g. ``getenv`KDBHDB`` ?

tszwailam: I have `.dqe.hdbdir` set as `getenv`KDBHDB` in `config/settings/dqe.q` and will fix that typo

PR #234, opened 2020-02-21
Masterslave algorithm

nicolew02:
## Master slave Algorithm
Algorithm to determine which slave process is the master process.
Each process starts up thinking it's the master and as it connects to other processes of the same type, it negotiates. Process making outbound connection initiates negotiation. 

## Checklist
 - [ ] checkifmaster function to return dictionary of process details 
 - [ ] table stored in each process that holds own and other processes details
 - [ ] table updated with which process is master and with new/dropped connections 
 - [ ] if master process drops, renegotiation of processes to determine new master (.z.pc override)
 - [ ] init function to initialise
 - [ ] sendtoslaves function - publish msg to all slaves (only master to do this)
 - [ ] sendtomaster function - propagates message to the master

### DIFF
--- a/masterslave.q
+++ b/masterslave.q
@@ -1,0 +1,62 @@
+//Master slave algorithm
+//Each process initially assumes it's the master process
+//Master process is determined by oldest process start time 
+//Connect to other slave processes and retrieve details to determine which process is master
+
+//run lines
+//q torq.q -load lb.q -proctype lb -procname lb1 -p 6100 -debug -parentproctype wdb
+//q torq.q -load lb.q -proctype lb -procname lb2 -p 6101 -debug -parentproctype wdb
+
+
+// the list of processes to connect to
+.servers.CONNECTIONS:`lb
+
+// custom function, this is invoked when a new outbound connection is created
+// to be customised to invoke negotiation of processes
+.servers.connectcustom:{.lg.o[`connect;"found new connection"]; show x}
+
+// create connections
+.servers.startup[]
+
+\d .masterslave
+
+//table scehma of connected lb processes with ismaster status
+statustable:([handle:`int$()] procname:`symbol$();starttimeUTC:`timestamp$();ismaster:`boolean$())
+
+//store own process start timestamp
+start:.z.p
+
+//set details dict with own details
+details:{`procname`starttimeUTC`ismaster!(.proc.procname;.masterslave.start;1b)} 
+
+//get details of procname provided
+getdetails:{[processname] (first exec w from .servers.SERVERS where procname like processname) ".masterslave.details[]"}
+
+//update .masterslave.statustable with other proc details and update ismaster col to determine which process is master 
+addmember:{[processname]
+    update ismaster:0b from (`.masterslave.statustable upsert .masterslave.getdetails[processname],(enlist `handle)!(enlist first exec w from .servers.SERVERS  where procname like processname)) where starttimeUTC<>min starttimeUTC}
+
+//is the process itself the master
+ammaster:{exec ismaster from .masterslave.statustable where handle=0}
+
+//find which process is the master
+findmaster:{first exec handle from .masterslave.statustable where ismaster=1b}
+
+//wrapper func for finding master
+checkifmaster:{
+    .servers.startup[];
+    .masterslave.addmember each string exec procname from .servers.SERVERS where proctype like "lb", not null w
+ }
+
+//add pc override here
+//remove dropped connection from statustable
+//tell other alive processes to renegotiate who the master is 
+
+
+\d .
+
+//populate statutable and find which process is master
+.masterslave.checkifmaster[]
+
+
+//.z.pc call 
--- a/masterslave.q
+++ b/masterslave.q
@@ -4,8 +4,8 @@
 //Connect to other slave processes and retrieve details to determine which process is master
 
 //run lines
-//q torq.q -load lb.q -proctype lb -procname lb1 -p 6100 -debug -parentproctype wdb
-//q torq.q -load lb.q -proctype lb -procname lb2 -p 6101 -debug -parentproctype wdb
+//q torq.q -load masterslave.q -proctype lb -procname lb1 -p 6100 -debug -parentproctype wdb
+//q torq.q -load masterslave.q -proctype lb -procname lb2 -p 6101 -debug -parentproctype wdb
 
 
 // the list of processes to connect to


masterslave.q
```diff
@@ -0,0 +1,62 @@
...
+\d .masterslave
+
+//table scehma of connected lb processes with ismaster status
+statustable:([handle:`int$()] procname:`symbol$();starttimeUTC:`timestamp$();ismaster:`boolean$())
+
+//store own process start timestamp
+start:.z.p
+
+//set details dict with own details
+details:{`procname`starttimeUTC`ismaster!(.proc.procname;.masterslave.start;1b)}
```
jonnypress: I think probably for now we can drop the ismaster value
We can just pass the name and the start time, and then both sides of the negotiation can work out which is the master (which might be neither of them)

masterslave.q
```diff
@@ -1,62 +1,88 @@
...
+ // the list of processes to connect to
+ .servers.CONNECTIONS:distinct .servers.CONNECTIONS,.proc.proctype;
+
+ // custom function, this is invoked when a new outbound connection is created
+ // to be customised to invoke negotiation of processes
+ .servers.connectcustom:{.lg.o[`connect;"found new connection"];
+                         .masterslave.checkifmaster[];
+                         };
+ // create connections
+ .servers.startup[]
```
jonnypress: Having .servers.startup[] in here is fine for testing purposes, but in final version this shouldn't be in here, i.e. someone who loads this should do

.ms.init[]
and then call 
.servers.startup[] later in their code

Then init function should also be within the .masterslave namespace

masterslave.q
```diff
@@ -1,62 +1,88 @@
...
+ };
 
 \d .masterslave
 
 //table scehma of connected lb processes with ismaster status
-statustable:([handle:`int$()] procname:`symbol$();starttimeUTC:`timestamp$();ismaster:`boolean$())
+statustable:([handle:`int$()] procname:();starttimeUTC:`timestamp$();ismaster:`boolean$());
```
jonnypress: why did you change procname to ()? symbol is better.

masterslave.q
```diff
@@ -1,62 +1,88 @@
...
-    update ismaster:0b from (`.masterslave.statustable upsert .masterslave.getdetails[processname],(enlist `handle)!(enlist first exec w from .servers.SERVERS  where procname like processname)) where starttimeUTC<>min starttimeUTC}
+ `.masterslave.statustable upsert .masterslave.getdetails[processname],
+  (`handle`ismaster)!(first exec w from .servers.SERVERS  where procname like processname,not null w;1b)
+ };
+
+masterupdate:{update ismaster:starttimeUTC=min starttimeUTC from `.masterslave.statustable}
 
 //is the process itself the master
-ammaster:{exec ismaster from .masterslave.statustable where handle=0}
+ammaster:{exec ismaster from .masterslave.statustable where handle=0};
```
jonnypress: This is probably returning a list, should return an atom

masterslave.q
```diff
@@ -1,62 +1,88 @@
...
-findmaster:{first exec handle from .masterslave.statustable where ismaster=1b}
+//find which process is the master - can only be run after checkifmaster has been run
+findmaster:{exec handle, procname from .masterslave.statustable where ismaster=1b};
 
 //wrapper func for finding master
 checkifmaster:{
-    .servers.startup[];
-    .masterslave.addmember each string exec procname from .servers.SERVERS where proctype like "lb", not null w
- }
+ (.masterslave.addmember')[string exec procname from .servers.SERVERS where proctype like "lb", not null w];
```
jonnypress: Don't hardcode the proctype in here. Should just be 

where proctype=.proc.proctype
.i.e proctype is the same as the local process

masterslave.q
```diff
@@ -1,62 +1,88 @@
...
 \d .
 
 //populate statutable and find which process is master
-.masterslave.checkifmaster[]
-
+init[];
 
-//.z.pc call 
+//.z.pc call
+.z.pc:{.masterslave.pc[x y;y]}.z.pc;
```
jonnypress: need to wrap this in an error trap (in case .z.pc isn't initially defined) i.e. 
.z.pc:{.masterslave.pc[x y;y]}@[value;`.z.pc;{{[x]}}]
Also as per comment above, better to have .masterslave.pc only take one param, and then .z.pc is 

.z.pc:{x y; .masterslave.pc[y]}@[value;`.z.pc;{{[x]}}]

.z.pc should only be defined within init[]

masterslave.q
```diff
@@ -1,62 +1,88 @@
...
+ };
+
+deletedropped:{[W]delete from `.masterslave.statustable where handle=W};
 
 //add pc override here
 //remove dropped connection from statustable
-//tell other alive processes to renegotiate who the master is 
-
+//tell other alive processes to renegotiate who the master is
+pc:{[result;W]
```
jonnypress: pc should probably only take one param here, W

masterslave.q
```diff
@@ -1,62 +1,88 @@
...
+//3 processes, one master, master dies, one of the remaining two takes over
+
+init:{
+ // the list of processes to connect to
+ .servers.CONNECTIONS:distinct .servers.CONNECTIONS,.proc.proctype;
+
+ // custom function, this is invoked when a new outbound connection is created
+ // to be customised to invoke negotiation of processes
+ .servers.connectcustom:{.lg.o[`connect;"found new connection"];
+                         .masterslave.checkifmaster[];
```
jonnypress: If we get a new connection in, then we only want to get the details for that single new connection. WE don't want to re-run the whole thing for all connections, which I think is what is happening here.

masterslave.q
nicolew02: I was getting type errors because I kept switching between symbol and string procnames.
Will change so there will only be sym procnames

masterslave.q
```diff
@@ -0,0 +1,109 @@
...
+
+
+
+deletedropped:{[W]delete from `.masterslave.statustable where handle=W};
+
+//add pc override here
+//remove dropped connection from statustable
+//tell other alive processes to renegotiate who the master is
+pc:{[W]
+ .masterslave.deletedropped[W];
```
jonnypress: You need a check in here to not do this if W=0
When the process starts up in the background it will do a close event on the 0 handle, which we don't want to remove from the table.

masterslave.q
```diff
@@ -0,0 +1,109 @@
...
+
+
+\d .masterslave
+init:{
+ // the list of processes to connect to
+ .servers.CONNECTIONS:distinct .servers.CONNECTIONS,.proc.proctype;
+
+ // custom function, this is invoked when a new outbound connection is created
+ // to be customised to invoke negotiation of processes
+ .servers.connectcustom:{[f;x] .masterslave.checkifmaster[];
```
jonnypress: I think this will work, but there is still too much work going on here (too much IPC)
Basically .servers.connectcustom will get invoked when a new connection is made, and it will be passed in (as the x argument in this case) the new connections. So, you only need this process to negotiate with those new processes, not all the old ones too. In this version, if there are 10 connections, and a new one arrives, then the way you have it every process will make 11 outbound requests, whereas really they only need to be making one

nicolew02: Is `f` the function to be called on the new connection handle`x`?

And instead of running .ms.checkifmaster when a new connection is made,
I should call .ms.addmember with the new connection procname?

jonnypress: > Is f the function to be called on the new connection handlex?
Yes. f is the pre-existing definition of .servers.connectcustom (which may be nothing)
> I should call .ms.addmember with the new connection procname
I think so

masterslave.q
```diff
@@ -0,0 +1,102 @@
...
+
+//get details of procname provided
+getdetails:{[processname]
+ (first exec w from .servers.SERVERS where procname like raze string processname,not null w) (`.masterslave.details;[])
+ };
+
+//update .masterslave.statustable with other proc details and update ismaster col to determine which process is master
+addmember:{[processname]
+ `.masterslave.statustable upsert .masterslave.getdetails[processname],
+  (`handle`ismaster)!(first exec w from .servers.SERVERS  where procname like processname,not null w;1b)
```
jonnypress: Is this return value used for anything?

masterslave.q
```diff
@@ -0,0 +1,102 @@
...
+start:.z.p;
+
+//set details dict with own details
+details:{
+ `procname`starttimeUTC!(.proc.procname;.masterslave.start)
+ };
+
+//get details of procname provided
+getdetails:{[processname]
+ (first exec w from .servers.SERVERS where procname like raze string processname,not null w) (`.masterslave.details;[])
```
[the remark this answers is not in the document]
nicolew02: will change it to =, I think it was cause I was originally using strings

masterslave.q
nicolew02: as in when it returns `.masterslave.statustable?
This func is just to upsert process details dictionary into the .ms.statustable, I'm not aware of any other return value?

ThomasSmyth:
Lots of terminology to update in here.

Issue #238, opened 2020-03-12
Modify RDB so it doesn't hit the timeout (-T) at end of day


jonnypress:
If the RDB process is set up with a timeout (-T) then it can sometimes breach the timeout and hit a stop error at EOD (for high volume datasets)

This has to be handled in both cases: 
1. when the RDB saves down the data itself
2. when the RDB offloads the save down to the WDB

The 2nd case is cleaner to handle, the first case is a bit trickier as it is all invoked from the single end-of-day message.

I think the simplest solution for now is to set a timer on start up for the time defined by .eodtime.nextroll which runs 1 minute before the roll time every 24 hours. The timer function captures the current value if \T and sets it to 0. When the end-of-day is finished (both versions - reloadenabled is both 0 and 1) it resets the \T flag back to what it was previously.

PR #244, opened 2020-03-19
RDB timeout change to 0 before EOD write-down

robsketch:
## RDB EOD Timeout adjustment
If `reloadenabled` is `0b`, the RDB will do the EOD writedown and will not use the WDB.  In this case, the timeout command line parameter will be set to 0 one minute before EOD write-down each day, so that if the writedown takes longer than the original -T parameter set in process.csv, it will not run into a 'stop error and write-down will complete successfully.

If `reloadenabled` is `1b`, then the RDB will offload the save-down to the WDB.  Wiping memory from RDB at EOD can take time so should set the timeout to be zero in this case too.  This can be done in the `reload` function in the RDB which is called by the WDB.

This PR addresses issue #238 

## Checklist
- [x] Add ` timeoutreset` function in RDB process which sets -T to be 0 if `reloadenabled` is false
- [x] Set a timer function for `timeoutreset` to be called one minute before EOD writedown
- [x] Add functionality to when RDB offloads the save-down to the WDB
- [x] Test case when `reloadenabled` is false
- [x] Test case when `reloadenabled` is true

### DIFF
--- a/code/processes/rdb.q
+++ b/code/processes/rdb.q
@@ -98,7 +98,7 @@ endofday:{[date]
 	/-save and wipe the tables
 	writedown[hdbdir;date];
 	/-reset timeout to original timeout
-    restoretimeout[];
+        restoretimeout[];
 	/-reapply the attributes
 	/-functional update is equivalent of {update col:`att#col from tab}each tables
 	(![;();0b;].)each a where 0<count each a[;1];
--- a/code/processes/rdb.q
+++ b/code/processes/rdb.q
@@ -112,8 +112,6 @@ endofday:{[date]
 	
 reload:{[date]
 	.lg.o[`reload;"reload command has been called remotely"];
-	/-set timeout to be 0 before tables are wiped
-	system"T 0";
 	/-get all attributes from all tables before they are wiped
 	/-get a list of pairs (tablename;columnname!attributes)
 	a:{(x;raze exec {(enlist x)!enlist((#);enlist y;x)}'[c;a] from meta x where not null a)}each tabs:subtables except ignorelist;
@@ -174,7 +172,7 @@ notpconnected:{[]
 
 /-resets timeout to 0 before EOD writedown
 timeoutreset:{.rdb.timeout:system"T";system"T 0"};
-restoretimeout:{system["T ", string timeout]};
+restoretimeout:{system["T ", string .rdb.timeout]};
 \d .
 
 /- make sure that the process will make a connection to each of the tickerplant and hdb types
--- a/code/processes/rdb.q
+++ b/code/processes/rdb.q
@@ -205,6 +205,5 @@ reload:.rdb.reload
 
 /-change timeout to zero before eod flush
 .rdb.timeoutreset[]
-.timer.repeat[.eodtime.nextroll-1000000000*60;0W;1D00:00:00.000000000;
+.timer.repeat[.eodtime.nextroll-00:01;0W;1D;
   (`.rdb.timeoutreset;`);"Set rdb timeout to 0 for EOD writedown"];
-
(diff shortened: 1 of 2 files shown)


code/processes/rdb.q
```diff
@@ -3,6 +3,8 @@
 /-changes added 
 /-Can specify the hdb directory rather than relying on the tickerplant
 
+
+
```
ThomasSmyth: 🤔

code/processes/rdb.q
```diff
@@ -194,3 +203,8 @@ reload:.rdb.reload
 
 /-set the partition that is held in the rdb (for use by the gateway)
 .rdb.setpartition[]
+
+/-change timeout to zero before eod flush
+.rdb.timeoutreset[]
+.timer.repeat[.eodtime.nextroll-1000000000*60;0W;1D00:00:00.000000000;(`.rdb.timeoutreset;`);"Set rdb timeout to 0 for EOD writedown"];
```
ThomasSmyth: Same as this?
```suggestion
.timer.repeat[.eodtime.nextroll-00:01;0W;1D;(`.rdb.timeoutreset;`);"Set rdb timeout to 0 for EOD writedown"];
```

code/processes/rdb.q
```diff
@@ -194,3 +202,9 @@ reload:.rdb.reload
 
 /-set the partition that is held in the rdb (for use by the gateway)
 .rdb.setpartition[]
+
+/-change timeout to zero before eod flush
+.rdb.timeoutreset[]
+.timer.repeat[.eodtime.nextroll-1000000000*60;0W;1D00:00:00.000000000;
```
ThomasSmyth: Same as this?
```suggestion
.timer.repeat[.eodtime.nextroll-00:01;0W;1D;
```

code/processes/rdb.q
```diff
@@ -3,6 +3,7 @@
 /-changes added 
 /-Can specify the hdb directory rather than relying on the tickerplant
 
+
```
ThomasSmyth: 🤔

robsketch: back to normal sorry

robsketch:
Test when `reloadenabled` is true and RDB offloads to WDB:

Just have to add the timeout change in the RDB in its `reload` function.  This reload function is called by the WDB in order to wipe memory from RDB.

```
q)h:hopen`::6002:admin:admin
q)h"\\T"
180i
```
Check that timeout is changed prior to EOD memory wipe:
```
q)h"\\T"
0i
q)h".rdb.reloadenabled"
1b
```

Call end of day and see if timeout is changed back to its initial value upon completion:
```
q)h:hopen`::6000:admin:admin
q)h".u.end[.z.d]"
q)h:hopen`::6002:admin:admin
q)h"\\T"
180i
```

So the `reloadenabled` case is now covered

code/processes/rdb.q
```diff
@@ -97,6 +97,8 @@ endofday:{[date]
 	a:{(x;raze exec {(enlist x)!enlist((#);enlist y;x)}'[c;a] from meta x where not null a)}each tables`.;
 	/-save and wipe the tables
 	writedown[hdbdir;date];
+	/-reset timeout to original timeout
+        restoretimeout[];
```
jonnypress: funny indentation

code/processes/rdb.q
```diff
@@ -110,6 +112,8 @@ endofday:{[date]
 	
 reload:{[date]
 	.lg.o[`reload;"reload command has been called remotely"];
+	/-set timeout to be 0 before tables are wiped
+	system"T 0";
```
jonnypress: You can't do this here and have it take effect ... but I think this is un-necessary? The timer will set it to 0, and it will then be set back to what it should've been when the reload is complete

code/processes/rdb.q
```diff
@@ -166,6 +172,9 @@ getpartition:{[] rdbpartition}
 notpconnected:{[]
 	0 = count select from .sub.SUBSCRIPTIONS where proctype in .rdb.tickerplanttypes, active}
 
+/-resets timeout to 0 before EOD writedown
+timeoutreset:{.rdb.timeout:system"T";system"T 0"};
+restoretimeout:{system["T ", string timeout]};
```
jonnypress: For clarity, i would call this .rdb.timeout rather than just timeout (given that in the line above you have referenced .rdb.timeout)

PR #251, opened 2020-04-01
RDB and GW bug fix

jonnypress:
Bug fixes for RDB and Gateway. Only subscribe once in RDB, remove non-generic attribute check from GW.

### DIFF
--- a/code/processes/rdb.q
+++ b/code/processes/rdb.q
@@ -192,8 +192,6 @@ reload:.rdb.reload
 
 .lg.o[`init;"searching for servers"];
 .servers.startup[];
-/-subscribe to the tickerplant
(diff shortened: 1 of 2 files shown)


code/processes/rdb.q
```diff
@@ -192,8 +192,6 @@ reload:.rdb.reload
 
 .lg.o[`init;"searching for servers"];
 .servers.startup[];
```
jonathonmcmurray, line 194: should we also get rid of this `.servers.startup` given we have `.servers.startupdepcycles` below which should do the same thing I think?

jonnypress: Yip good point - will do that.

PR #256, opened 2020-04-16
changed startup function 

mg-aquaq:
changed .server.startup to .server.startupdependent to ensure connection to dqedb

### DIFF
--- a/code/processes/dqc.q
+++ b/code/processes/dqc.q
@@ -17,7 +17,7 @@ compcounter:([id:`long$()]counter:`long$();procs:();results:());
 
 init:{                                                                                                          /- this function gets called at every EOD by .u.end
   .lg.o[`init;"searching for servers"];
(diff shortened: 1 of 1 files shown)


code/processes/dqc.q
```diff
@@ -17,7 +17,7 @@ compcounter:([id:`long$()]counter:`long$();procs:();results:());
 
 init:{                                                                                                          /- this function gets called at every EOD by .u.end
   .lg.o[`init;"searching for servers"];
-  .servers.startup[];                                                                                           /- Open connection to discovery
+  .servers.startupdependent[enlist `dqedb; 30];                                                                 /- Open connection to discovery. Retry until connected to dqe.
```
ThomasSmyth, line 20: Don't think this needs to be enlisted, _seemed_ to work fine for me as an atom. Timeout probably fine :+1:

PR #258, opened 2020-04-22
Add symdir optional parameter to loadparams

robsketch:
## Add symdir optional parameter to loadparams

The `symdir` parameter will allow you to specify a single location for the data to be enumerated against.  At the moment data will always be enumerated to the `dbdir` parameter, and you may want to specify a single directory to enumerate against while loading in multiple files that belong to the same data set.

# Checklist:
- [x] Add logic for if `symdir` is included in `loadparams` to determine where to enumerate
- [x] Add `symdir` to the list `reqtypes` for type checking
- [x] Testing


jonnypress:
Can you add this option to the documentation as well please (docs/utilities.md). There's a table of params for dataloader.q

PR #264, opened 2020-04-30
Edited torq.sh to fix port truncation issue

### LINKED ISSUE #257: ./torq.sh stop all - kill command fails
When running `./torq.sh stop all` or any variation of `./torq.sh stop`, the following error is output for each running process:
`kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]`
checking the status using `summary` shows the processes are still running.

From what I can see this is originating from line 142 in `torq.sh` - debugging the kill command shows there is no output from `lsof -i :$port -sTCP:LISTEN | awk '{if(NR>1)print $2}'`

This is failing because the line above:
`port=$(($(eval echo \$"$(getfield "$procno" port)")))`
seems to truncate the value of the port number. I'm not sure whether `getfield` is causing this when pulling from the config or not.

My appconfig/process.csv is very basic:
```
host,port,proctype,procname,U,localtime,g,T,w,load,startwithall,extras,qcmd
testhost,9998,tickerplant,tp_1,,1,1,3,,,1,,q
testhost,9997,rdb,rdb_1,,1,1,60,4000,,1,,q
testhost,9996,hdb,hdb_1,,1,1,60,4000,,1,,q
```

Adding an `echo $port` and running `./torq.sh stop all` with processes running outputs the following:
```
00:09:32 | Shutting down tp_1...
998
kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]
00:09:32 | Shutting down rdb_1...
997
kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]
00:09:32 | Shutting down hdb_1...
996
kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]
```

I'm running 64 bit Kubuntu 18.04.4 and bash 4.4.20

Is this something I'm doing wrong?

InDeNiall104:
In regards to ./torq.sh stop all - kill command fails #257 
Upon looking into the issue these are the steps I took to try and fix the problem.

Original process.csv
image

Error replication
err rep

The first thing I investigated was the default process.csv upon cloning TorQ and it was interesting that the ports were inputted via the {KDBBASEPORT} variable set in setenv.sh

Default process.csv
image

After reformatting the custom process.csv and defining the host to be localhost and the desired base port to 9998 in setenv.sh and upon running ./torq.sh start all, ./torq.sh summary and ./torq.sh stop all it can be seen that they execute fine and with no truncation of the ports occuring.

Changing the custom process.csv
image

Setting the KDBBASEPORT in setenv.sh
image
Running the ./torq.sh scripts
var fix

Next I tried to find what exactly was causing the truncation of the port if it was hard coded in e.g. 9998.. etc
It turns out the **port=$(($(eval echo \$"$(getfield "$procno" port)")))** at line 140 (was line 142 in older version) of the torq.sh script the eval echo section was truncating the port number.

image

So to get around this I just simplified the port variable from **port=$(($(eval echo \$"$(getfield "$procno" port)")))** to **port=$(getfield "$procno" port)** and this just outputs the full hard coded port unchanged.

However now the ./torq.sh start all and ./torq.sh stop all only works for hard coded ports and not for the default {KDBBASEPORT} method so for this to handle both formats I needed to alter the stop () function section of the torq.sh script.
After some tinkering I was able to successfully refactor the stop () function in the torq.sh script and it seems to be working perfectly now with no errors. The main part that helped fix the issue was the following:

image

This first set the port to the full unchanged port entered in the process.csv file, it then checked to see if the port entered was not made up of numbers i.e in the {KDBBASEPORT} format. 

If it was found to be in the {KDBBASEPORT} format then to eval echo the port and this is capable of handling the port in variable form ({KDBBASEPORT}) and continues on to shutdown/kill the processes associated with the {KDBBASEPORT} variable. 

If not then the port remained in it's full unchanged hard coded form and the script then continues on and successfully shutdowns/kills the processes on their respective ports. 

Can now work with both hard coded ports and {KDBBASEPORT}'s
fix hc

This closes #257

### DIFF
--- a/torq.sh
+++ b/torq.sh
@@ -132,25 +132,21 @@ stop() {
   host=$(getfield "$procno" "host")
   if [[ $host == "localhost" || $host == `hostname -i`  ||
         ${hostips[@]}  =~ $host || ${hostnames[@]} =~ $host ]]; then
-    if [[ -z $(findproc "$1") ]]; then                                                              # check process not running
-      echo "$(date '+%H:%M:%S') | $1 is not currently running"
-    else
-      echo "$(date '+%H:%M:%S') | Shutting down $1..."
-      procno=$(awk '/,'$1',/{print NR}' "$CSVPATH")
-      port=$(($(eval echo \$"$(getfield "$procno" port)")))
-      eval "kill -15 `lsof -i :$port -sTCP:LISTEN | awk '{if(NR>1)print $2}'`"
-    fi
+    kcmd="kill -15"
   else
-    if [[ -z $(findproc "$1") ]]; then
-    echo "$(date '+%H:%M:%S') | $1 is not currently running"
-    else
-      echo "$(date '+%H:%M:%S') | Shutting down $1..."
-      procno=$(awk '/,'$1',/{print NR}' "$CSVPATH")
-      port=$(($(eval echo \$"$(getfield "$procno" port)")))
-      eval "$cmd `lsof -i :$port -sTCP:LISTEN | awk '{if(NR>1)print $2}'`"
-    fi
+    kcmd="$cmd"
   fi
-  
+  if [[ -z $(findproc "$1") ]]; then
+    echo "$(date '+%H:%M:%S') | $1 is not currently running"
+  else
+    echo "$(date '+%H:%M:%S') | Shutting down $1..."
+    procno=$(awk '/,'$1',/{print NR}' "$CSVPATH")
+    port=$(getfield "$procno" port)
+    if [[ ! $port =~ ^[0-9]+$ ]]; then
+      port=$(($(eval echo \$$port)))
+    fi
+    eval "$kcmd `lsof -i :$port -sTCP:LISTEN | awk '{if(NR>1)print $2}'`"
+  fi
  }
 
 getall() {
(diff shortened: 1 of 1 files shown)


torq.sh
```diff
@@ -132,25 +132,21 @@ stop() {
...
-    else
-      echo "$(date '+%H:%M:%S') | Shutting down $1..."
-      procno=$(awk '/,'$1',/{print NR}' "$CSVPATH")
-      port=$(($(eval echo \$"$(getfield "$procno" port)")))
-      eval "$cmd `lsof -i :$port -sTCP:LISTEN | awk '{if(NR>1)print $2}'`"
+  else
+    echo "$(date '+%H:%M:%S') | Shutting down $1..."
+    procno=$(awk '/,'$1',/{print NR}' "$CSVPATH")
+    port=$(getfield "$procno" port)
+    if [[ ! $port =~ ^[0-9]+$ ]]; then
```
ThomasSmyth, line 145: Just need a comment explaining this edition, this is probably sufficient:
```suggestion
    # check if port is numerical, if not it is an expression that needs evaluated
    if [[ ! $port =~ ^[0-9]+$ ]]; then
```

PR #266, opened 2020-05-05
Change k4unit.q so that it works across all versions of kdb

mg-aquaq:
k4unit.q currently uses 
```
\ts *lang*)*code*
```
to run q and k code, but this no longer works in kdb 4.0. To fix this, we have changed it to 
```
*lang*)\ts *code*
```

### DIFF
--- a/tests/k4unit.q
+++ b/tests/k4unit.q
@@ -85,13 +85,13 @@ KUrt:{ / (run tests) - run contents of KUT, save results to KUTR
 	neg before-count KUTR}
 
 KUpexec:{[prefix;lang;code;repeat;allowfail]
-	s:prefix,(string lang),")",$[1=repeat;string code;"do[",(string repeat),";",(string code),"]"];
+	s:prefix,$[1=repeat;string code;"do[",(string repeat),";",(string code),"]"];
 	if[1<.KU.VERBOSE;.lg.o[`k4unit;s]];$[.KU.DEBUG&allowfail;value s;@[value;s;`FA1L]]}
 
 KUfailed:{`FA1L~x}
 
 KUexec:{[lang;code;repeat]
-	value $[1=repeat;string code;"do[",(string repeat),";",(string code),"]"]}
+	value$[1=repeat;string code;"do[",(string repeat),";",(string code),"]"]}
 
 KUact:{[action;lang;code;repeat;ms;bytes;file]
 	msx:0;bytesx:0j;ok:okms:okbytes:valid:0b;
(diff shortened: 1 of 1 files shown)


mg-aquaq:
Going forward with kdb 4.0, the k4unit tests won't work due to the error mentioned above. This is a quick fix that removes the `lang` variable from `KUexec`, but if this is the right direction then we can work on removing the `lang` variable from the entire script.

tests/k4unit.q
```diff
@@ -91,7 +91,7 @@ KUpexec:{[prefix;lang;code;repeat;allowfail]
 KUfailed:{`FA1L~x}
 
 KUexec:{[lang;code;repeat]
-	value(string lang),")",$[1=repeat;string code;"do[",(string repeat),";",(string code),"]"]}
+	value $[1=repeat;string code;"do[",(string repeat),";",(string code),"]"]}
```
jonnypress, line 95: I think you also have to fix KUpexec as well

jonathonmcmurray:
One thing to mention - if `lang` is going to be ignored, I think we should make it an optional field in test CSVs (no point forcing people to add a field that has no effect, but if we completely remove it it'd break any existing CSVs, which should be avoided).

Might be an idea to inspect the header row of CSV & generate the string of type chars for reading whole CSV based on the headers present? (would also mean that the order of fields doesn't matter etc.)

jonathonmcmurray:
Actually, I think perhaps we can still support `lang` if we just change the order of things:

```
(base) jmcmurray@homer ~ $ q
KDB+ 4.0 2020.03.30 Copyright (C) 1993-2020 Kx Systems
l64/ 24()core 128387MB jmcmurray homer 127.0.1.1 EXPIRE 2020.06.30 AquaQ #55345

q)k)\ts a:#:1 2 3
0 992
q)a
3
```

i.e. putting the lang prefix *before* `\ts` seems to work fine in 4.0

mg-aquaq:
> Actually, I think perhaps we can still support `lang` if we just change the order of things:
> 
> ```
> (base) jmcmurray@homer ~ $ q
> KDB+ 4.0 2020.03.30 Copyright (C) 1993-2020 Kx Systems
> l64/ 24()core 128387MB jmcmurray homer 127.0.1.1 EXPIRE 2020.06.30 AquaQ #55345
> 
> q)k)\ts a:#:1 2 3
> 0 992
> q)a
> 3
> ```
> 
> i.e. putting the lang prefix _before_ `\ts` seems to work fine in 4.0

This seems to work well with all versions, so I think this is the best course of action

mg-aquaq:
Did some tests with both q and k code and this fix seems to work just fine

jonathonmcmurray:
Can you update the title & description of the PR? (given these will become the commit message when merged & they are no longer accurate)

mg-aquaq:
@jonathonmcmurray I think this is ready to go.

PR #273, opened 2020-05-13
Fix output to show full expected and actual schemas

mg-aquaq:
Error messages were only showing some of the attributes that were wrong,
now show full actual and expected attributes for clarity.


mg-aquaq:
Functionality of check has not been changed, just how on of the error messages behaves.

PR #274, opened 2020-05-14
Make output from schemacheck more descriptive

mg-aquaq:
We have improved the output of schemacheck so that it shows the columns whose meta does not match expectation, as well as the actual and expected schema of those columns.

### DIFF
--- a/code/dqc/schemacheck.q
+++ b/code/dqc/schemacheck.q
@@ -9,5 +9,5 @@ schemacheck:{[tab;colname;types;forkeys;attribute]
   f: {@[x;where 0=count each string x;{`$"_"}]};
   $[all c:((flip origschema)each key flip origschema)~'(flip checkschema)each key flip checkschema;
     (1b;"Schema of ",(string tab)," matched proposed schema");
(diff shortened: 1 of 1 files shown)


code/dqc/schemacheck.q
```diff
@@ -5,7 +5,9 @@ schemacheck:{[tab;colname;types;forkeys;attribute]
   .lg.o[`dqc;"checking schema of table mathces expectation"];
   origschema:0!meta tab;
   checkschema:([]c:colname;t:types;f:forkeys;a:attribute);
-  $[all c:((flip origschema)each key flip origschema)~'(flip checkschema)each key flip checkschema;
+  / Function to replace empty symbols with underscores in error messages.
+  f: {@[x;where 0=count each string x;{`$"_"}]};
```
ThomasSmyth, line 9: Would it make more sense to use `.Q.s1` to add backticks in front of column names?
```
q).Q.s1`a``b
"`a``b"
q).Q.s1'[`a``b]
"`a"
,"`"
"`b"
```

mg-aquaq: I'm not sure, the problem was replacing empty symbols with underscores so that we could show where there was no attribute on a column

ThomasSmyth: Okay then I think having backticks for all is clearer than underscores.

PR #275, opened 2020-05-18
dqe - Reference Data Check 

tszwailam:
Comparing two column in two separate tables. 
A column in the table to be checked would be compared to the reference column of the reference table.

### DIFF
--- a/code/dqc/refdatacheck.q
+++ b/code/dqc/refdatacheck.q
@@ -1,0 +1,10 @@
+\d .dqc
+
+/- Check whether the referenced column of a table is in another column of another table
+refdatacheck:{[tablea;tableb;cola;colb]
+  .lg.o[`dqe;"checking whether refrence date is covered in the other column"];
+  $[all tablea[cola] in tableb[colb];
+    (1b;"All data from ",(string cola)," of ",(string tablea), "exists in ",(string colb), " of ",string tableb);
+    (0b;"The following data did not exist in",(string colb), " of ",(string tableb),": ",string tablea[cola][where not (tablea[cola] in tableb[colb])])
+]
+}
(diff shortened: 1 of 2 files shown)


code/dqc/refdatacheck.q
```diff
@@ -0,0 +1,9 @@
+\d .dqc
+
+/- Check whether the referenced column of a table is in another column of another table
+refdatacheck:{[tablea;tableb;cola;colb]
+  .lg.o[`dqe;"checking whether refrence data is covered in the other column"];
+  $[all tablea[cola] in tableb[colb];
+    (1b;"All data from ",(string cola)," of ",(string tablea), "exists in ",(string colb), " of ",string tableb);
+    (0b;"The following data did not exist in",(string colb), " of ",(string tableb),": ",string tablea[cola][where not (tablea[cola] in tableb[colb])])]
```
ThomasSmyth, line 8: Can slim this a little and add some logging:
```suggestion
  msg:$[c:all r:tablea[cola]in tableb colb;
    "All data from ",(string cola)," of ",(string tablea), "exists in ",(string colb), " of ",string tableb;
    "The following data did not exist in",(string colb), " of ",(string tableb),": ",string tablea[cola]where not r];
    .lg.o[`dqe]msg;
    (c;msg)
```
Can also change the id from `dqe` to something that ties all log messages in this function together, maybe `refdatacheck`.

tszwailam: For the suggested change on changing the id from `dqe` to lets say `refdatacheck`, would you want that to be changed on all functions? and would you want to change it on line 5 too?

ThomasSmyth: Best just concentrate on this function for now but yes all log lines in this function should have the same id.

PR #279, opened 2020-05-21
Add release notes to README.md

mg-aquaq:
We forgot to add the release notes to the README.


mg-aquaq:
Forgot to turn my trusty whitespace remover off, so apologies, but there is an option in GitHub to view non-whitespace changes.

PR #280, opened 2020-05-21
readability changes/log fix for symfilegrowth/refdatacheck

tszwailam:
Error in refdatacheck log msg fixed - string also fixed
a space in symfilegrowth's description for readability


code/dqc/refdatacheck.q
```diff
@@ -2,10 +2,10 @@
 
 /- Check whether the referenced column of a table is in another column of another table
 refdatacheck:{[tablea;tableb;cola;colb]
-  .lg.o[`refdatacheck;"checking whether refrence data is covered in the other column"];
+  .lg.o[`refdatacheck;"checking whether reference data is covered in the other column"];
   msg:$[c:all r:tablea[cola]in tableb colb;
```
ThomasSmyth, line 6: If not using `msg` any more can remove the variable.

mg-aquaq: I believe it is still being used to return the msg at the end of the function, after logging

PR #282, opened 2020-05-26
Adding Nullcount query to dqe

tszwailam:
nullcount query keeps track of the number of nulls in a table in a partitioned HDB

To add to that, wondering if I should add a column variable to the function for it to work on a specific column just like infinitycount, or to leave i to count number of nulls in a table like nullchk in the dqc.

### DIFF
--- a/code/dqe/nullcount.q
+++ b/code/dqe/nullcount.q
@@ -1,7 +1,8 @@
 \d .dqe
 
 /- Given a table name as a symbol (tn), returns the number of nulls in tn.
+/- Works on partitioned tables in an hdb
 nullcount:{[tn]
   .lg.o[`nullcount;"Getting count of nulls in ",string tn];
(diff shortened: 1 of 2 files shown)


code/dqe/nullcount.q
```diff
@@ -0,0 +1,8 @@
+\d .dqe
+
+/- Given a table name as a symbol (tn), returns the number of nulls in tn.
+/- Works on partitioned tables in an hdb
+nullcount:{[tn]
```
drgdavies, line 5: Do we need to also consider columns as possibility here?
Could add column and where null it does whole table?

mg-aquaq: If we return a dictionary with multiple things to the DQE, will it add them as multiple entries in the resultstab table? If so, we could do this easily. Or would we have to call this function once for each column?

PR #283, opened 2020-05-27
Add datadog functionality to DQC process

mg-aquaq:
Send metrics to datadog when a check is completed.


mg-aquaq:
Wasn't expecting this to be approved, still need to make a few changes.

PR #286, opened 2020-06-04
Add check for the differences between time columns

mg-aquaq:
This can be used to measure latency between timestamps from the exchange and timestamps from the ticker plant.

### DIFF
--- a/code/dqc/timediff.q
+++ b/code/dqc/timediff.q
@@ -6,7 +6,7 @@ timediff:{[tn;ca;cb;pt;tl]
   .lg.o[`timediff;"Checking the time differences in columns ",(", "sv string(ca;cb))," of table ",(string tn)];
   ot:$[tl>re:1-(sum a)%count a:tl>(tn ca)-tn cb;
     (1b;"No major problem with data flow");
-    (0b;"ERROR: ",(string re*100),"% of differences between columns ",(string ca),", ",m:(string cb)," are more than ",(string tl)," milliseconds.")
+    (0b;"ERROR: ",(string re*100),"% of differences between columns ",(string ca),", ",m:(string cb)," are greater than the timespan ",(string tl),".")
     ];
   .lg.o[`timediff;ot 1];
   ot
--- a/code/dqc/timediff.q
+++ b/code/dqc/timediff.q
@@ -1,0 +1,12 @@
+\d .dqc
+
+/ Takes a table name tn and two column names ca and cb, as well as a percentage
+/ pt and a tolerance tl in milliseconds.
+timediff:{[tn;ca;cb;pt;tl]
+  .lg.o[`timediff;"HELLO"];
+  .lg.o[`timediff;"Checking the time differences in columns ",(", "sv string(ca;cb))," of table ",(string tn)];
+  $[tl>re:1-(sum a)%count a:?[tn;();();(>;(0D+tl*00:00:00.001);(-;ca;cb))];
+    (1b;"No major problem with data flow");
+    (0b;"ERROR: ",(string re*100),"% of differences between columns ",(string ca),", ",m:(string cb)," are more than ",(string tl)," milliseconds.")
+    ]
+  }
--- a/code/dqc/timediff.q
+++ b/code/dqc/timediff.q
@@ -4,7 +4,7 @@
 / pt and a tolerance tl as a timespan (eg 0D00:00:00.001).
 timediff:{[tn;ca;cb;pt;tl]
   .lg.o[`timediff;"Checking the time differences in columns ",(", "sv string(ca;cb))," of table ",(string tn)];
-  ot:$[tl>re:1-(sum a)%count a:tl>(tn ca)-tn cb;
+  ot:$[pt>re:1-(sum a)%count a:tl>(tn ca)-tn cb;
     (1b;"No major problem with data flow");
     (0b;"ERROR: ",(string re*100),"% of differences between columns ",(string ca),", ",m:(string cb)," are greater than the timespan ",(string tl),".")
     ];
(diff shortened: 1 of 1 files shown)


mg-aquaq:
There's no config for this, as I tested it with a different version of quote to the one in TorQ, where I added an "exchange time" column.

code/dqc/timediff.q
```diff
@@ -0,0 +1,12 @@
+\d .dqc
+
+/ Takes a table name tn and two column names ca and cb, as well as a percentage
+/ pt and a tolerance tl in milliseconds.
+timediff:{[tn;ca;cb;pt;tl]
+  .lg.o[`timediff;"HELLO"];
```
ThomasSmyth, line 6: For debugging?

code/dqc/timediff.q
```diff
@@ -0,0 +1,12 @@
...
+  .lg.o[`timediff;"Checking the time differences in columns ",(", "sv string(ca;cb))," of table ",(string tn)];
+  $[tl>re:1-(sum a)%count a:?[tn;();();(>;(0D+tl*00:00:00.001);(-;ca;cb))];
+    (1b;"No major problem with data flow");
```
ThomasSmyth, line 9: Adding logging for these output messages too.

code/dqc/timediff.q
mg-aquaq: oops, shouldn't have let that slip through

code/dqc/timediff.q
mg-aquaq: As in.. save the result to a variable, log it, then return  the variable?

code/dqc/timediff.q
```diff
@@ -0,0 +1,12 @@
...
+    (0b;"ERROR: ",(string re*100),"% of differences between columns ",(string ca),", ",m:(string cb)," are more than ",(string tl)," milliseconds.")
+    ]
```
ThomasSmyth, line 11: Something like this should do
```suggestion
  msg:$[c:tl>re:1-(sum a)%count a:?[tn;();();(>;(0D+tl*00:00:00.001);(-;ca;cb))];
    "No major problem with data flow";
    "ERROR: ",(string re*100),"% of differences between columns ",(string ca),", ",m:(string cb)," are more than ",(string tl)," milliseconds."
    ];
    .lg.o[`timediff;msg];
    (c;msg)
```

code/dqc/timediff.q
```diff
@@ -0,0 +1,13 @@
+\d .dqc
+
+/ Takes a table name tn and two column names ca and cb, as well as a percentage
+/ pt and a tolerance tl in milliseconds.
+timediff:{[tn;ca;cb;pt;tl]
+  .lg.o[`timediff;"Checking the time differences in columns ",(", "sv string(ca;cb))," of table ",(string tn)];
+  ot:$[tl>re:1-(sum a)%count a:?[tn;();();(>;(0D+tl*00:00:00.001);(-;ca;cb))];
```
jonnypress, line 7: ?[tn;();();(>;(0D+tl*00:00:00.001);(-;ca;cb))]
What is this... an exec? 
Does this work on an ondisk table? Does it need to?

code/dqc/timediff.q
```diff
@@ -0,0 +1,13 @@
+\d .dqc
+
+/ Takes a table name tn and two column names ca and cb, as well as a percentage
+/ pt and a tolerance tl as a timespan (eg 0D00:00:00.001).
+timediff:{[tn;ca;cb;pt;tl]
+  .lg.o[`timediff;"Checking the time differences in columns ",(", "sv string(ca;cb))," of table ",(string tn)];
+  ot:$[tl>re:1-(sum a)%count a:tl>(tn ca)-tn cb;
+    (1b;"No major problem with data flow");
+    (0b;"ERROR: ",(string re*100),"% of differences between columns ",(string ca),", ",m:(string cb)," are more than ",(string tl)," milliseconds.")
```
jonathonmcmurray, line 9: does the "miiliseconds" at the end of the message still make sense if `t1` is a timespan now? Might be OK to just omit that

mg-aquaq: You're right, it should always be a time span now, will change that.

code/dqc/timediff.q
```diff
@@ -0,0 +1,13 @@
+\d .dqc
+
+/ Takes a table name tn and two column names ca and cb, as well as a percentage
+/ pt and a tolerance tl as a timespan (eg 0D00:00:00.001).
+timediff:{[tn;ca;cb;pt;tl]
+  .lg.o[`timediff;"Checking the time differences in columns ",(", "sv string(ca;cb))," of table ",(string tn)];
+  ot:$[tl>re:1-(sum a)%count a:tl>(tn ca)-tn cb;
```
jonathonmcmurray, line 7: is this comparison correct? if I'm reading it right, you're comparing the tolerance timespace to the percentage of times that differ by greater than the tolerance. Should this be

```suggestion
  ot:$[pt>re:1-(sum a)%count a:tl>(tn ca)-tn cb;
```

?

code/dqc/timediff.q
```diff
@@ -0,0 +1,13 @@
+\d .dqc
+
+/ Takes a table name tn and two column names ca and cb, as well as a percentage
+/ pt and a tolerance tl as a timespan (eg 0D00:00:00.001).
+timediff:{[tn;ca;cb;pt;tl]
+  .lg.o[`timediff;"Checking the time differences in columns ",(", "sv string(ca;cb))," of table ",(string tn)];
+  ot:$[pt>re:1-(sum a)%count a:tl>(tn ca)-tn cb;
```
jonathonmcmurray, line 7: If you switch the more than to a less than, can you get rid of the `1-` like this?

```suggestion
  ot:$[pt>re:(sum a)%count a:tl<(tn ca)-tn cb;
```

mg-aquaq: Probably, but I've just merged it

PR #290, opened 2020-06-10
Set RDB EOD attributes on the GW only is reloadenabled is true

pauljloughran:
The RDB attributes are not being updated on the GW whilst the EOD sort is being completed.
This means that that the GW thinks that only data for T-1 is avaiable in the RDB instead of T-1 and T.


jonnypress:
Thanks Paul!

PR #302, opened 2020-07-10
Add top function to torq.sh

robsketch:
As the title suggests, this PR adds an option called `top` to display the statistics found in `top.q` which is available in kdb 4.0

More info regarding the code profiler can be found [here](https://code.kx.com/q/kb/profiler/).

Runs using `./torq.sh top <processname>`


robsketch:
Use case:
```
rsketch@homer:~/deploy$ ./torq.sh top rdb1
total self name                   file                                       ..
-----------------------------------------------------------------------------..
100   100  "\\sleep 10"           ""                                         ..
100   0    "code/processes/rdb.q" ""                                         ..
100   0    "rocesses/rdb.q:202:0" "/home/rsketch/test1/code/processes/rdb.q" ..
100   0    "h/test1/torq.q:615:0" "/home/rsketch/test1/torq.q"               ..
100   0    ".os.sleep"            "/home/rsketch/test1/code/common/os.q"     ..
100   0    ".proc.loadf0"         "/home/rsketch/test1/torq.q"               ..
100   0    "ers.startupdepcycles" "/home/rsketch/test1/code/handlers/trackser..
\\
rsketch@homer:~/deploy$
```

top.q
```diff
@@ -0,0 +1,13 @@
+if[2=count .z.x;(`$"::",.z.x 0).z.i;system"l ",.z.x 1;exit 0]
```
jonnypress, line 1: Can you add a reference line to the top of this i.e. where it comes from in code.kx.com
Can you also add something which makes this exit with a verbose error message if somebodies tries to run this with .z.K < 4

torq.sh
```diff
@@ -127,6 +127,15 @@ summary() {
   fi
  }
 
+top() {
+  pid=$(findproc "$1");
+  if [[ -z $pid ]]; then
+    echo "ERROR: Cannot run top because process is not running";
+  else
+    q top.q $pid;
```
jonnypress, line 135: It would be good to extend this script so that we have a default QCMD in the same way we have a default value for QCON, and QCMD is invoked here (would also need to change line 66 of current script to also use the default)

top.q
```diff
@@ -0,0 +1,13 @@
...
+i:0p;T:([name:();file:();line:();col:();text:()]total:0#0;self:0#0)
+pct:{.01*"j"$1e4*x};nm:{[n;f;l;c;t] -20 sublist$[""~n;$[""~f;t;f,":",string[l],":",string c];n]}
+top:{x xdesc `total`self xcols 0!update total:pct total%sum self,pct self%sum self,name:nm'[name;file;line;col;text]from T}
+
+.z.ts:{@[{t:``pos _ select from .Q.prf0 p where not .Q.fqk each file;
+ T+:select total:1 by name,file,line,col,text from t;
+ if[count T;T[last[t]`name`file`line`col`text;`self]+:1];
+ if[00:00:01<.z.p-i;i::.z.p;1"\033c";show top`self]};::;{system"t 0";'x}]}
+
+$[null p:"I"$.z.x 0;[system"p 0W";.z.pg:{p::x;system"p 0";system"t 10"};system"q "," "sv string[(.z.f;"j"$system"p")],.z.x];system"t 10"]
```
jonathonmcmurray, line 13: Related to Jonny's comment about `QCMD` - I think this line will fail if `QCMD` is something other than just `q` - maybe need to read env var here or something?

top.q
jonathonmcmurray: I assume this also only works on Linux, maybe add an error message as well if someone tries to run it on Windows? (not sure if it works on Mac). I know we're adding it to `torq.sh` but maybe people will see `top.q` & try to run it manually on Windows.

code/profiler/top.q
```diff
@@ -0,0 +1,16 @@
+if[`l64<>.z.o; -2"error: linux 64-bit is required";exit 0]
```
jonathonmcmurray, line 1: convention is that `exit 0` means "success" and non-zero exit code is used for failure, probably best to make this non-zero

torq.sh
```diff
@@ -18,6 +18,10 @@ if [ -z $QCON ]; then
   QCON="qcon"                                                                                       # set environment if not predefined
 fi
 
+if [ -z $QCMD ]; then                                                                               # set environment if not predefined
```
jonnypress, line 21: Update this comment - it's not environment related. Also update the comment for QCON

PR #309, opened 2020-07-22
Rdb log replaying, timezone configuration

JRobbins91:


code/common/eodtime.q
```diff
@@ -17,16 +17,11 @@ adjtime:{[p]
...
-     z:`timespan$(mod) . "j"$z, 1D;                         // force time adjust to be between 0D and 1D
-     cp+$[z <= p;z+np;z]                                    // if past time already today, make it next period
-     };
-
 // function to get time (in UTC) of next roll after UTC timestamp, p
 getroll:{[p]
-     getperiod[p;1D;"d"$p]
+     z:rolltimeoffset-adjtime[p];                           // convert rolltimeoffset from rolltimezone to UTC
+     z:`timespan$(mod) . "j"$z, 1D;                         // force time adjust to be between 0D and 1D
+     ("d"$p) + $[z <= p;z+1D;z]                             // if past time already today, make it tomorrow
```
ThomasSmyth, line 24: Can remove the conditional by using the comparison to determine whether to add a day:
```suggestion
     ("d"$p) + $[z <= p;z+1D;z]                             // if past time already today, make it tomorrow
     ("d"$p)+1D*z<=p                                        // if past time already today, make it tomorrow
```

code/processes/segmentedtickerplant.q
```diff
@@ -11,40 +11,48 @@
...
 
 init:{[b]
   .u.upd:{[b;t;x]
-    .stplg.totalmsgcount+:1;
     // Type check allows update messages to contain multiple tables/data
-    $[0h<type t;.stplg.upd[b]'[t;x];.stplg.upd[b][t;x]]
+    $[0h<type t;
+      [.stplg.upd[b]'[t;x];.stplg.totalmsgcount+:count t];
+      [.stplg.upd[b][t;x];.stplg.totalmsgcount+:1]
+    ];
```
ThomasSmyth, line 41: Is `x` a list too? Can probably simplify to:
```suggestion
   .stplg.upd[b]'[(),t;(),x];
   .stplg.totalmsgcount+:count t;
```

code/segmentedtickerplant/stplog.q
```diff
@@ -11,26 +17,36 @@ createdld:{[name;date]
   ]
  };
 
-// Functions to generate log names in one of three modes //////////////////////////////
+// Functions to generate log names in one of five modes ///////////////////////////////
```
ThomasSmyth, line 20: Think I commented elsewhere but I think this reference to the number of functions should be removed so it doesn't need updated with each new function that is added.

code/segmentedtickerplant/pubsub.q
```diff
@@ -22,40 +22,40 @@ endp:{
...
-  (x;$[99=type v:value x;v;0#v])
  };
 
 // Add handle to subscriber in sub filtered mode
+// Where clause and column filters are parsed before adding to subrequestfiltered table
 addfiltered:{[x;y]
-  if[not .z.w in subrequestfiltered[x]`handle;
-    @[`.stpps.subrequestfiltered;x;:;(enlist[`handle]!enlist .z.w),y[x]]];
-  (x;$[99=type v:value x;v;0#v])
+  filts:$[null y[x]`filts;();enlist parse string y[x]`filts];
```
jonnypress, line 50: Could we make the parsing optional on type? So parse it if it's a string, if it's a mixed list just treat it as a functional form where clause (at the user's risk). Some q based subscribers might prefer that approach. Also, need to error trap the parse to protect against nonsense being inserted.

code/segmentedtickerplant/pubsub.q
```diff
@@ -102,7 +100,10 @@ closesub:{[h]
 // Subscriber will call with null y parameter in sub all mode
 // In sub filtered mode, y will contain tables to subscribe to and filters to apply
 .u.sub:{[x;y]
-  if[not x in .stpps.t;'x];
+  if[not x in .stpps.t;
+    .lg.e[`rdb;"Table ",string[x]," not in list of stp pub/sub tables"];
+    :()
```
jonnypress, line 105: I think you should return this error to the calling process rather than just logging an error
Give this is .u.sub, you need to mimic the return values of the old .u.sub

PR #310, opened 2020-08-04
added timecheck on each upd; made upd/ts functions more extendible

jonnypress:
- added end-of-day/end-of-period check on each upd
- make upd/zts/logname dictionaries more easily extendible
- change log file naming to be on seconds rather than minutes
- read in logmeta table on startup, and set log file counter to be max value from that table
- initialise directory name from schema name rather than hardcoded
- default the timer to 1000ms if not set
- remove totalmsgcount (not needed)
- synchronise updates of msgcount with flush to log file
- add rowcounts per table
- revert autobatch to be one log file message one update per batch
- make errorlog name configurable (avoid clashes with table names)
- fix for closelog to suppress incorrect log messages being printed
- fix for .u.end publish (one param rather than two)
- add baseline config

### DIFF
--- a/code/processes/segmentedtickerplant.q
+++ b/code/processes/segmentedtickerplant.q
@@ -7,6 +7,8 @@
 // Filtered - apply filters to published data, filters defined on client side
 
 .proc.loadf[getenv[`KDBCODE],"/common/os.q"];
+.proc.loadf[getenv[`KDBCODE],"/common/timezone.q"];
+.proc.loadf[getenv[`KDBCODE],"/common/eodtime.q"];
 
 // Load schema
 $[`schemafile in key .proc.params;
@@ -20,8 +22,8 @@ $[`schemafile in key .proc.params;
 
 // updtab stores functions to add/modify columns
 // Default functions timestamp updates
-// TO DO - add function to load type-sepcified updtab funcs
-@[`.stpps.updtab;.stpps.t;:;{(enlist(count first x)#.z.p+.eodtime.dailyadj),x}];
+// Preserve any prior definitions, but default all tables if not specified
+.stplg.updtab:(.stpps.t!(count .stpps.t)#{(enlist(count first x)#y),x}),.stplg.updtab
 
 // In none or tabular mode, intraday rolling not required
 if[.stplg.multilog in `none`tabular;.stplg.multilogperiod:1D];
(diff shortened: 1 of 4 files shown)


code/processes/segmentedtickerplant.q
```diff
@@ -33,21 +35,32 @@ if[.stplg.multilog~`custom;
...
+   .stplg.ts now:.z.p+.eodtime.dailyadj; 
+   .stplg.checkends now};
   // Error mode - write failed updates to separate TP log
   if[.stplg.errmode;
     .stplg.openlogerr[.stplg.dldir];
     .stp.upd:.u.upd;
     .u.upd:{[t;x] .[.stp.upd;(t;x);{.stplg.badmsg[x;y;z]}[;t;x]]}
   ];
+  // default the timer if not set
+  if[not system"t"; system"t 1000"];
```
ThomasSmyth, line 63: Worth defining a default timer value at the top of this file/in config to make it easier to update?

code/processes/segmentedtickerplant.q
```diff
@@ -33,21 +35,32 @@ if[.stplg.multilog~`custom;
...
+    .stplg.seqnum+:1;
     @[`.stplg.msgcount;t;+;1];
-  }[b;;];
-  .z.ts:.stplg.zts[b];
+  };
+  // set .z.ts to execute the timer func and then check for end-of-period/end-of-day
+  .stplg.ts:.stplg.zts[b];
+  .z.ts:{
+   .stplg.ts now:.z.p+.eodtime.dailyadj; 
+   .stplg.checkends now};
```
ThomasSmyth, line 55: One space indent :cry:

code/segmentedtickerplant/stplog.q
```diff
@@ -69,37 +74,38 @@ zts[`autobatch]:{
...
 zts[`defaultbatch]:{
-  .stpps.pub'[.stpps.t;value each .stpps.t];
-  @[`.;.stpps.t;:;.stpps.schemas[.stpps.t]];
-  ts .z.p+.eodtime.dailyadj;
+  // publish and clear all tables
+  .stpps.pubclear[.stpps.t];
  };
 
 // Immediate mode - publish and write immediately
 upd[`immediate]:{[t;x]
```
ThomasSmyth, line 91: Missing explicit argument `now`?

code/processes/segmentedtickerplant.q
jonnypress: fixed

code/processes/segmentedtickerplant.q
jonnypress: I think fine without - it's easy to do -t on command line

code/segmentedtickerplant/stplog.q
```diff
@@ -52,55 +52,67 @@ logname[`custom]:{[dir;tab;p]
...
  };
 
 zts[`autobatch]:{
   {[t]
-    x:batch[t];
-    if[count x;
-      .stpps.pub[t]'[x];
-      `..loghandles[t] (`upd;t),/:enlist each x]
+    if[count value t;
+      `..loghandles[t] enlist (`upd;t;flip value t);
```
JRobbins91, line 77: Looks like we're writing data to disk as lists for defaultbatch and immediate, but as a table for autobatch. We're publishing as tables for defaultbatch and autobatch, but lists for immediate. Would this cause any problems potentially?

jonnypress: You are right - this should be 
value flip value t
So should then be the same (lists)

code/segmentedtickerplant/pubsub.q
```diff
@@ -22,7 +22,7 @@ endp:{
 // Function to send end of day messages to subscribers      
 // Assumes that .u.end has been defined on the client side   
 end:{
-  (neg raze union/[value subrequestall;exec handle from .stpps.subrequestfiltered])@\:(`.u.end;x;y);
+  (neg raze union/[value subrequestall;exec handle from .stpps.subrequestfiltered])@\:(`.u.end;x);
  };
```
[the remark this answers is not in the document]
jonnypress: I dont think I changed this, but I think union suffices here?

PR #314, opened 2020-08-10
Pcap decoder

etennoQ:
Adds ability to parse pcap files into tables

pcaps that are dropped into the pcaps directory will be scanned by file alerter, parsed into tables and sent to the tickerplant to be added to rdb/hdb etc

Unit tests included


code/common/pcapdecoder.q
```diff
@@ -0,0 +1,87 @@
+\d .pcap
```
jonathonmcmurray, line 1: this file is in `code/common/` which means it will be loaded into (almost) every TorQ process - is this what is desired? It seems like it may only be required in the filealerter process, in which case it could be moved to `code/filealerter/` & make sure filealerter is configured to load process code (there's a config variable, I think it's `.proc.loadprocesscode:1b` or something similar)

code/common/pcapdecoder.q
```diff
@@ -0,0 +1,87 @@
...
+
+// size of headers in bytes and dict of protocol code conversions
+globheader: 24;
+packetheader: 16;
+allcodes:(enlist 6)!(enlist `TCP);
```
jonathonmcmurray, line 6: Can you include a link to the PCAP spec here as an explanation for these values?

On a side note, I assume in theory these values could be different for different versions of the PCAP spec (e.g. some future version of the spec could change these), so might be good to specify the version this is written for (maybe even have the code check the version, there's probably some header in the file with that info?)

code/common/pcapdecoder.q
```diff
@@ -0,0 +1,87 @@
...
+  filebytesize: count n;
+  gettablerow[n;]\[{y>(first x[0])+40}[;filebytesize];(),0]    
+  } read1 file;
+ 
+ data: update sym:` from data;
+ `time`sym xcols data
+ }
+
+
+gettablerow:{[n;x]  // data for a single row
```
jonathonmcmurray, line 23: Could you add a comment somewhere what `n` & `x` are? Or give them more descriptive names - most other variables seem in this script seem to have pretty good descriptive names

code/common/pcapdecoder.q
```diff
@@ -0,0 +1,87 @@
...
+ time:     first gettime[n;x];
+ flags:    getflags[n;x];
+ protocol: getprotocol[n;x];
+ 
+ totallength: (enlist 2;enlist "h")1: datafromfile[n;x;18;2];
```
jonathonmcmurray, line 28: I might be misunderstanding this line, but I think what's happening here is reading two bytes as a short - I think this might be simplified with `sv` when you're reading a single value e.g.

```q
q)(enlist 2;enlist"h")1:0x0010
16
q)type (enlist 2;enlist"h")1:0x0010
0h
q)first (enlist 2;enlist"h")1:0x0010
,16h
q)first first (enlist 2;enlist"h")1:0x0010
16h
q)0x0 sv 0x0010
16h
q)type 0x0 sv 0x0010
-5h
```

with `1:` you end up with a nested list of shorts, but with `sv` you get an atom, which probably simplifies things later on as well

https://code.kx.com/q/ref/sv/#bytes-to-integer

code/common/pcapdecoder.q
```diff
@@ -0,0 +1,87 @@
...
+ dest: ips[1];
+
+ info: getinfo[n;x];
+ srcport:  first info[0] mod 65536;
+ destport: first info[1] mod 65536;
+ seq:      first info[2] mod 4294967296;
+ ack:      first info[3] mod 4294967296;
+ win:      first info[4] mod 65536;
+ tsval:    first info[5] mod 4294967296;
+ tsecr:    first info[6] mod 4294967296;
```
jonathonmcmurray, line 47: It might be neater to do all this `mod` stuff inside the `getinfo` function & return a dictionary from that function, then just join the extra keys below in line 50. Probably also worth having a comment explaining the "magic numbers" being used here.

tests/pcap/runpcaptests.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+./torq.sh start discovery1
+./torq.sh start tickerplant1 
+./torq.sh start rdb1 -extras -.rdb.replaylog 0
+/usr/bin/rlwrap q torq.q -load ${KDBCODE}/processes/filealerter.q -proctype filealerter -procname filealerter1 -test ${KDBTESTS}/pcap -debug -.fa.decodepcaps 1 -.fa.polltime 00:10 -.servers.CONNECTIONS rdb -.fa.alreadyprocessed ${KDBTESTS}/pcap/filealerterprocessed -.fa.inputcsv ${KDBTESTS}/pcap/filealerter.csv
```
jonathonmcmurray, line 6: this hardcoded path for `rlwrap` might break depending on specific set up of linux box where script is run

config/filealerter.csv
```diff
@@ -1 +1,2 @@
-path,match,function,newonly,movetodirectory
\ No newline at end of file
+path,match,function,newonly,movetodirectory
+{KDBPCAPS},*.pcap,processpcaps,1,
```
jonathonmcmurray, line 2: this might be better placed in FSP?

tests/pcap/runpcaptests.sh
andrew-west: should it be replaced with `${RLWRAP}?`

jonathonmcmurray: I think that should work assuming `setenv.sh` has been sourced; might also be OK to just do `rlwrap` & assume it'll be on the path (this is what you're doing with `q` anwyay by the looks of it)

andrew-west: ah okay thanks, changed it to `rlwrap`

code/filealerter/pcapdecoder.q
```diff
@@ -0,0 +1,98 @@
...
+allcodes:(enlist 6)!(enlist `TCP);
+
+
+// returns table of packet data
+buildtable:{[file]
+ // check that version number of input file is 2.4
+ pcapversioncheck: all 2 0 4 0 = "h"$4#4_read1 file;
+
+ // if version is correct, gettablerow is iterated over each datapacket, extracting data
+ $[pcapversioncheck;[
```
jonathonmcmurray, line 21: Rather than this large multi-line conditional, I'd reverse the condition & make it exit early e.g.

```q
if[not pcapversioncheck;.lg.o[`alerter;"pcap version number of ", file, " is incorrect, so could not be decoded"];:()];
```

`:()` will return from the function without doing anything further. IMO this makes the function a little clearer, as atm I think it's easy to miss that a lot of it is wrapped in a conditional statement

code/filealerter/pcapdecoder.q
```diff
@@ -0,0 +1,98 @@
...
+
+// mapping from protocol id numbers to protocol name
+// add desired protocols to dictionary for decoder to be able to translate their id's
+allcodes:(enlist 6)!(enlist `TCP);
+
+
+// returns table of packet data
+buildtable:{[file]
+ // check that version number of input file is 2.4
+ pcapversioncheck: all 2 0 4 0 = "h"$4#4_read1 file;
```
jonathonmcmurray, line 18: `read1` can read a specific number of bytes from a specific location e.g.

```q
read1(file;4;4)
```

Probably a little more efficient than reading the whole file into mem & dropping/taking

BTW I think you can also skip the `"h"$`, q can compare between different datatypes:

```q
q)2 0 4 0=0x02000400
1111b
```

code/filealerter/pcapdecoder.q
```diff
@@ -0,0 +1,98 @@
...
+ totallength: 0x0 sv datafromfile[filebinary;x;18;2];
+ IPheader:    4*"J"$last string filebinary[x[0]+globheader+packetheader+16];
+ TCPheader:   4* first "0123456789abcdef"?/:string filebinary[x[0]+globheader+packetheader+48];
+ len:         (first first totallength - IPheader + TCPheader) mod 65536;
+
+ length: first raze ((enlist "h";enlist 2)1: filebinary[x[0]+36 37]) mod 65536;
+ data:   datafromfile[filebinary;x;length - len;len];
+
+ // array containing starting point for next byte and dictionary of data for current packet
+ (x[0] + length + 16;`time`src`dest`srcport`destport`protocol`flags`seq`ack`win`tsval`tsecr`length`len`data!(time;ips[`src];ips[`dest];info[`srcport];info[`destport];protocol;flags;info[`seq];info[`ack];info[`win];info[`tsval];info[`tsecr];length;len;data))
```
jonathonmcmurray, line 53:
```suggestion
 (x[0] + length + 16;ips,info,`time`length`len`data!(time;length;len;data))
```

if the column order is important:

```suggestion
 (x[0] + length + 16;`time xcols ips,info,`time`length`len`data!(time;length;len;data))
```

code/filealerter/pcapdecoder.q
```diff
@@ -0,0 +1,98 @@
...
+ protocol: getprotocol[filebinary;x];
+ info:     getinfo[filebinary;x];
+ ips:      getips[filebinary;x];
+
+ totallength: 0x0 sv datafromfile[filebinary;x;18;2];
+ IPheader:    4*"J"$last string filebinary[x[0]+globheader+packetheader+16];
+ TCPheader:   4* first "0123456789abcdef"?/:string filebinary[x[0]+globheader+packetheader+48];
+ len:         (first first totallength - IPheader + TCPheader) mod 65536;
+
+ length: first raze ((enlist "h";enlist 2)1: filebinary[x[0]+36 37]) mod 65536;
```
jonathonmcmurray, line 49: I think this one could be `0x0 sv` as well?

etennoQ: i dont think this one works the same way since the binary here is little endian so it is read using `(enlist "h";enlist 2)1:` rather than the previous big endian examples: `(enlist 2;enlist"h")1:`. Ill have a look at possible simplifications though

jonathonmcmurray: Ah, I missed the endianness, fair enough

etennoQ: sv seems to work with little endian binary as long as you reverse the binary first: 
`(0x0 sv reverse filebinary[x[0]+36 37]) mod 65536`

code/processes/filealerter.q
```diff
@@ -7,6 +7,10 @@ polltime:@[value;`.fa.polltime;0D00:00:10]							// The period to poll the file
 alreadyprocessed:@[value;`.fa.alreadyprocessed;.proc.getconfigfile["filealerterprocessed"]]	// The location of the table on disk to store the information about files which have already been processed
 skipallonstart:@[value;`.fa.skipallonstart;0b]							// Whether to skip all actions when the file alerter process starts up (so only "new" files after the processes starts will be processed) 
 moveonfail:@[value;`.fa.moveonfail;0b]								// If the processing of a file fails (by any action) then whether to move it or not regardless
+decodepcaps:@[value;`.fa.decodepcaps;0b]							// Whether to parse PCAPs found by filealerter (causes connection to tp)
```
jonnypress, line 10: I don't think this should be in here, and neither should the tickerplanttypes/tpconnssleepintv/tpcheckcycles
We don't need to build a start up dependency on the TP - we only need a connection to the TP when the file is parsed. 
So you should modify the sendtotickerplant function to take the tickerplanttype as a parameter, and create the connection at run time, and fail appropriately (e.g. print error) if it doesn't exist

code/processes/filealerter.q
```diff
@@ -171,3 +188,7 @@ loadcsv[];
 loadprocessed[alreadyprocessed];
 
 .timer.rep[.proc.cp[];0Wp;polltime;(`FArun`);0h;"run filealerter";1b]
+
+if[decodepcaps;
```
jonnypress, line 192: Don't need this bit at the bottom

code/processes/filealerter.q
```diff
@@ -41,7 +45,20 @@ find:{[path;match]
...
+	if[not decodepcaps; :()];
+	.lg.o[`alerter;"processing pcap file"];
+	.lg.o[`alerter;"decoding pcap file"];
+	table: .pcap.buildtable[hsym `$(path,"/",file)];
+	.lg.o[`alerter;"sending table to tickerplant"];
+	sendtotickerplant[`packets;table[cols table]]
+	}
+
+sendtotickerplant:{[t;x]
+	.servers.gethandlebytype[`tickerplant;`any](`.u.upd;t;x)
```
jonnypress, line 60: Pass tickerplanttype as parameter

config/settings/filealerter.q
```diff
@@ -8,4 +8,11 @@ polltime:0D00:00:10						// The period to poll the file system
 alreadyprocessed:first .proc.getconfigfile["filealerterprocessed"]	// The location of the table on disk to store the information about files which have already been processed
 skipallonstart:0b						// Whether to skip all actions when the file alerter process starts up (so only "new" files after the processes starts will be processed)
 moveonfail:0b							// If the processing of a file fails (by any action) then whether to move it or not regardless
+tickerplanttypes:`tickerplant					// List of tickerplant types to connect to
```
jonnypress, line 11: remove these

setenv.sh
```diff
@@ -21,6 +21,7 @@ export KDBAPPCODE=${TORQAPPHOME}/code
 export KDBHDB=${TORQDATA}/hdb
 export KDBWDB=${TORQDATA}/wdbhdb
 export KDBTPLOG=${TORQDATA}/tplogs
+export KDBPCAPS=${TORQDATA}/pcaps
```
jonnypress, line 24: Shouldn't be here - should be part of FSP only

code/processes/filealerter.q
```diff
@@ -41,7 +42,24 @@ find:{[path;match]
...
+	.lg.o[`alerter;"processing pcap file"];
+	.lg.o[`alerter;"decoding pcap file"];
+	table: .pcap.buildtable[hsym `$(path,"/",file)];
+	
+	.lg.o[`alerter;"checking connection to tickerplant"];
+	if[not tickerplanttype in .servers.SERVERS[`proctype]; .servers.startup[]];
+	if[not tickerplanttype in .servers.SERVERS[`proctype]; .lg.e[`alerter;"connection to tickerplant failed"]; :()];
 	
+	.lg.o[`alerter;"sending table to tickerplant"];
+	sendtotickerplant[tickerplanttype;`packets;table[cols table]]
```
jonnypress, line 57: I think the name of the table should be a parameter to the function (and that would also then influence the FSP changes)

code/processes/filealerter.q
```diff
@@ -41,7 +42,24 @@ find:{[path;match]
...
+	}
+
+sendtotickerplant:{[tptype;t;x]
+	.servers.gethandlebytype[tptype;`any](`.u.upd;t;x)
```
jonnypress, line 61: I think you should change this to make sure you get a non-null handle back

code/processes/filealerter.q
```diff
@@ -171,3 +189,5 @@ loadcsv[];
 loadprocessed[alreadyprocessed];
 
 .timer.rep[.proc.cp[];0Wp;polltime;(`FArun`);0h;"run filealerter";1b]
+
+.servers.CONNECTIONS:distinct .servers.CONNECTIONS,tickerplanttype
```
jonnypress, line 193: I would add a .servers.startup[] here

code/processes/filealerter.q
```diff
@@ -41,7 +42,24 @@ find:{[path;match]
...
+
+//-decodes pcap file and sends to tickerplant
+processpcaps:{[path;file]
+	.lg.o[`alerter;"processing pcap file"];
+	.lg.o[`alerter;"decoding pcap file"];
+	table: .pcap.buildtable[hsym `$(path,"/",file)];
+	
+	.lg.o[`alerter;"checking connection to tickerplant"];
+	if[not tickerplanttype in .servers.SERVERS[`proctype]; .servers.startup[]];
+	if[not tickerplanttype in .servers.SERVERS[`proctype]; .lg.e[`alerter;"connection to tickerplant failed"]; :()];
```
jonnypress, line 54: I think you can get rid of this section
If you have done .servers.startup on initialisation, then it should have registered an interest in the tickerplant
.servers.gethandlebytype which is used by sendtotickerplant will return handle if it has one, or re-open it if it doesn't

code/processes/filealerter.q
etennoQ: is it possible to add a third parameter to file alerter functions? the documentation seems to suggest that these functions should only take file and path since you dont specify actual parameter values in the csv

jonnypress: I think the way to do it is have this function take a third argument, and then in FSP create a project in code/filealerter which creates a new function with the table name projected

code/processes/filealerter.q
```diff
@@ -41,7 +42,23 @@ find:{[path;match]
...
+	.lg.o[`alerter;"processing pcap file"];
+	.lg.o[`alerter;"decoding pcap file"];
+	table: .pcap.buildtable[hsym `$(path,"/",file)];
 	
+	.lg.o[`alerter;"checking connection to tickerplant"];
+	sendtotickerplant[tickerplanttype;pcaptab;table[cols table]]
+	}
+
+sendtotickerplant:{[tptype;t;x]
+	if[.servers.gethandlebytype[tptype;`any]~`int$();
```
jonnypress, line 57: Could you make this 

if[count .servers.gethandlebytype[tptype;`any]

it's more supportable this way as if we were to change the handle list to be a different type for some reason then this would succeed either way

config/filealerter.csv
```diff
@@ -1 +0,0 @@
-path,match,function,newonly,movetodirectory
```
jonnypress, line 1: Why did you remove this line of headers? I think it should remain?

jonnypress (approved):

PR #315, opened 2020-08-10
Update tickerlogreplay.q to work with STP

robsketch:
- add segmented mode flag for tickerlogreplay to know its using STP
- add logic to define `logstoreplay` depending on if STP is being used
- update `realupd` to know if segmentedmode is being used or not
- tabular and none logging modes for stp now use timestamp rather than just .z.d
- stpmeta log file loaded using get rather than -11!

To do still:
- replay logs based on `stpmeta` table

### DIFF
--- a/code/processes/tickerlogreplay.q
+++ b/code/processes/tickerlogreplay.q
@@ -4,7 +4,7 @@
 
 // Variables
 firstmessage:@[value;`firstmessage;0]			// the first message to execute
-segmentedmode:@[value;`segmented;1b]		 // if using segmented tickerplant, then set to true, otherwise set false for old tickerplant
+segmentedmode:@[value;`segmentedmode;1b]		 // if using segmented tickerplant, then set to true, otherwise set false for old tickerplant
 autoreplay:@[value;`autoreplay;1b]      // replay tplogs automatically set to 1b to be backward compatible.
 lastmessage:@[value;`lastmessage;0W]			// the last message to replay
 messagechunks:@[value;`messagechunks;0W]		// the number of messages to replay at once
@@ -85,7 +85,7 @@ if[basicmode and (messagechunks within (0;-1 + 0W));
  .err.ex[`replayinit; "if using basic mode, messagechunks must not be used (it should be set to 0W). basicmode will use .Q.hdpf to overwrite tables at the end of the replay";1]];
 if[not partitiontype in `date`month`year; .err.ex[`replayinit;"partitiontype must be one of `date`month`year";1]];
 if[messagechunks=0;.err.ex[`replayinit;"messagechunks value cannot be 0";2]];
-if[segmentedmode and (0<>firstmessage or 0W<>lastmessage);.err.ex[`replayinit;"firstmessage must be 0 and lastmessage must be 0W while in segmented mode"]]
+if[segmentedmode and ((0<>firstmessage) or 0W<>lastmessage);.err.ex[`replayinit;"firstmessage must be 0 and lastmessage must be 0W while in segmented mode"];1]
 
 trackonly:messagechunks < 0 
 if[trackonly;.lg.o[`replayinit;"messagechunks value is negative - log replay progress will be tracked"]];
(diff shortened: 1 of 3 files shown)


code/processes/tickerlogreplay.q
```diff
@@ -80,12 +81,11 @@ if[`.replay.usage in key .proc.params; -1 .proc.getusage[]; exit 0];
 // Check if some variables are null
 // some must be set
 .err.exitifnull each `.replay.schemafile`.replay.hdbdir, $[all null (tplogdir;tplogfile); `.replay.tplogfile; ()];
-
 if[basicmode and (messagechunks within (0;-1 + 0W));
  .err.ex[`replayinit; "if using basic mode, messagechunks must not be used (it should be set to 0W). basicmode will use .Q.hdpf to overwrite tables at the end of the replay";1]];
 if[not partitiontype in `date`month`year; .err.ex[`replayinit;"partitiontype must be one of `date`month`year";1]];
-
 if[messagechunks=0;.err.ex[`replayinit;"messagechunks value cannot be 0";2]];
+if[segmentedmode and (0<>firstmessage or 0W<>lastmessage);.err.ex[`replayinit;"firstmessage must be 0 and lastmessage must be 0W while in segmented mode"]]
```
jonnypress, line 88: .err.ex needs a 3rd param I thnk (the exit code)
This condition is wrong due to left-of-right: (0<>firstmessage or 0W<>lastmessage)

code/processes/tickerlogreplay.q
```diff
@@ -107,16 +107,19 @@ tablestoreplay:$[tablelist~enlist`all; tables[`.]; tablelist,()]
...
-		 enlist hsym tplogfile]; 
-		[.lg.o[`replayinit;"reading log files from directory ",string tplogdir];
-		 if[()~key hsym tplogdir; .err.ex[`replayinit;"specified tplogdir ",(string tplogdir)," does not exist";4]];
-		 raze ` sv' tplogdir,/:key tplogdir:hsym tplogdir]];
+	[if[()~key hsym tplogfile; .err.ex[`replayinit;"specified tplogfile ",(string tplogfile)," does not exist";3]];
+	$[segmentedmode; raze ` sv' tplogfile,/:key tplogfile:hsym tplogfile; enlist hsym tplogfile]]; 
+	[.lg.o[`replayinit;"reading log files from directory ",string tplogdir];
+	if[()~key hsym tplogdir; .err.ex[`replayinit;"specified tplogdir ",(string tplogdir)," does not exist";4]];
+	$[segmentedmode; `$raze (string pars),/:'string key each pars:hsym`$(string[tplogdir],"/"),/:string[key hsym tplogdir],'"/"; 
+    raze ` sv' tplogdir,/:key tplogdir:hsym tplogdir]]];
```
jonnypress, line 117: I see what you are doing here, but I would probably leave this section as is, and "expand" the log directory when it comes to actually doing the replay, based on whether it is in segmented mode or not. So this bit stays as is (maybe need some minor changes) and then at replay time it expands the directory into the list of log files

code/processes/tickerlogreplay.q
```diff
@@ -211,7 +211,8 @@ replaylog:{[logfile]
          @[`.;`upd;:;.replay.initialupd]];
 	@[`.;`upd;:;.replay.realupd]];
  .replay.tablecounts:.replay.errorcounts:.replay.pathlist:()!();
- .replay.replaydate:"D"$-10#string logfile;
+ // extract date from timestamp if stp, extract date from date if tp
+ $[segmentedmode;.replay.replaydate:"D"$-6_-12#string logfile;.replay.replaydate:"D"$-10#string logfile];
```
robsketch, line 215: This works assuming stp log file names will always have timestamp as suffix, but if there is a better way to do this without hardcoded numbers would be interested to know (tabular and none mode now append timestamp so works for all stp log name cases)

jonnypress: No I think that is fine. We are in control of the log file format, and we append timestamp to end of all, so it's ok to hardcode a number in here. Is 10 correct though? I thought the format was YYYYMMDDHHMMSS ?

robsketch: The 10 is for the old tp log file format, I think it grabs the date from "yyyy.mm.dd"

PR #318, opened 2020-08-18
Add error trapping, setmeta function, adjust namespaces

robsketch:
PR addresses some older comments from main STP branch:

- more error trapping
- new stpmeta function to error trap when `set` is called
- change from root namespace to .stpps and .stplg in `code/processes/segmentedtickerplant.q` to match TorQ code structure

To do:
- [x] change filtered publish from -25! to just use handle
- [x] change end and endp to be one function; function called is determined by if count of the arg is > 1
- [x] define `utabs` in `subscriptions.q` based on `.stpps.subrequestall` instead of `.u.w`
- [x] add tptype flag to subscriptions.q for tp logic instead of using proc`procname

### DIFF
--- a/code/common/subscriptions.q
+++ b/code/common/subscriptions.q
@@ -38,8 +38,8 @@ subscribe:{[tabs;instrs;setschema;replaylog;proc]
 			.lg.o[`subscribe;"autoreconnect was set to true but server functionality is disabled - unable to use autoreconnect"]]];
 	/-check the tables which are available on the server to subscribe to
 	/-if using segmented tickerplant, use .u.w equivalent
-	if[`default~tpmode;utabs:@[proc`w;(key;`.u.w);()]];
-	if[`segmented~tpmode;utabs:@[proc`w;(key;`.stpps.subrequestall);()]];
+	if[`default~.proc.tptype;utabs:@[proc`w;(key;`.u.w);()]];
+	if[`segmented~.proc.tptype;utabs:@[proc`w;`.stpps.t;()]];
 	subtabs:$[tabs~`;utabs;tabs];
 	/-make tabs a list if it isn't already
 	subtabs,:();
@@ -88,7 +88,7 @@ subscribe:{[tabs;instrs;setschema;replaylog;proc]
 			/-the second element of details is a pair (log count;logfile)
 			/-or list of pairs for stp ((count 1;log 1);(count 2;log 2);..)
 			d:$[0=type details[1;0];details 1;enlist details 1];
-			{[d]@[{-11!x;};d;{.lg.e[`subscribe;"could not replay the log file: ", x]}]}each d;
+			{[d]@[{.lg.o[`subscribe;"replaying log file ",.Q.s1 x]; -11!x;};d;{.lg.e[`subscribe;"could not replay the log file: ", x]}]}each d;
 			/-reset the upd function back to original upd
 			@[`.;`upd;:;origupd]];
 		if[replaylog&null raze[details 1]1;
--- a/code/common/subscriptions.q
+++ b/code/common/subscriptions.q
@@ -1,5 +1,9 @@
 /-script to create subscriptions, e.g. to tickerplant
 
+\d .proc
+
+tptype:@[value;`tptype;`segmented];    // [segmented|default] tp type to determine subscription method
+
 \d .sub
 
 AUTORECONNECT:@[value;`AUTORECONNECT;1b];									//resubscribe to processes when they come back up
@@ -33,7 +37,9 @@ subscribe:{[tabs;instrs;setschema;replaylog;proc]
 			[.servers.connectcustom:{x@y;.sub.autoreconnect[y]}[.servers.connectcustom]; .sub.reconnectinit:1b];
 			.lg.o[`subscribe;"autoreconnect was set to true but server functionality is disabled - unable to use autoreconnect"]]];
 	/-check the tables which are available on the server to subscribe to
-	utabs:@[proc`w;(key;`.u.w);()];
+	/-if using segmented tickerplant, use .u.w equivalent
+	if[`default~tpmode;utabs:@[proc`w;(key;`.u.w);()]];
+	if[`segmented~tpmode;utabs:@[proc`w;(key;`.stpps.subrequestall);()]];
 	subtabs:$[tabs~`;utabs;tabs];
 	/-make tabs a list if it isn't already
 	subtabs,:();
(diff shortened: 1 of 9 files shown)


config/settings/segmentedtickerplant.q
```diff
@@ -12,6 +12,7 @@ replayperiod:`day               // [period|day|prior]
 \d .proc
 
 loadprocesscode:1b        
+tptype:`segmented
```
jonnypress, line 15: is this required?

config/settings/tickerplant.q
```diff
@@ -2,6 +2,7 @@
 
 \d .proc
 loadcommoncode:0b		// do not load common code
+tptype:`default
```
jonnypress, line 5: is this required?

code/common/subscriptions.q
```diff
@@ -33,7 +37,9 @@ subscribe:{[tabs;instrs;setschema;replaylog;proc]
 			[.servers.connectcustom:{x@y;.sub.autoreconnect[y]}[.servers.connectcustom]; .sub.reconnectinit:1b];
 			.lg.o[`subscribe;"autoreconnect was set to true but server functionality is disabled - unable to use autoreconnect"]]];
 	/-check the tables which are available on the server to subscribe to
-	utabs:@[proc`w;(key;`.u.w);()];
+	/-if using segmented tickerplant, use .u.w equivalent
+	if[`default~tpmode;utabs:@[proc`w;(key;`.u.w);()]];
```
jonnypress, line 41: where is tpmode defined?

code/segmentedtickerplant/pubsub.q
```diff
@@ -15,14 +15,20 @@ subrequestfiltered:([]tbl:`$();handle:`int$();filts:();columns:())
...
+//  };
 
 // Function to send end of day messages to subscribers      
 // Assumes that .u.end has been defined on the client side   
+// end:{
+//   (neg distinct raze union/[value subrequestall;exec handle from subrequestfiltered])@\:(`.u.end;x);
+//  };
+
+// combine endp and end
+// keep original available until tested
```
jonnypress, line 29: do you need the commented versions above? 
I think for now it would be better to keep the two version of the function above, you could probably just create a allhandles[] functions that returns 
distinct raze union/[value subrequestall;exec handle from subrequestfiltered]
and is then used by both

code/common/subscriptions.q
robsketch: meant to be tptype, it's a typo

config/settings/tickerplant.q
robsketch: There was talk of adding a `tptype` variable (either `default` or `segmented`) for different subscription logic depending on if stp or old tp is being used.  Before logic was using the `proc[proctype]` parameter passed into the subscribe function, but didn't want the logic to depend on the proctype.  I was just unsure of where/how to define the `tptype` variable in the old tp and stp to use in the subscription code, so put it here as a start.

code/common/subscriptions.q
```diff
@@ -1,139 +1,177 @@
...
+ .sub.SUBSCRIPTIONS::0!(4!SUBSCRIPTIONS)upsert enlist proc,`table`instruments`createdtime`active!(tab;instrs;.proc.cp[];1b);
+ }
 
 reconnectinit:0b;		//has the reconnect custom function been initialised
+
+reducesubs:{[tabs;utabs;instrs;proc]
+ // for a given list of subscription tables, remove any which have already been subscribed to
+
+ // if asking for all tables, subscribe to the full list available from the publisher
+ subtabs:$[tabs~`;utabs;tabs],();
```
ThomasSmyth, line 34: Single space indent? :rage:

code/common/subscriptions.q
```diff
@@ -1,139 +1,177 @@
...
+ .lg.o[`subscribe;"replaying the log file(s)"];
+ // store the orig version of upd
+ origupd:@[value;`..upd;{{[x;y]}}];
+ // only use tables user has access to
+ subtabs:realsubs[`subtabs];
+ if[count where nullschema:0=count each schemalist;
+    tabs:(schemalist where not nullschema)[;0];
+    subtabs:tabs inter realsubs[`subtabs]];
+ // set the replayupd function to be upd globally
+ if[not (tabs;realsubs[`instrs])~(`;`);
```
ThomasSmyth, line 73: This should work out the same
```suggestion
 if[not(tabs;realsubs`instrs)~``;
```

code/common/subscriptions.q
```diff
@@ -1,139 +1,177 @@
...
+  if[replaylog;realsubs:replay[tabs;realsubs;details[`schemalist];details[`logfilelist]]];
+  .lg.o[`subscribe;"subscription successful"];
+  updatesubscriptions[proc;;realsubs[`instrs]]each realsubs[`subtabs]];
+
+ // return the names of the tables that have been subscribed for and
+ // the date from the name of the tickerplant log file (assuming the tp log has a name like `: sym2014.01.01
+ // plus .u.i and .u.icounts if existing on TP - details[1;0] is .u.i, details[2] is .u.icounts (or null)
+ logdate:0Nd;
+ if[tptype=`standard;
+   :(`subtables`tplogdate!(details[`schemalist][;0];(first "D" $ -10 sublist string last first details[`logfilelist])^logdate)),{(where 101 = type each x)_x}(`i`icounts`d)!(details[`logfilelist][0;0];details[`rowcounts];details[`date])];
```
ThomasSmyth, line 134: Very long line, could be split in 2:
```suggestion
   d:`subtables`tplogdate!(details[`schemalist][;0];(first"D"$-10 sublist string last first details`logfilelist)^logdate);
   :d,{(where 101=type each x)_x}`i`icounts`d!(details[`logfilelist][0;0];details`rowcounts;details`date)];
```

code/common/subscriptions.q
```diff
@@ -1,139 +1,177 @@
...
+ if[tptype~`segmented;
+   retdic:`logdir`subtables!(details[`logdir];details[`schemalist][;0]);
+   retdic,:{(where 101 = type each x)_x}(`i`icounts`d`tplogdate)!(details[`logfilelist];details[`rowcounts];details[`date];details[`date]);
+   :retdic]
```
ThomasSmyth, line 138: Can extract all require params from `details` in one go and return here instead
```suggestion
   :retdic,{where[101=type each x]_x}`i`icounts`d`tplogdate!details`logfilelist`rowcounts`date`date;
```

PR #321, opened 2020-08-25
Changes and fixes to how DQS handles time variables

tszwailam:
changing variable name gmttime to utctime, how kdb homepage names it
changing variable .eodtime.nextroll, which dictates when .u.end is ran, for whether utctime is 1b/0b

### DIFF
--- a/code/processes/dqe.q
+++ b/code/processes/dqe.q
@@ -69,7 +69,7 @@ loadtimer:{[d]
 /- adds today's date to the time from config csv, before loading the queries to the timer
 configtimer:{[]
   t:.dqe.readdqeconfig[.dqe.configcsv;"S**SN"];
-  t:update starttime:?[((.z.N,.z.n).dqe.utctime)>starttime;01:00+(.z.P,.z.p).dqe.utctime;(`date$(.z.D,.z.d).dqe.utctime)+starttime] from t;
+  t:update starttime:(`date$(.z.D,.z.d).dqe.utctime)+starttime from t;
   {.dqe.loadtimer[x]}each t
   }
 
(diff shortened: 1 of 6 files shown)


ThomasSmyth (approved):
All seems very sensible. Definitely got all references? Anything in FSP to address?

tszwailam:
> All seems very sensible. Definitely got all references? Anything in FSP to address?

There shouldnt be anything in FSP to address. however, we did add another part to saving down for the dqe. Because the dqe is used to query hdbs, lets say for today(08.25) it would be looking at data from (08.24), we thought it would make sense for it to save down T+1 (08.24) instead of 08.25

code/processes/dqe.q
```diff
@@ -67,13 +71,21 @@ loadtimer:{[d]
...
   t:.dqe.readdqeconfig[.dqe.configcsv;"S**SN"];
-  t:update starttime:(`date$(.z.D,.z.d).dqe.gmttime)+starttime from t;
+  t:update starttime:(`date$(.z.D,.z.d).dqe.utctime)+starttime from t;
   {.dqe.loadtimer[x]}each t
   }
 
 writedownengine:{
   if[0=count .dqe.tosavedown`.dqe.resultstab;:()];
+  dbprocs:exec distinct procname from raze .servers.getservers[`proctype;;()!();0b;1b]each`hdb`dqedb`dqcdb;  // Get a list of all databases.
+  restemp1:select from .dqe.resultstab where procs in dbprocs;
```
drgdavies, line 81: Can you explain what the purpose is of splitting these up and then saving separately?

I'm guessing this is to do with getting the right date but how is that being determined from dbprocs?  Needs to be more commenting around this for those coming behind.

tszwailam: When procs in resultstab is a database, that means that data is from yesterday. For example, today's(08.27) would be looking at hdb's data from 08.26. However, when procs arent databases, lets say its looking at symfiles, then it should save down as todays date. Thats why we split it up

drgdavies (approved):

PR #322, opened 2020-08-26
endofday dict parameter

etennoQ:
Added dictionary of process data to arguments of endofday function in subscribers

From standard tickerplant, endofday is run with and empty dictionary
From SegmentedTP, endofday is run with the dictionary full of process data


jonnypress (approved):
looks good

PR #324, opened 2020-09-25
Install script

SanderRoomus:
I would like to see how the github.io page would look like, but can't change the source to be from my branch, to check. 
Or is there other way to see how the new page would be displayed with the changes?

Otherwise the install script will be published. 

Changed the gettingstarted to refer to the new InstallGuide page as well.


installtorqapp.sh
```diff
@@ -0,0 +1,259 @@
+#!/bin/bash
+#Example usage:
+#sh installtorqapp.sh torq=TorQ-3.7.0.tar.gz releasedir=deploy data=datatemp installfile=TorQ-Finance-Starter-Pack-1.9.0.tar.gz
```
jonathonmcmurray, line 3: is there a reason this isn't something like

```suggestion
#sh installtorqapp.sh -torq TorQ-3.7.0.tar.gz -releasedir deploy -data datatemp -installfile TorQ-Finance-Starter-Pack-1.9.0.tar.gz
```

seems like a more common way of passing args to a bash script?

SanderRoomus: From what I found on command line parameters in bash, then the - flags only work properly when they are single characters like -t  TorQ-3.7.0.tar.gz -r deploy and so on. 
Making them long command line parameters I did not find an elegant option that would have the functionality that it has currently. 
Can use getopt, but at the same time having as simple output line with the same functionality I did not find possible. 
Did you have anything else other than getopt in mind?

docs/InstallGuide.md
```diff
@@ -0,0 +1,143 @@
...
+
+`installfile=/home/user/TorQ-FSP/TorQ-Finance-Starter-Pack-master.tar.gz`
+
+</td>
+</tr>
+<tr>
+<td> data </td>
+<td>
+
+An optional parameter. That is if you want to have your database live in a different part of the system rather than the place where the code lives. Can be used as follows:
```
jonnypress, line 89: Change "database" to "data directory as defined by KDBDATA" (or whatever the variable is called)

docs/InstallGuide.md
```diff
@@ -0,0 +1,143 @@
...
+
+`find $1 -type f -name "*.sh" -exec sed -i "s/export KDBBASEPORT=.*/export KDBBASEPORT=7373/g" {} \;`
+
+Create an sh script env_spec_dev.sh and then add the parameter to the install script start line. 
+
+`env=/home/user/env_spec_dev.sh`
+
+This will change the KDBBASEPORT to a new value.
+Similar actions can be done with other variables.
+But requires user basic knowledege of sed commands.
```
jonnypress, line 122: Similar actions can be done with other variables, and required user basic knowledge of sed commands

docs/InstallGuide.md
```diff
@@ -0,0 +1,143 @@
...
+</td>
+
+</tr>
+</table>
+
+## Version control
+
+The installtion script currently works with:
+- TorQ v3.7.0 or higher
+- TorQ-FSP v1.9.0 and higer.
```
jonnypress, line 137: higer = higher

installtorqapp.sh
jonathonmcmurray: I think there are a couple of options - you could take a look at how flags are handled in `torq.sh` (I think it's good if we can match the argument patterns as closely as possible between the two scripts). For some other examples of implementations, this SO answer seems pretty decent: https://stackoverflow.com/a/14203146/8694303

SanderRoomus: Talking with Jonny the behaviour will be as follows:
bash installtorqapp.sh --torq TorQ-3.7.0.tar.gz --releasedir deploy --data datatemp --installfile TorQ-Finance-Starter-Pack-1.9.0.tar.gz

As torq.sh command line parameters that have -csv for example isn't straightforward to implement and to retain the free ordering of variables. 

This behaviour still retains the free ordering of the parameters and is more alike with what torq.sh has and kdb has.

jonnypress (approved):

PR #328, opened 2020-10-05
Documentationfix

SanderRoomus:
Tested these in a private session, and these are currently displayed as follows:
http://aquaqanalytics.github.io/TorQ/InstallGuide/
Just making these changes to correct places and having them in master after testing.

### DIFF
--- a/docs/InstallGuide.md
+++ b/docs/InstallGuide.md
@@ -21,7 +21,7 @@ Full usage of the parameters available in the table below.
 
 The folder structure after installation will look like this:
 
-<center><img src="graphics/Installscript_folder_structure.png" width="600"></center>
+<center><img src="../graphics/Installscript_folder_structure.png" width="600"></center>
 
 Then to run the TorQ stack:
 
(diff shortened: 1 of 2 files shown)


docs/InstallGuide.md
```diff
@@ -34,99 +34,44 @@ Check if the stack is up
...
-Similar actions can be done with other variables, and required user basic knowledge of sed commands.
-The script will scan through the code in the TorQApp directory and the 
-bin directory from the deploy folder. 
-If DEV and UAT run on different data sources then using env variable the install script can replace them with the correct server address.
-This is essentially the environment-specific config file.
-
-</td>
-
-</tr>
+<style type="text/css">
```
jonathonmcmurray, line 37: do we need to use all this HTML stuff? is there no way to do this with markdown? (I think markdown is generally a bit easier to read without rendering etc., i.e. if people were trying to look at the raw markdown files)

SanderRoomus: Jonny requested the parameters to be in table, and markdown what I found after my research then the formatting of a table is quite primitive. 
So the right hand side would look like a mess to read on the github page. 
Like so:
image

jonathonmcmurray (approved):

PR #336, opened 2020-10-15
move code from code/dataaccess to code/common/dataaccess

mmeenagh1:
moved code from code/dataaccess to code/common/dataaccess + put an init function in code/common/dataaccess.q


jonnypress:
I think it's ok for now at least (maybe better) to have all the dataaccess code in code/dataaccess
It can be loaded into a process by adding something like 

-loaddir /path/to/code/dataaccess

mmeenagh1:
> I think it's ok for now at least (maybe better) to have all the dataaccess code in code/dataaccess
> It can be loaded into a process by adding something like
> 
> -loaddir /path/to/code/dataaccess

Hi Jonny - I just moved it back to code/dataaccess

It's set up to load the scripts in code/dataaccess by calling `.proc.loaddir in the init function (see code below) - I thought if I used the "-loaddir" command line parameter users might call the init function, and the code wouldn't be available

`init:{
  .lg.o[`.dataaccess.init;"running .dataaccess.init[]"];
  .dataaccess.tablepropertiespath:first .proc.getconfigfile["tableproperties.csv"];         //- config defining any non standard attribute/primary time columns
  .dataaccess.checkinputspath:first .proc.getconfigfile["checkinputs.csv"];                 //- The name of the input csv to drive what gets done
  additionalscripts:getenv[`KDBCODE],"/common/dataaccess";                                  //- load all q scripts in this path
  .proc.loaddir additionalscripts;
  tablepropertiesconfig:readtableproperties[tablepropertiespath;"sssss"];
  checkinputsconfig:readcheckinputs[checkinputspath;"sbs*"];
  .lg.o[`.dataaccess.init;"running .dataaccess.init[] - finished"];
 };`

PR #337, opened 2020-10-16
Change in to within

mg-aquaq:
1.	the date clause here is generated the wrong way round so we end up with "date in (T - 1; T - 1 - N)" and it returns nothing
2.	we're using "in" when I think we should use "within"?

### DIFF
--- a/code/dqc/pctAvgDailyChange.q
+++ b/code/dqc/pctAvgDailyChange.q
@@ -10,7 +10,7 @@
 pctAvgDailyChange:{[fname;tabname;rt;ndays;thres]                                       
   .lg.o[`pctAvgDailyChange;"Checking daily average change"];  
   if[ndays>-1+count .Q.pv; :(0b;"error: number of days exceeds number of available dates")];
-  previous:select avg resvalue from rt where date within((-1*ndays)+last date;-1+last date),funct=fname,table=tabname;
+  previous:select avg resvalue from rt where date within(-1*ndays;-1)+last date,funct=fname,table=tabname;
   current:select avg resvalue from rt where date=last date,funct=fname,table=tabname;
   c:abs[current[0;`resvalue]- previous[0;`resvalue]]<=thres*previous[0;`resvalue];
   (c;"count ",$[c;"doesn't differ";"differs"]," from ",(string ndays)," days average by more than ",(string thres),"%")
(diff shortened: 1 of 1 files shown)


code/dqc/pctAvgDailyChange.q
```diff
@@ -10,7 +10,7 @@
 pctAvgDailyChange:{[fname;tabname;rt;ndays;thres]                                       
   .lg.o[`pctAvgDailyChange;"Checking daily average change"];  
   if[ndays>-1+count .Q.pv; :(0b;"error: number of days exceeds number of available dates")];
-  previous:select avg resvalue from rt where date in (-1+last date;(-1*ndays)+last date),funct=fname,table=tabname;
+  previous:select avg resvalue from rt where date within((-1*ndays)+last date;-1+last date),funct=fname,table=tabname;
```
jonathonmcmurray, line 13: think this would work too:

```suggestion
  previous:select avg resvalue from rt where date within(-1*ndays;-1)+last date,funct=fname,table=tabname;
```

jonathonmcmurray (approved):

ThomasSmyth:
Order shouldn't matter for `in`:
```
q)til[3]in 0 1
110b
q)til[3]in 1 0
110b
```
`in` will return data for 2 specific dates whereas `within` will use the range, if n>1 then this would be more than 2 days and I think this is only supposed to compare 2 dates. Let me know if my assumptions are wrong.

mg-aquaq:
> 
> 
> Order shouldn't matter for `in`:
> 
> ```
> q)til[3]in 0 1
> 110b
> q)til[3]in 1 0
> 110b
> ```
> 
> `in` will return data for 2 specific dates whereas `within` will use the range, if n>1 then this would be more than 2 days and I think this is only supposed to compare 2 dates. Let me know if my assumptions are wrong.

I think we are wanting to use `within`, as we are looking at the previous n days and calculating an average. This is the comment in the function

```
/- Check that current result of a given function applied to a given table is
/- within threshold limits of n days average taken from results table.
/- Parameters: fname - name of function from dqe engine; tabname - name of
/- table; rt - results table in dqedb; ndays - number of previous days to
/- compute daily average; thres - threshold is a number from 0 to 1 that
/- corresponds to a range from 0% to 100%
```

ThomasSmyth:
Cheers, apt fix then :+1:

ThomasSmyth (approved):
@jonathonmcmurray commented has a more terseversion if you wish to add that.

PR #338, opened 2020-10-20
Chained stp

etennoQ:
Code base changes for SCTP functionality

### DIFF
--- a/code/processes/segmentedtickerplant.q
+++ b/code/processes/segmentedtickerplant.q
@@ -7,6 +7,7 @@
 // Filtered - apply filters to published data, filters defined on client side
 
 createlogs:@[value;`createlogs;1b]; // allow tickerplant to create a log file
+.sctp.tph:@[value;`.sctp.tph;-1];   // placeholder tp handle value in stp process
 
 // subscribers use this to determine what type of process they are talking to
 tptype:`segmented
@@ -15,7 +16,7 @@ tptype:`segmented
 .proc.loadf[getenv[`KDBCODE],"/common/timezone.q"];
 .proc.loadf[getenv[`KDBCODE],"/common/eodtime.q"];
 
-$[not null .sctp.sctploggingmode;[
+$[.sctp.chainedtp;[
   .proc.loadf[getenv[`KDBCODE],"/common/timer.q"];
   .proc.loadf[getenv[`KDBCODE],"/common/subscriptions.q"];
   schemafile:""
@@ -59,7 +60,7 @@ generateschemas:{
   // updtab stores functions to add/modify columns
   // Default functions timestamp updates
   // Preserve any prior definitions, but default all tables if not specified
-  $[not null .sctp.sctploggingmode;
+  $[.sctp.chainedtp;
     .stplg.updtab:(.stpps.t!(count .stpps.t)#{[x;y] x}),.stplg.updtab;
     .stplg.updtab:(.stpps.t!(count .stpps.t)#{(enlist(count first x)#y),x}),.stplg.updtab
     ]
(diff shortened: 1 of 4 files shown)


code/segmentedtickerplant/pubsub.q
```diff
@@ -91,7 +91,10 @@ closesub:{[h]
   delhandlef[;h]each t;
  };
 
-.z.pc:{[f;x] @[f;x;()]; closesub x}@[value;`.z.pc;{{}}]
+.z.pc:{[f;x]
```
ThomasSmyth, line 94: Can you add a comment giving an outline of what this function does?

code/processes/segmentedtickerplant.q
```diff
@@ -136,4 +136,5 @@ generateschemas[];
 
 // Create log directory, open all table logs
 // use name of schema to create directory
-.stplg.init[dbname::(string .proc.procname),"_"]
\ No newline at end of file
+`dbname set (string .proc.procname);
```
jonnypress, line 139: do we need a variable called dbname if it is always just going to be the string of .proc.procname?

code/segmentedtickerplant/stplog.q
```diff
@@ -215,14 +228,21 @@ $[.sctp.chainedtp;
...
     .lg.o[`endofperiod;"flushing remaining data to subscribers and clearing tables"];
     .stpps.pubclear[.stplg.t];
-    p:.z.p+.eodtime.dailyadj;
+    p:.z.p;
     .lg.o[`endofday;"executing end of day for ",.Q.s1 .eodtime.d];
-    .stpps.end[x;y];
+    .stpps.end[date;data];
     if[p>.eodtime.nextroll:.eodtime.getroll[p];system"t 0";'"next roll is in the past"];
-    if[value `..createlogs; closelog each logtabs]
+    if[value `..createlogs;
```
jonnypress, line 238: same comment applies here as above - needs some comments + can we create a function out of it if it used more than once 
(I don't know if it is used more than once)

code/segmentedtickerplant/stplog.q
```diff
@@ -190,55 +198,61 @@ endofdaydata:{
 
 // Send close of period message to subscribers, update logging period times
 // roll logs if flag is specified - we don't want to roll logs if end-of-day is also going to be triggered
-$[.sctp.chainedtp;
-  endofperiod:{[x;y;z]
+$[not null .sctp.sctploggingmode;
```
jonnypress, line 201: For end of day / end of period... instead of creating the definition like this, can you just use if statements within the function definition, rather than a switch where we declare the function. 
The way you have it now is what we want to do for the functions that are called often (e.g. every update) but it makes the code more confusing. So for the functions that are called rarely (endofday/endofperiod) can you just put if statements  within the function definition, rather than having two separate definitions

code/segmentedtickerplant/stplog.q
```diff
@@ -260,8 +274,8 @@ init:{[dbname]
   getnextendUTC[]; 
   i::1; // default value for log seq number
 
-  if[value `..createlogs;
-    createdld[.proc.procname;.eodtime.d];
+  if[(value `..createlogs) or .sctp.sctploggingmode=`create;
+    createdld[dbname;.eodtime.d];
```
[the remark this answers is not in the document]
etennoQ: dbname is the name of the function parameter here so whatever the input is will be the name. I was avoiding hardcoding it because the tests create unique test logs with different names
in a normal initialisation the procname is the function input

jonnypress (approved):
I think that looks good to me

PR #346, opened 2020-11-05
Permissions fixes

jonathonmcmurray:
Number of items addressed in this PR:

* Creating a user & group with same name caused an infinite loop, now rejects a name if already used for user/group
* On newer versions of kdb+, we can inject permissions into HTTP (GET) requests without modifying `.z.ph` by setting `.h.val` - previously HTTP requests just weren't supported in TorQ permissions
* Parsing to check permissions was broken for some nested expressions containing lambdas (e.g. the requests sent by TorQ monitor to other processes)
* When checking permissions on expressions containing lambdas, was only checked for references to variables that were already permissioned for at least some users i.e. was essentially "allowlisting" all variables by default, even in non-permissive mode. Now checks against all defined variables in all namespaces.
* Allow `systemuser` access to `checkresulthandler`, necessary for monitor process to work
* Allow lambda expressions of the form `(lambda;dict)`
* Fix conditional in `.pm.exe` so `eval` output is returned were applicable (e.g. `"count x"` - before would always return 1, regardless of what `x` contained)
* Skip empty lines from qcon (sends `""` rather than `"\n"` at q console)

Above changes are also unit tested, with exception of HTTP requests; couldn't find a good way to test this without spawning additional processes etc., as calling `.z.ph` directly means that `.z.w=0` and permissions checks are bypassed. Tested manually and appears to work fine (tom permissioned for `f:{x+y}`, bob not):

```
(base) jmcmurray@homer ~/git/TorQ (pm_fixes) $ curl -w '\n' 'http://tom:pass@127.0.0.1:15000/?f\[1;2\]'
<html><head><style>a{text-decoration:none}a:link{color:024C7E}a:visited{color:024C7E}a:active{color:958600}body{font:10pt verdana;text-align:justify}</style></head><body><pre>3
</pre></body></html>
(base) jmcmurray@homer ~/git/TorQ (pm_fixes) $ curl -w '\n' 'http://bob:pass@127.0.0.1:15000/?f\[1;2\]'
'pm: user role does not permit running function [f]
(base) jmcmurray@homer ~/git/TorQ (pm_fixes) $
```

### DIFF
--- a/code/handlers/permissions.q
+++ b/code/handlers/permissions.q
@@ -39,13 +39,19 @@ virtualtable:([name:`symbol$()]table:`symbol$();whereclause:())
 publictrack:([name:`symbol$()] handle:`int$())
 
 / api
-adduser:{[u;a;h;p]user,:(u;a;h;p)}
+adduser:{[u;a;h;p]
+  if[u in key groupinfo;'"pm: cannot add user with same name as existing group"];
+  user,:(u;a;h;p)}
 removeuser:{[u]user::.[user;();_;u]}
-addgroup:{[n;d]groupinfo,:(n;d)}
+addgroup:{[n;d]
+  if[n in key user;'"pm: cannot add group with same name as existing user"];
+  groupinfo,:(n;d)}
 removegroup:{[n]groupinfo::.[groupinfo;();_;n]}
 addrole:{[n;d]roleinfo,:(n;d)}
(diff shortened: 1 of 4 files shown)


toosuto-r (approved):
Looks good, nice fix for the overly-permissive lambdas

jonnypress (approved):

jonnypress:
looks good

toosuto-r (approved):
Nice, approved

PR #347, opened 2020-11-17
Testing framework code

mpotterq:
Updated unit/integration testing framework code. Features include:
 - More comprehensive error/failure handling in tests
 - Pass in flags to testing script to debug, write to disk, version and more
 - Run all tests in given subfolders in one script

See docs for more info


tests/k4unit.q
```diff
@@ -61,50 +62,82 @@ KUltf:{ / (load test file) - load tests in csv file <x> into KUT
 	/KUT,:update file:x,action:lower action,lang:`q^lower lang,ms:0^ms,bytes:0j,repeat:1|repeat from `action`ms`lang`code`repeat`comment xcol("SISSI*";enlist .KU.DELIM)0:x:hsym x;
 	neg before-count KUT}
 
-KUltd:{ / (load test dir) - load all *.csv files in directory <x> into KUT
+KUltd:{ / (load test dir) - load all *.csv files in directory <x> into KUT except process.csv
 	before:count KUT;
-	files:(` sv)each(x,'key x);KUltf each files where(lower files)like"*.csv";
+	files:(` sv)each(x,'key x);KUltf each files where 1=sum (lower files) like/: ("*process.csv";"*.csv");
```
jonathonmcmurray, line 67: I think it might be more clear/readable to be a little more verbose here:

```suggestion
	files:(` sv)each(x,'key x);KUltf each files where (lower[files]like"*.csv")&not lower[files]like"*process.csv";
```

tests/k4unit.q
```diff
@@ -61,50 +62,82 @@ KUltf:{ / (load test file) - load tests in csv file <x> into KUT
...
+	if[1<.KU.VERBOSE;.lg.o[`k4unit;s]];
+	$[.KU.DEBUG & allowfail;value s;@[value;s;{(`err;`$x;y)}[;code]]]
+	}
 
+// If in error - it now returns the error as well as the offending code - basically the same function as KUpexec?
 KUexec:{[lang;code;repeat]
-	value(string lang),")",$[1=repeat;string code;"do[",(string repeat),";",(string code),"]"]}
+	strexec:(string lang),")",$[1=repeat;string code;"do[",(string repeat),";",(string code),"]"];
+	@[value;strexec;{(`err;`$x;y)}[;code]]
+	}
```
jonathonmcmurray, line 99: Could we just make `KUexec` a projection of `KUpexec`?

```
KUexec:KUpexec["";;;1b]
```

or something like that?

tests/k4unit.q
```diff
@@ -61,50 +62,82 @@ KUltf:{ / (load test file) - load tests in csv file <x> into KUT
...
+
+// Generate error logs from beforeeach, before, after and aftereach tests
+KUerrparse:{[action;out]
+	vals:1_' out where `err ~/: first each out:raze each flip value out;
+	.lg.e[`KUexecerr;] each KUerrparseinner[action;;;;] .' vals;
```
jonathonmcmurray, line 104: given that k4unit.q is automatically loaded & tests run as torq.q loads (when `-test` arg is present), I think `.lg.e` would cause torq to exit if not run in debug mode? Is that the expected behaviour?

tests/runall.q
```diff
@@ -0,0 +1,24 @@
+// Define run-all function
+.k4.runall:{[run;res]
+  // Find all the run scripts, generate strings and execute
+  rundirs:dirs where `run.sh in' key each dirs:.Q.dd[run;] each key run;
+  runtime:string .z.p;
+  command:{1_string[x],"/run.sh -r ",y}[;runtime] each rundirs;
+  {show "Executing ",x;system x} each command ,\: " -wq";
```
jonathonmcmurray, line 7: why not include the `-wq` in command?

```suggestion
  command:{1_string[x],"/run.sh -r ",y," -wq"}[;runtime] each rundirs;
  {show "Executing ",x;system x} each command;
```

tests/runall.q
```diff
@@ -0,0 +1,24 @@
...
+  // Get errors from most recent log files and return output list
+  reclogs:.Q.dd[logdir;] each l where not (l:key logdir:.Q.dd[resdir;`logs]) like "*",ssr[string .z.d;".";"_"],"*";
+  logerr:err!read0 each err:reclogs where reclogs like "*err*";
+  (results;errors;logerr)
+  };
+
+// Only execute function if the necessary flags are passed in
+if[all `rundir`resdir in key args:.Q.opt .z.x;
+  output:.k4.runall . hsym each `$first each args[`rundir`resdir];
+  show each ("Test results:";results:output 0;"Test failures:";fails:output 1;"Logged errors:";errors:output 2)
```
jonathonmcmurray, line 23: could avoid the intermediate `output` variable with `set'` seeing as these are global anyway

```suggestion
  `results`fails`errors set' .k4.runall . hsym each `$first each args[`rundir`resdir];
  show each ("Test results:";results;"Test failures:";fails;"Logged errors:";errors)
```

tests/runtests.q
```diff
@@ -1,8 +1,42 @@
...
+  if[not -11h=type key hsym `$ef:respath,"/",errfile;system "touch ",ef;.lg.o[testname;"Creating file ",ef]];
+
+  // Timestamp tests
+  res:`runtime xcols update runtime:first rtime from delete timestamp from res;
+  err:`runtime xcols update runtime:first rtime from delete timestamp from err;
+
+  // Write test results and errors to these files
+  .lg.o[testname;"Writing ",string[count KUTR]," results rows and ",string[count KUerr]," error rows"];
+  hclose abs neg[hopen hsym `$rf] $[hcount hsym `$rf;1;0]_csv 0: res;
+  hclose abs neg[hopen hsym `$ef] $[hcount hsym `$ef;1;0]_csv 0: err;
```
jonathonmcmurray, line 24:
```suggestion
  hclose abs neg[hopen rf] $[hcount rf:hsym `$rf;1;0]_csv 0: res;
  hclose abs neg[hopen ef] $[hcount ef:hsym `$ef;1;0]_csv 0: err;
```

think these lines could do with a little explanation, there's a number of things happening & it's not all immediately clear why IMO

jonathonmcmurray (commented):
There are quite a lot of system commands in here - is TorQ unit testing now going to be Linux only? If possible, I'd say we should use `.os` functions from TorQ's `os.q` to maintain cross-platform support (might not cover all system commands being used though)

tests/k4unit.q
jonathonmcmurray: Think about this further... it might be better to modify `KUltf` to skip any CSVs that aren't in k4unit format, rather than hardcoding a single name to skip? This would allow e.g. defining some mock data in a CSV to be loaded for tests etc., which atm I think would probably cause a problem when `KUltf` tries to load differently formatted files?

jonathonmcmurray:
While we're making changes to testing... I don't think we have any mention of testing in the TorQ documentation. Would be good to have something, covering how to run tests, differences to "official" k4unit & writing tests in the format expected by this extended framework

tests/runtests.q
```diff
@@ -1,8 +1,42 @@
...
 KUrt[];
+show each ("k4unit Test Results";KUTR;"k4unit Test Errors";KUerr);
 
-show "k4unit Test Results"
-show KUTR
-show "k4unit Test Errors"
-show KUerr
+// If enabled write results to disk
+if[all `write`testresults in key .Q.opt .z.x;.[.k4.writeres;(KUTR;KUerr),clargs;{.lg.e[`test;"Error: ",x]}]];
+if[not `debug in key .Q.opt .z.x;exit 0];
```
jonathonmcmurray, line 42: for any automated testing/CI type stuff, I think these will generally see a 0 exit code as "success" and anything >0 as "fail", so it might be nice here to make it do `exit 0` only if there are no failed tests, and `exit 1` or something if there are any failures?

mpotterq:
> While we're making changes to testing... I don't think we have any mention of testing in the TorQ documentation. Would be good to have something, covering how to run tests, differences to "official" k4unit & writing tests in the format expected by this extended framework

Yes good point, the docs ended up in another PR, will get them brought over

tests/k4unit.q
mpotterq: When not run in debug mode any errors in the test code itself is logged to an err_proc.log file. The .lg.e seems to work fine

tests/k4unit.q
mpotterq: Yeah I've thought a little about this, as under this testing framework you could have various CSVs in the same folder. I'll give this a go

tests/k4unit.q
jonathonmcmurray: Ah yep, looking at it now I see `.proc.initialised:1b` is set before loading & running tests, so `.lg.e` won't exit :+1:

docs/unittesting.md
```diff
@@ -0,0 +1,264 @@
...
+This will automatically load in the test code and all CSVs in that directory into a test TorQ process and run the tests in those files, then output the results to the screen.
+
+**Simple RDB Test**
+
+A similar scenario is when the test code is loaded into a 'production' TorQ process, ie. an RDB or a tickerplant, in order to test the functioning of that process. A simple RDB example can be seen here:
+
+```
+action,ms,bytes,lang,code,repeat,minver,comment
+true,0,0,q,`segmentedtickerplant~.rdb.tickerplanttypes,1,,"Check the TP type"
+true,0,0,q,all (null;{-11h=type x}) @' .rdb.subscribeto,1,,"Check RDB is subscribed to null symbol (all tables)"
```
jonathonmcmurray, line 37: think this is the same? unless there's a reason to make this more complex

```suggestion
true,0,0,q,`~.rdb.subscribeto,1,,"Check RDB is subscribed to null symbol (all tables)"
```

FWIW, I'm not sure that these are great example tests, as it's just checking the value of some configurable variables, rather than actually testing any functionality? I suppose the idea is to keep it simple for example, but IMO it'd be better to show testing actual funcationality

docs/unittesting.md
```diff
@@ -0,0 +1,264 @@
...
+true,0,0,q,1~count key .rdb.subcsv,1,,"Check that the subcsv exists"
+```
+
+These tests simply check that certain process variables have been set properly, and these are triggered similarly to the previous set, except that an RDB process is used to load in the tests rather than a 'blank' test process:
```
jonathonmcmurray, line 41: It might be worth mentioning here that by using the appropriate `proctype`, relevant config will be loaded in, as well as `${KDBCODE}/code/{procname}/` folder depending on config (`.proc.loadprocesscode` is the relevant config variable for that, I think - we have it turned on for most processes in TorQ by default I think).

This allows what I think is a fairly clean, test-friendly pattern for developing a TorQ process - placing majority of logic in q script(s) in `code/{procname}/` & then having a "main" `code/processes/{procname}.q` script that simply initialises the process (connects to discovery & any other procs required, does any subscriptions/recovery, sets timers etc.). That way, using a line like the one below, all the logic can be loaded without initialising the process, and functions can be easily tested in a standalone fashion.

`procname` generally isn't as important (can have procname specific config & code, but typically it's proctype), and can be set to something other than e.g. `rdb1` to allow tests to run without conflicting with an already running stack.

docs/unittesting.md
```diff
@@ -0,0 +1,264 @@
...
+```
+
+We can get the code which failed and run it here to see what it returns. From doing some quick debugging we can see that one of the items being added to `t1` is wrong, it should be ten rather than 1. Once this is fixed we can run the WDB tests again and we see that there are now no errors and all the tests pass! We can then run all of our tests again as at the start and no new test failures come up and the latest error logs are empty.
+
+### Notes on Best Practice
+
+Here are some recommendations on making test development more straightforward:
+
+- Use TorQ connection management when dealing with multiple processes
+- Run the tests from a 'blank' test process where possible, as this ensures that test code and process code don't get in each other's way
```
jonathonmcmurray, line 261: I think I disagree with this point, related to my comment above; in general, the whole point is to test "process code", so loading a blank test process complicates things. Typically better IMO to load a test process with relevant `-proctype` so process config & code can be loaded to be tested

mpotterq: I see what you mean, I guess I'm thinking about this mainly from an integration testing perspective where you'll have various processes interacting, whereas from a unit testing perspective your approach makes more sense. I'll try and make the distinction clearer.

docs/unittesting.md
mpotterq: I've added a note on this stuff, but are unit testing docs the best place to go into detail on it?

jonathonmcmurray: What you have added looks good, other than this bit:

> which loads in the code in `${KDBCODE}/processes/rdb.q`

that's not true, for regular TorQ rdb that file is loaded by `-load` flag, with just the `-procname` it _won't_ be loaded (which is good - for unit testing, don't want to initialise the process fully & make connections to other TorQ processes etc.). It'll just be the relevant config & (usually) code from `${KDBCODE}/rdb` ("usually" because loading that code is config-dependent, but usually turned on for TorQ processes)

jonnypress:
This looks good, when @jonathonmcmurray is happy please merge

jonathonmcmurray (approved):
All looks good.

I do have a small suggestion in regards to simplifying debugging; atm, we can set `-.KU.DEBUG 1` to enable k4unit debug mode (which bypasses error trap when running tests, with the idea being you get into debug mode at point of error & can debug from there).

However, `tests/runtests.q` is loaded by `.proc.loadf` in `torq.q`, which error traps the loading of the script, so even with TorQ `-debug -stop`, we get into debug mode, but within `.proc.loadf` rather than at the point the error was actually thrown - have to manually reload the `tests/runtests.q` file & let tests fail again to be in the right place to properly debug.

All of this is a long-winded way to say... I think we should

* automatically set `.KU.DEBUG` based on TorQ flags (probably if we have `-stop`, set `.KU.DEBUG` to true)
* load `tests/runtests.q` without error trap, so that it fails in the "right" place; when `.KU.DEBUG` is false, tests will be error trapped anyway, so this shouldn't cause a problem

Just a thought, otherwise, I'm happy with this :+1:

PR #348, opened 2020-11-23
Add cheat sheet

jonnypress:
Adding in cheat sheet to help with the main points that are useful to look at / need enforced.

### DIFF
--- a/docs/cheatsheet.md
+++ b/docs/cheatsheet.md
@@ -1,0 +1,198 @@
+#Cheat Sheet
+
+The purpose of this cheatsheet is to provide a very condensed guide to the bits you need to know about TorQ to either debug a running process, or to extend a TorQ stack. It does not replace the main body of documentation, and ordering is presumed order-of-relevance. The below uses the default configuration of TorQ and TorQ Finance Starter Pack, though a lot of it is configurable. 
+
+##Debugging Running Processes
+
+Each TorQ process has several internal tables which are useful. Connect to these using standard tools (qcon, an IDE etc.). Default user:pass of admin:admin will usually work. 
+
+###.usage.usage
+.usage.usage is used to track the queries that are executed against the process and is usually the first place to look for problems. Data is stored in memory for 24 hours and also persisted to disk in a usage_* file which is rolled daily. Every query and timer function call is logged, except for `.u.upd or `upd messages as this would bloat the log file considerably. Queries are logged before they are executed (status="b") and after (status="c" (complete) or status="e" (error)). 
+
+If a query blocks a process and makes it unresponsive, it will have an entry (status="b") in the log file on disk.  
+
+.usage.usage can be queried like any other kdb+ table to diagnose problems e.g. 
+
+```
+100 sublist `timer xdesc .usage.usage
+select from .usage.usage where time within ... 
+```
+
+Note that this table is not especially helpful for gateway queries which are executed in an async call back manner. The gateway part of the the request will (should) usually have a very short run time so the back end services should be interrogated to see what slow parts are. 
+
+[More info.](http://aquaqanalytics.github.io/TorQ/handlers/#logusageq)
+
+##.timer.timer
+.timer.timer shows information about the timer calls which are scheduled / have been run. Pay attention to the "active" field- if a timer call fails it will be removed from the timer (active=0b). To avoid this if required, wrap the function being executed by the timer in an error trap in the standard way. Use .timer.timer in combination with .usage.usage to work out if there are slow running/too frequent timers which are causing problems. 
+
+[More info.](http://aquaqanalytics.github.io/TorQ/utilities/#timerq)
+
+###Log Files 
+Log files are stored in the log directory specified by KDBLOG. Each process creates 3 log files: 
+ - an out log (out_*) with standard log messages
+ - an error log file (err_*) with errors
+ - a usage log file (usage_*) with a log of every request that hits the process. 
+
+The error log file should be empty. Don't ignore the out_ log file, there is a lot of information in there which can be used to debug. One thing that is a bit awkward is that if there is an error then the error log message timestamp has to be matched off against the out message log messages. You can force a process to write to a single log file if the process is started with the -onelog command line parameter. 
+
+###.clients.clients
+This shows inbound connections (connections created into this process). It may have interesting information about connection open/close. If it has a lot of rows it means some clients are connecting and disconnecting frequently. [More info.](http://aquaqanalytics.github.io/TorQ/handlers/#trackclientsq)
+
+###.servers.SERVERS
+This shows outbound connections (connections created by this process to other processes). It's useful for tracking connections which have died. 
+
+##Starting/Stopping and Debugging Processes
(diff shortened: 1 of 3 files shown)


docs/cheatsheet.md
```diff
@@ -0,0 +1,168 @@
...
+  summary                                  to view summary table
+  top <processname>                        to show top.q statistics for a single process
+Optional flags:
+  -csv <fullcsvpath>                       to run a different csv file
+  -extras <args>                           to add/overwrite extras to the start line
+  -csv <fullcsvpath> -extras <args>        to run both
+  -force                                   to force stop process(es) using kill -9
+```
+
+It is very, very difficult to try to debug a running process remotely. Do not do this. Use torq.sh to stop the process and run it in the foreground in a test environment.
```
ThomasSmyth, line 60: Could do with being in bold.

docs/cheatsheet.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Adding Custom Processes
+
+When adding custom processes to TorQ it is important to [understand the significance of proctype](http://aquaqanalytics.github.io/TorQ/gettingstarted/#process-identification). In a nutshell, only processes which do exactly the same thing should have the same proctype. If two processes do roughly the same thing then -parentproctype can be used to share common functionality. Also there isn't any formal association between proctype and code name file name e.g. a process that loads code/processes/rdb.q can have any proctype we like. 
+
+How TorQ [manages connections](http://aquaqanalytics.github.io/TorQ/conn/) is important. Avoid using hopen, use TorQ connection management.
+
+TorQ uses the fail fast principle (if you are going to fail, may as well do it as quickly as possible). This helps avoid processes starting up in inconsistent or unexpected states. If running a process with the -debug option, add the -stop or -trap options to stop at, or trap and continue through, start up errors. 
+
+Code is loaded in a specific order, which can be overridden. To determine the order, inspect the bottom of the torq.q script (the last 50 lines or so).
```
ThomasSmyth, line 168: This is a little bit like throwing a dart at a map, perhaps there is some specific function(s) that could be referenced to help find this section of `torq.q`? Obviously line numbers could change over time and that won't work.

ThomasSmyth (approved):
Good stuff. Going forward it might be worth having anyone training with TorQ to bear this in mind and suggest additions that might not be obvious to anyone familiar with TorQ.

docs/cheatsheet.md
```diff
@@ -0,0 +1,168 @@
+# Cheat Sheet
+
+The purpose of this cheatsheet is to provide a very condensed guide to the bits you need to know about TorQ to either debug a running process, or to extend a TorQ stack. It does not replace the main body of documentation, and ordering is presumed order-of-relevance. The below uses the default configuration of TorQ and TorQ Finance Starter Pack, though a lot of it is configurable. 
+## Debugging Running Processes
+
+Each TorQ process has several internal tables which are useful. Connect to these using standard tools (qcon, an IDE etc.). Default user:pass of admin:admin will usually work. 
+
+### .usage.usage
+.usage.usage is used to track the queries that are executed against the process and is usually the first place to look for problems. Data is stored in memory for 24 hours and also persisted to disk in a usage_* file which is rolled daily. Every query and timer function call is logged, except for `.u.upd or `upd messages as this would bloat the log file considerably. Queries are logged before they are executed (status="b") and after (status="c" (complete) or status="e" (error)).
```
jonathonmcmurray, line 9:
```suggestion
.usage.usage is used to track the queries that are executed against the process and is usually the first place to look for problems. Data is stored in memory for 24 hours and also persisted to disk in a usage_* file which is rolled daily. Every query and timer function call is logged, except for `` `.u.upd`` or `` `upd`` messages as this would bloat the log file considerably. Queries are logged before they are executed (status="b") and after (status="c" (complete) or status="e" (error)). 
```

think that'll properly format the function names to avoid losing the backtick

docs/cheatsheet.md
```diff
@@ -0,0 +1,168 @@
...
+If a query blocks a process and makes it unresponsive, it will have an entry (status="b") in the log file on disk.  
+
+.usage.usage can be queried like any other kdb+ table to diagnose problems e.g. 
+
+```
+100 sublist `timer xdesc .usage.usage
+select from .usage.usage where time within ... 
+```
+
+Note that this table is not especially helpful for gateway queries which are executed in an async call back manner. The gateway part of the the request will (should) usually have a very short run time so the back end services should be interrogated to see what slow parts are. [More info.](http://aquaqanalytics.github.io/TorQ/handlers/#logusageq)
```
jonathonmcmurray, line 20:
```suggestion
Note that this table is not especially helpful for gateway queries which are executed in an async call back manner. The gateway part of the request will (should) usually have a very short run time so the back end services should be interrogated to see what slow parts are. [More info.](http://aquaqanalytics.github.io/TorQ/handlers/#logusageq)
```

double word

docs/cheatsheet.md
```diff
@@ -0,0 +1,168 @@
...
+wget https://raw.githubusercontent.com/AquaQAnalytics/TorQ-Finance-Starter-Pack/master/installlatest.sh
+// execute it
+bash installlatest.sh 
+// It will finish with a message like below:
+=============================================================
+INSTALLATION COMPLETE
+=============================================================
+Installation is finished. For a regular installation, run it as follows in the working directory: ./deploy/bin/torq.sh start all
+// you may need to change the value of KDBBASEPORT held in deploy/bin/setenv.sh to avoid conflicts with other TorQ stacks if running on a shared host
+// once done execute start line and check it:
```
jonathonmcmurray, line 131:
```suggestion
```bash
# pull in latest TorQ FSP install script 
wget https://raw.githubusercontent.com/AquaQAnalytics/TorQ-Finance-Starter-Pack/master/installlatest.sh
# execute it
bash installlatest.sh 
# It will finish with a message like below:
=============================================================
INSTALLATION COMPLETE
=============================================================
Installation is finished. For a regular installation, run it as follows in the working directory: ./deploy/bin/torq.sh start all
# you may need to change the value of KDBBASEPORT held in deploy/bin/setenv.sh to avoid conflicts with other TorQ stacks if running on a shared host
# once done execute start line and check it:
```

think it makes sense to mark this as a bash block & use `#` for comments. also prompt is shown in lines below, but not these lines?

docs/cheatsheet.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+## Modifying Existing Installations
+We always try to esnure new versions of TorQ are backwardly compatible. Try to avoid modifying TorQ itself and instead make application specific modifications. The installation script deploys TorQ in a structure which keeps TorQ and the application separate, and this should be adhered to whenever possible. 
+
+To modify the config for a process, do it either in the [config directory](http://aquaqanalytics.github.io/TorQ/gettingstarted/#configuration-loading) or as a command line start up parameter as- almost all config variables that are defined in a config file and exists in a non-root namespace can be overridden from the command line. 
+
+To make a process load additional files, you can:
+
+- append the additional files to the -load parameter on the start line
```
jonathonmcmurray, line 154: might be worth mentioning this can be done in `process.csv` & then `torq.sh` will pick it up?

jonathonmcmurray (approved):
couple of formatting nitpicks

PR #352, opened 2020-11-26
Nulltp

### LINKED ISSUE #100: Starting rdb without a tickerplant
While it might not be a conventional approach, at the moment if you try to start an rdb process without supplying a valid tickerplant type (i.e. tickerplanttypes:` in the rdb settings), it results in the rdb getting stuck in a loop where it attempts to subscribe/connect to something that doesn't exist (and incorrectly says subscription successful!).
I think it should be possible to start the rdb with no tickerplant given, such that the subscribe step is skipped when tickerplanttypes is null. It shouldn't cause the process to essentially hang while it keeps trying to connect to a non-existent tickerplant.
In terms of EOD, in this case the rdb could save down all tables in the top namespace (except those in ignorelist), or you could have a more sophisticated system for adding tables to a writedown list (since there would be no .rdb.subtables).
This issue is probably also applicable to the wdb or similar processes that are set up to connect to a tickerplant.

etenno:
resolves #100 

this is an option which allows rdb to not connect to tickerplant on startup

- processes in .servers.CONNECTIONS are connected to on startup
- tickerplant type is held in .rdb.tickerplanttypes, rdb connects and subscribes to tp depending on value of .rdb.connectonstart boolean
- config for feature is held in TorQ-FSP PR # 80


code/processes/rdb.q
```diff
@@ -190,10 +191,16 @@ reload:.rdb.reload
...
+    // removes tp from connection list if in nulltickerplant mode
+    .servers.CONNECTIONS:.servers.CONNECTIONS except .rdb.tickerplanttypes;
+    .rdb.tplogdate:.z.d; // defines tplogdate for setpartition
+    ]
+
 .lg.o[`init;"searching for servers"];
 
 //check if tickerplant is available and if not exit with error 
-.servers.startupdepcycles[.rdb.tickerplanttypes;.rdb.tpconnsleepintv;.rdb.tpcheckcycles]
+.servers.startupdepcycles[.servers.CONNECTIONS;.rdb.tpconnsleepintv;.rdb.tpcheckcycles]
```
jonnypress, line 203: Instead of doing it like this, could we have a variable called .rdb.connectonstart, which defaults to true. If true, it creates the connections using startupdepcycles and subscribes. In both cases I think it probably needs to set the partition.  So similar to what you have, but I think having a variable called nulltickerplant and then removing the tp from the connection list is a bit confusing.

code/processes/rdb.q
```diff
@@ -192,9 +193,13 @@ reload:.rdb.reload
...
 
-//check if tickerplant is available and if not exit with error 
-.servers.startupdepcycles[.rdb.tickerplanttypes;.rdb.tpconnsleepintv;.rdb.tpcheckcycles]
-.rdb.subscribe[]; 
+// connects and subscribes to tickerplant only if connectonstart is true
+$[.rdb.connectonstart;
+ [.servers.CONNECTIONS,:.rdb.tickerplanttypes;
+ .servers.startupdepcycles[.rdb.tickerplanttypes;.rdb.tpconnsleepintv;.rdb.tpcheckcycles];
+ .rdb.subscribe[];]; // defines tplogdate for setpartition
+ .rdb.tplogdate:.z.d;
```
jonnypress, line 201: I think this should use .proc.cd[] to pick the local or GMT date

code/processes/rdb.q
```diff
@@ -192,9 +193,13 @@ reload:.rdb.reload
 
 .lg.o[`init;"searching for servers"];
 
-//check if tickerplant is available and if not exit with error 
-.servers.startupdepcycles[.rdb.tickerplanttypes;.rdb.tpconnsleepintv;.rdb.tpcheckcycles]
-.rdb.subscribe[]; 
+// connects and subscribes to tickerplant only if connectonstart is true
+$[.rdb.connectonstart;
+ [.servers.CONNECTIONS,:.rdb.tickerplanttypes;
```
jonnypress, line 198: Could you add some indentation here to make it more clear that this is two conditions

code/processes/rdb.q
```diff
@@ -10,6 +10,7 @@ hdbtypes:@[value;`hdbtypes;`hdb];                           //list of hdb types
 hdbnames:@[value;`hdbnames;()];                             //list of hdb names to search for and call in hdb reload
 tickerplanttypes:@[value;`tickerplanttypes;`tickerplant];   //list of tickerplant types to try and make a connection to
 gatewaytypes:@[value;`gatewaytypes;`gateway]                //list of gateway types
+connectonstart:@[value;`connectonstart;1b];                 //rdb connects to tickerplant as soon as it is started
```
jonnypress, line 13: This needs to go into config/settings/rdb.q as well

jonnypress (approved):

PR #353, opened 2020-11-27
Add helper functions for test framework

mpotterq:
Some helpful functions for running tests

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -13,6 +13,7 @@ pages:
     - Analytics Library: analyticslib.md 
     - Monitoring: monit.md 
     - Visualisation: visualisation.md
+    - Unit Testing: unittesting.md
     - TorQ Blog Posts: blog.md
 copyright: 'Copyright &copy 2020 AquaQ Analytics Limited.  Kx &reg and kdb+ are registered trademarks of Kx Systems Inc.'
 
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -13,7 +13,7 @@ pages:
     - Analytics Library: analyticslib.md 
     - Monitoring: monit.md 
     - Visualisation: visualisation.md
-    - Unit Testing: unittesting.md
+    - Testing: unittesting.md
     - TorQ Blog Posts: blog.md
 copyright: 'Copyright &copy 2020 AquaQ Analytics Limited.  Kx &reg and kdb+ are registered trademarks of Kx Systems Inc.'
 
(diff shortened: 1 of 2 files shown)


mkdocs.yml
```diff
@@ -13,6 +13,7 @@ pages:
     - Analytics Library: analyticslib.md 
     - Monitoring: monit.md 
     - Visualisation: visualisation.md
+    - Unit Testing: unittesting.md
```
jonnypress, line 16: Could you just call this testing rather than unit testing (this covers more than just unit testing)

tests/helperfunctions.q
```diff
@@ -0,0 +1,25 @@
...
+	};
+
+deadproccheck:{[proctype;procname]
+	/ pairs are of the form "<PID> ssh" for 'PID TTY...' part of pgrep, and "<PID> q" if process is running
+	pidnamepairs:.proc.sys "pgrep -lf \"stackid ",getenv[`KDBBASEPORT]," -proctype ",proctype," -procname ",procname,"\"";
+	not "q" in last each pidnamepairs
+	};
+
+opentorqhandle:{[port]
+	hopen `$"::",port,":admin:admin"
```
jonnypress, line 12: Could you change the password to a variable e.g. testuserpass:"admin:admin" or something. 
Then, when people change the default passwords they just have to change the variable in one of the csv files or other files that they load rather than having to overwrite this functino

jonathonmcmurray: assuming test processes will have a `proctype` that's already got a defined username/pass, could just make this a wrapper of `.servers.opencon` (which accepts an hsym of `` `:host:port``) & automatically use the "right" username/pass for that proctype?

jonnypress: Yeah... good point. I think both maybe have value? Perhaps two methods, along the lines of openbyname and openbyport. If we use port then we are potentially hard coding port numbers into the tests which isn't good. But if we do openbyname then we are requiring that every process that the test process may need to connect to needs to have it in the accesslist, which is also a bit of a pain. But two ways of doing it is probably ok.

mpotterq: So on closer inspection this function isn't actually being used anywhere. Instead connections to procs are done by calling `.servers.startup` and then using the `gethandle` function. Will I just get rid of this or shall I incorporate these suggestions in `gethandle`?

jonnypress: I think gethandle is fine and remove this one then

jonnypress (approved):

PR #356, opened 2020-12-03
nullifying log functionality when sctp not in create mode

etennoQ:
sets handles of logs to null in sctp when not in create logging mode, this prevents errors when data is passed down handles for files that do not exist


code/segmentedtickerplant/stplog.q
```diff
@@ -290,14 +290,24 @@ init:{[dbname]
     // add the info to the meta table
     .stpm.updmeta[multilog][`open;logtabs;.z.p+.eodtime.dailyadj];
     ]
+
+  // set log handles to null in sctp if not in create mode
+  if[.sctp.chainedtp and not .sctp.loggingmode=`create; 
+    `..loghandles set .stplg.t!(count .stplg.t)#(::);
+    // only need to log errors if sctp is creating its own logs
+    `.stplg.badmsg set {[x;y;z]};
```
mpotterq, line 298: With a namespace variable you can just use equals, ie.
`.stplg.badmsg:{[x;y;z]};`

code/segmentedtickerplant/stplog.q
```diff
@@ -290,14 +290,24 @@ init:{[dbname]
...
 \d .
 
 // Close logs on clean exit
 .z.exit:{
   if[not x~0i;.lg.e[`stpexit;"Bad exit!"];:()];
-  .lg.o[`stpexit;"Exiting process and closing off log files."];
-  .stpm.updmeta[.stplg.multilog][`close;.stpps.t;.z.p];
-  .stplg.closelog each .stpps.t;
+  .lg.o[`stpexit;"Exiting process"];
+  if[.sctp.loggingmode=`create; // only close logs if sctp is in create mode
```
mpotterq, line 308: This still needs to be triggered for normal STP usage, probably do something like if its an SCTP and mode is not create then `:()`

code/segmentedtickerplant/sctp.q
```diff
@@ -50,6 +50,6 @@ init:{
 
 // Extract data from incoming table as a list
 upd:{[t;x]
-  x:flip value each x;
```
mpotterq, line 53: Is this based on Jonny's comment from earlier?

etennoQ: yes, sorry unrelated to the logging corrections

mpotterq: haha no problem!

code/segmentedtickerplant/stplog.q
etennoQ: ah i see what you mean, yeah ill just stop the execution early instead

mpotterq:
Looks good to me!

PR #357, opened 2020-12-03
Fix bugs in permissions framework

jonathonmcmurray:
Addressing a couple of bugs in permissions framework:

* expressions such as `first each x` were breaking because `each` becomes lambda on parse & PM framework was always using `value` instead of `eval` when first item in expression is a lambda - switched to `eval` when lambda is from `.q` namespace
* indexing into lists/dicts was broken as variable name was being interpreted by PM framework as a function name

Unit tests for both cases included

### DIFF
--- a/code/handlers/permissions.q
+++ b/code/handlers/permissions.q
@@ -158,8 +158,8 @@ exe:{v:$[(100<abs type first x);val;valp]x;
 
 qexe:{v:val x; if[maxsize<-22!v; 'err[`size][]]; v}
 
-/ check if arg is symbol, and if so if type is <100h i.e. variable
-isvar:{$[-11h<>type x;0b;100h>type get x]}
+/ check if arg is symbol, and if so if type is <100h i.e. variable - if name invalid, return read error
+isvar:{$[-11h<>type x;0b;100h>type @[get;x;{[x;y]'err[`selt][x]}[x]]]}
 
 mainexpr:{[u;e;b;pr]
   / store initial expression to use with value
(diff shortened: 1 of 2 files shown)


code/handlers/permissions.q
```diff
@@ -158,17 +158,24 @@ exe:{v:$[(100<abs type first x);val;valp]x;
 
 qexe:{v:val x; if[maxsize<-22!v; 'err[`size][]]; v}
 
+isvar:{[f]
+  / if first element is not a symbol, not a var ref
+  if[-11h<>type f;:0b];
+  / if first element is symbol, check if it's a variable (type<100h) or named function
+  :$[99h<type get f;0b;1b];
```
ThomasSmyth, line 165: Use `not` instead of conditional?
```suggestion
  :not 99h<type get f;
```
Possible short enough to be one liner after that.

code/handlers/permissions.q
```diff
@@ -158,17 +158,24 @@ exe:{v:$[(100<abs type first x);val;valp]x;
...
   / store initial expression to use with value
   ie:e;
   e:$[10=type e;parse e;e];
   / variable reference
-  if[-11h=type e;
-    if[not achk[u;e;`read;pr]; $[b;'err[`selt][e]; :0b]];
-    $[b; :qexe $[e in key virtualtable;exec (?;table;enlist whereclause;0b;()) from virtualtable[e];e]; :1b];
+  if[isvar f:first e;
+    if[not achk[u;f;`read;pr]; $[b;'err[`selt][f]; :0b]];
+    $[b; :qexe $[f in key virtualtable;exec (?;table;enlist whereclause;0b;()) from virtualtable[f];e]; :1b];
```
ThomasSmyth, line 175: Can bring common explicit return outside.
```suggestion
    :$[b;qexe $[f in key virtualtable;exec (?;table;enlist whereclause;0b;()) from virtualtable[f];e];1b];
```

ThomasSmyth (approved):

jonnypress (approved):

PR #358, opened 2020-12-07
added timezone adjustment

### LINKED ISSUE #354: rolltime in different timezone causes endofday error
existing code assume columns be GMT 
`/-change timeout to zero before eod flush
.timer.repeat[.eodtime.nextroll-00:01;0W;1D;
  (`.rdb.timeoutreset;`);"Set rdb timeout to 0 for EOD writedown"];`

I had issue when the .eodtime.rolltimezone is set to be differently
e.g when i roll data after US market closed (18:00 NYTime, which .eodtime.nextroll=D23:00:00 in UTC)
when tp trigger to load at 18:00, .rdb.timeoutreset has not yet triggered, so .rdb.timeout is not defined and error out during endofday process when trying to restore timeout.

I think the fix should be:
`.timer.repeat[.eodtime.nextroll+.eodtime.adjtime[]-00:01;0W;1D;
  (`.rdb.timeoutreset;`);"Set rdb timeout to 0 for EOD writedown"];`

Or without changing the core processes. I load custom appconfig/code/rdb.q after loading the original rdb.q
`// Torq rdb assume in GMT, adjust to be in .proc.cp timezone
update periodstart:periodstart+.eodtime.adjtime[], nextrun:nextrun+.eodtime.adjtime[] from `.timer.timer where (first each funcparam)=`.rdb.timeoutreset;`

To be safe, I also define `.rdb.timeout:system"T"` in settings/rdb.q

etennoQ:
Problem:
The input given in the RDB proc script for the timeout reset is incorrect when the RDB is running in a different timezone.

Test Case: 
- TP is running in GMT and rolls at 6pm, 
- RDB is in EST time and rolls at 6pm EST time (23:00 GMT)

In this case, the TP rolls before the RDB has had a chance to reset the timeout (which will happen at 22:59 GMT).

Solution:
Include the GMT offset for .eodtime.nextroll by using the difference between the UTC current time and the current time of the RDB. This method accounts for whether the localtime flag is applied or not since .proc.cp[] - .z.p will either be the value of the appropriate time offset or 0 respectively. 

resolves #354

### DIFF
--- a/code/processes/rdb.q
+++ b/code/processes/rdb.q
@@ -200,5 +200,5 @@ reload:.rdb.reload
 .rdb.setpartition[]
 
 /-change timeout to zero before eod flush
-.timer.repeat[.eodtime.nextroll+.eodtime.adjtime[]-00:01;0W;1D;
+.timer.repeat[.eodtime.nextroll+`second$(.proc.cp[]-.z.p)-00:01;0W;1D;
   (`.rdb.timeoutreset;`);"Set rdb timeout to 0 for EOD writedown"];
--- a/code/processes/rdb.q
+++ b/code/processes/rdb.q
@@ -201,5 +201,5 @@ reload:.rdb.reload
 
 /-change timeout to zero before eod flush
 /-GMT offset rounded to nearest 15 mins and added to roll time
-.timer.repeat[.eodtime.nextroll+{00:01*15*"j"$(`minute$x)%15}(.proc.cp[]-.z.p)-00:01;0W;1D;
+.timer.repeat[.eodtime.nextroll-00:01+{00:01*15*"j"$(`minute$x)%15}(.proc.cp[]-.z.p);0W;1D;
   (`.rdb.timeoutreset;`);"Set rdb timeout to 0 for EOD writedown"];
--- a/code/processes/rdb.q
+++ b/code/processes/rdb.q
@@ -200,5 +200,5 @@ reload:.rdb.reload
 .rdb.setpartition[]
 
 /-change timeout to zero before eod flush
-.timer.repeat[.eodtime.nextroll-00:01;0W;1D;
+.timer.repeat[.eodtime.nextroll+.eodtime.adjtime[]-00:01;0W;1D;
   (`.rdb.timeoutreset;`);"Set rdb timeout to 0 for EOD writedown"];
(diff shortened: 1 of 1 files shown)


jonnypress:
> .proc.cp just calls .z.p (current UTC timestamp) rather than .z.P
This isn't true
If you use -localtime then .proc.cp will be .z.P
On linux if you want to test a process in a different timezone you can either use the -o flag to q (e.g. -o -5 would be EST) or you can run the process in that TZ e.g. 
TZ=EST q torq.q .... blah
Can you test it in a different timezone first? 
Because the eodtime stuff is in GMT, you should just be able to do something with the offset to GMT I think  e.g. .proc.cp[] - .z.p (rounded to a sensible value)

code/processes/rdb.q
```diff
@@ -199,6 +199,9 @@ reload:.rdb.reload
 /-set the partition that is held in the rdb (for use by the gateway)
 .rdb.setpartition[]
 
+/-rounds GMT offset to nearest hour
+localoffset: {01:00*"j"$(`minute$x)%60}(.proc.cp[]-.z.p)
+
 /-change timeout to zero before eod flush
-.timer.repeat[.eodtime.nextroll-00:01;0W;1D;
+.timer.repeat[.eodtime.nextroll+localoffset-00:01;0W;1D;
```
jonnypress, line 206: Can you do this without creating the localoffset variable? I assume it isn't used anywhere else. You also need to account for time differences which aren't a full hour (e.g. indian standard time)

code/processes/rdb.q
```diff
@@ -200,5 +200,6 @@ reload:.rdb.reload
 .rdb.setpartition[]
 
 /-change timeout to zero before eod flush
-.timer.repeat[.eodtime.nextroll-00:01;0W;1D;
+/-GMT offset rounded to nearest 15 mins and added to roll time
+.timer.repeat[.eodtime.nextroll+{00:01*15*"j"$(`minute$x)%15}(.proc.cp[]-.z.p)-00:01;0W;1D;
```
jonnypress, line 204: I don't think the -00:01 is correct here - it is being subtracted before the casting isn't it? Should be
.eodtime.nextroll+-00:01+{00:01*15*"j"$(`minute$x)%15}(.proc.cp[]-.z.p)
I think

jonnypress (approved):

PR #360, opened 2020-12-11
Add replaying of zipped files over named pipes

mpotterq:
Add the ability to replay zipped log files in the tickerlogreplay script using named pipes (kdb+ 4.0 and above) and associated unit tests
Branches from SegmentedTP


code/processes/tickerlogreplay.q
```diff
@@ -188,15 +189,17 @@ replaylog:{[logfile]
  .replay.tablecounts:.replay.errorcounts:.replay.pathlist:()!();
 
  // If not running in segmented mode, reset replay date and clean HDB directory on each loop
+ .replay.zipped:$[logfile like "*.gz";1b;0b];
  if[not .replay.segmentedmode;
-   .replay.replaydate:"D"$-10#string logfile;
+   .replay.replaydate:"D"$$[.replay.zipped;-3_-13#;-10#] string logfile;
```
ThomasSmyth, line 194: Need a comment explaining the magic numbers.

code/processes/tickerlogreplay.q
```diff
@@ -188,15 +189,17 @@ replaylog:{[logfile]
...
    if[.replay.clean;.replay.cleanhdb .replay.replaydate]
   ];
 
  if[lastmessage<firstmessage; .lg.o[`replay;"lastmessage (",(string lastmessage),") is less than firstmessage (",(string firstmessage),"). Not replaying log file"]; :()];
  .lg.o[`replay;"replaying data from logfile ",(string logfile)," from message ",(string firstmessage)," to ",(string lastmessage),". Message indices are from 0 and inclusive - so both the first and last message will be replayed"];
  // when we do the replay, need to move the indexing, otherwise we won't replay the last message correctly
-  -11!($[lastmessage<0W; lastmessage+1;lastmessage];logfile);
+  .replay.replayinner[$[lastmessage<0W;lastmessage+1;lastmessage];logfile];
```
ThomasSmyth, line 201: This should be the same - just adding on the boolean outcome.
```suggestion
  .replay.replayinner[lastmessage+lastmessage<0W;logfile];
```

code/processes/tickerlogreplay.q
```diff
@@ -337,31 +352,35 @@ getlogfile:{
...
   .lg.o[`initandrun;"Initialising replay settings."];
   .replay.tablestoreplay:$[`all~first .replay.tablelist;tables[];.replay.tablelist,()];
   .replay.logstoreplay:$[not null .replay.tplogfile;.replay.getlogfile[];.replay.getlogdir[]];
-  if[not count .replay.logstoreplay;.lg.e[`initandrun;m:"No log files found"];'m];
+  if[not count r:.replay.logstoreplay;.lg.e[`initandrun;m:"No log files found"];'m];
 
   // If in segmented mode, get replay date and clean HDB once
   if[.replay.segmentedmode;
-    .replay.replaydate:first l:"D"$(-6_-14#) each string .replay.logstoreplay;
+    .replay.replaydate:first l:"D"$$[first[r] like "*.gz";-9_-17#;-6_-14#] each string r;
```
ThomasSmyth, line 383: Comment to explain magic numbers.

code/processes/tickerlogreplay.q
```diff
@@ -28,6 +28,7 @@ partandmerge:@[value;`partandmerge;0b];                                 // setti
 tempdir:@[value;`tempdir;`:tempmergedir];                               // location to save data for partandmerge replay
 mergenumrows:@[value;`mergenumrows;10000000];                           // default number of rows for merge process
 mergenumtab:@[value;`mergenumtab;`quote`trade!10000 50000];             // specify number of rows per table for merge process
+zipped:@[value;`zipped;0b];                                             // whether in zipped mode or not
```
jonathonmcmurray, line 31: think this comment could be a bit more descriptive, maybe something like

`// enable support for gzipped log files`

code/processes/tickerlogreplay.q
```diff
@@ -188,15 +189,17 @@ replaylog:{[logfile]
  .replay.tablecounts:.replay.errorcounts:.replay.pathlist:()!();
 
  // If not running in segmented mode, reset replay date and clean HDB directory on each loop
+ .replay.zipped:$[logfile like "*.gz";1b;0b];
```
jonathonmcmurray, line 192: if this is being overwritten here, is there any point in having the config value at top of file? Seems like this isn't really a config variable, more an internal flag to know what mode it's running in.

If so, I think line 31 should be removed, as it seems like configuring it differently won't make any difference.

code/processes/tickerlogreplay.q
```diff
@@ -216,11 +219,23 @@ replaylog:{[logfile]
...
-  if[not (`$string dt) in key .replay.hdbdir;:()];
+  if[not (`$sd:string dt) in key .replay.hdbdir;.lg.o[`cleanhdb;"Date ",sd," not in HDB."];:()];
   delpaths:.os.pth each .Q.par[.replay.hdbdir;dt;] each $[`all~first .replay.tablelist;enlist `;.replay.tablestoreplay];
   {.lg.o[`cleanhdb;"Deleting ",x," from HDB."];.os.deldir x} each delpaths;
  };
 
+// Replay log file, if file is zipped and the kdb+ version is at least 4.0 then replay through named pipe
+replayinner:{[msgnum;logfile]
+  if[not .replay.zipped;-11!(msgnum;logfile);:()];
+  if[.z.K<4.0;.lg.e[`replaylog;m:"Zipped log files can only be directly replayed on kdb+ 4.0 or higher"];'m];
```
jonathonmcmurray, line 231: should we also check OS here? I think named pipes are only Linux, or maybe macos as well - pretty sure not Windows anyway.

code/processes/tickerlogreplay.q
```diff
@@ -337,31 +352,35 @@ getlogfile:{
...
   if[()~key hsym d:.replay.tplogdir;.lg.e[`getlogdir;m:"Specified log directory ",string[d]," does not exist"];'m];
+  if[d like "*.gz";.lg.e[`getlogdir;m:"Zipped log directories not supported."];'m];
   $[.replay.segmentedmode;.replay.getstplogs[d];.Q.dd[logdir;] each key logdir:hsym d]
  };
 
 // Use STP meta table and tplogdir to build log names
 getstplogs:{[logdir]
-  metatable:@[get;.Q.dd[hsym logdir;`stpmeta];{.lg.e[`getstpmeta;m:"Log directory must contain valid STP meta table"];'m}];
+  // If meta table is zipped, assume all other logs are zipped as well and build log names accordingly
+  if[z:`stpmeta.gz in key d:hsym logdir;system "gunzip ",1_string .Q.dd[d;`stpmeta.gz]];
```
jonathonmcmurray, line 366:
```suggestion
  if[z:`stpmeta.gz in key d:hsym logdir;system"gunzip ",1_string .Q.dd[d;`stpmeta.gz]];
```

the `system"gunzip ...` probably means this bit won't work on Windows either I think? might need a check for that and/or error trapping here incase the unzip fails etc.

also, it might be possible to avoid unzipping & later rezipping if you can use named pipes & `zcat` to read the file similar to https://github.com/AquaQAnalytics/TorQ-CME/blob/master/code/processes/cmedecoder.q#L35-L36. Not 100% sure if `get` works with named pipes though

mpotterq: Yeah I tried this and you can get the binary contents out via a named pipe but `get` seems to add some kind of header to the binary, so if you `-9!` the pipe output it doesn't recognise it as a q object. Any idea if this can be gotten round?

jonathonmcmurray: If it doesn't work with a simple `get`, probably not worth it tbh

jonnypress (approved):

PR #367, opened 2021-02-22
Update MultipleHosts.md with config script for TorQ install script

JedwardsEdwards:


docs/MultipleHosts.md
```diff
@@ -68,17 +71,69 @@ host,port,proctype,procname
...
+```
+# configurable variables
+PRIMARYHOSTA=prod.a
+PRIMARYHOSTB=prod.b
+SECONDARYHOSTA=dr.a
+SECONDARYHOSTB=dr.b
+STACKENV=prod
+STACKENV_S=dr
+
+
```
jonnypress, line 99: I don't understand everything below here. Why is this all needed? The only thing you should have here are these I think: 

PRIMARYHOSTA=prod.a
PRIMARYHOSTB=prod.b
SECONDARYHOSTA=dr.a
SECONDARYHOSTB=dr.b
STACKENV=prod
STACKENV_S=dr

Can you see if you can make this simpler? All of the below should come as code in the application pack.

JedwardsEdwards: Had a think about this, I could add a settings file for the rdbq/rdbt processes, versions of the process.csv/nontorqprocess.csv (named something like process_multiplehosts.csv by default), and a script to move the files about appropriately and do the sed stuff to the FSP. Then have the config script set the variables and call the script in the FSP with the variables as the arguments. All of the stuff past this point is the same for both the prod and dr versions so I think that would work.

JedwardsEdwards: I've made another update here, mainly reordering sections in order for it to proceed as the installtion should do. 

I also updated the config script to do less, it's still copying the environment variables to setenv.sh but I can scrap that as well if you think that's best. The other thing there being that the sed commands for the process.csv/nontorqprocess/csv run in the root install directory as they will be run before the csv files are copied by the user to appconfig.

jonnypress (approved):

PR #369, opened 2021-03-02
Sctpbug

etennoQ:
reapplying bug fixes that have been overwritten


jonnypress:
@mpotterq can you review this change please - it is backing out a line that you added in the named pipe replay

mpotterq:
Yeah Elliot and I chatted about this yesterday, I got it backwards when merging the changes and thought that the line was the bug fix, but it actually needs to be taken out

mpotterq (approved):

PR #371, opened 2021-03-05
Added gateway functionality for dahsboards

jtennyson724:
Added kxdash.q file which allows async queries while connected to the gateway via kx dashboards

### DIFF
--- a/code/gateway/kxdash.q
+++ b/code/gateway/kxdash.q
@@ -1,4 +1,5 @@
 \d .kxdash
+enabled:@[value;`enabled;{0b}];
 
 // use this to store the additional params that the kx dashboards seem to send in
 dashparams:`o`w`r`limit!(0;0i;0i;0W)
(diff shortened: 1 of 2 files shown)


code/gateway/kxdash.q
```diff
@@ -0,0 +1,53 @@
...
+.gw.formatresponse:{[status;sync;result]
+ if[`kxdash~first result;
+  res:last result;
+  :$[res`status;
+    (`.dash.rcv_msg;res`w;res`o;res`r;res`result);
+    (`.dash.snd_err;res`w;res`r;res`result)]];
+ $[not[status]and sync;'result;result]}
+
+// incorporate dashps into the .z.ps definition
+.z.ps:dashps
```
jonnypress, line 52: Change the definition of the formatresponse function to be in the .kxdash namespace, then add an init function which
a) sets .gw.formatresponse to have the definition from the .kxdash namespace
b) resets .z.ps

Add a variable .kxdash.enabled which defaults to false (same as the way TorQ handles other variables). If set to true, the init function is invoked. 

The definition of.z.ps should also probably be wrapped the way the other message handler definitions are wrapped, e.g. 

.z.ps:{x@y; .kxdash.dashps y}@[value;`.z.ps;{{value x}}]
The init function should be invoked if a

Short version is we probably don't want this to be implemented by default, but we want it to be easy to turn on.

shehir94:
@jonnypress have pushed some updates here addressing comments. Have tested with @MorrisStranger-aq

jonnypress (approved):

PR #378, opened 2021-04-26
TP schema path removed and sublist check added

robertcoen:


code/common/utils.q
```diff
@@ -67,8 +67,10 @@ gettableproperty:extractfromsubdict[;`tableproperties;];   //- extract from `tab
```
jonnypress, line 67: Can you rename this file please as part of this- we shouldn't have code/common/utils.q (utils for what?) It needs a propername, or if this is entirely related to the dataaccess stuff, can it live in the dataaccess directory?

code/dataaccess/schema.q
```diff
@@ -1,7 +1,7 @@
 \d .schema
 
-// load in tickerplant schema to determine default time column
-schema:.proc.loadf[getenv[`TORQAPPHOME],"/database.q"];
+// define the tickerplant to cross check for default timecolumn
+tickplant:`stp1
```
jonnypress, line 4: why is this called tickplant? should it not be tickerplant?

code/dataaccess/schema.q
```diff
@@ -1,7 +1,7 @@
 \d .schema
 
-// load in tickerplant schema to determine default time column
-schema:.proc.loadf[getenv[`TORQAPPHOME],"/database.q"];
+// define the tickerplant to cross check for default timecolumn
+tickerplant:`stp1
```
jonnypress, line 4: I don't think in the dataaccess codebase we should have a dependency on a tickerplant.
If we need to load a schema file, it should be defined separately as a separate piece of config

jonnypress (commented):
@robertcoen @dan-sh Can you guys take a look at this and see if it needs to be merged? I've added a comment on one of the lines

code/dataaccess/schema.q
dan-sh: All done. However, if the user prefers to not fill out the config it'll connect to the stp as a last resort.

jonnypress: The dataaccess code should be completely independent from any tickerplant related schemas. For example, if we have the use case where we don't have a tickerplant (because it is all historic data loaded at T+1) then what is the fallback? I think we should only have three types of parameters:

1. required (must be supplied)
2. required with a default/fallback option (does not need to be supplied)
3. optional (does not need to be supplied)

If the paramter is of type 1 (which I think is what you are saying here) then if it isn't supplied the functionality should fail on initialisation e.g. something like 

q).dataaccess.init[]
'failed to initialise as time column not supplied 

or something simiar, I haven't looked into the syntax. If init is successful, it should return a 1b.

jonnypress (approved):

PR #381, opened 2021-06-07
Fix eodtime not being incremented on SCTP

m-madill:
**Issue:**
``.eodtime.d`` was not being incremented on the segmented chained tickerplant (SCTP) following EOD. A symptom of this was when the SCTP was set to write its own TP logs (``.sctp.loggingmode:`create`` ), all of these logs were written to the same date directory as when the stack was started.

EOD log message showing ``.eodtime.d`` not incrementing on the SCTP:
sctp_eod

TP logs for STP:
stp_tplogs

TP logs for SCTP:
sctp_tplogs

``ls data/tplogs/sctp1_prod_2021.06.03``, shows logs from ``2021.06.04`` being stored in ``2021.06.03``.
sctp_tplogs1

**Fix:**
Move the ``.eodtime.d`` incrementation out of the ``.stplg.stpeod`` function (which is never called by the SCTP), into the shared ``.stplg.dayrollover`` function, meaning that both the standard STP & SCTP will both call this function and increment ``.eodtime.d`` at EOD.

Updated unit tests, to test SCTP successfully increments date and creates a new TP log directory.


jonnypress (commented):
Do we have a test to ensure that if the SCTP isn't running in create mode then a new log directory *doesn't* get created? 
i.e. does this just create a different problem - after one day SCTPs which aren't supposed to create logs start creating them?

m-madill:
> Do we have a test to ensure that if the SCTP isn't running in create mode then a new log directory _doesn't_ get created?
> i.e. does this just create a different problem - after one day SCTPs which aren't supposed to create logs start creating them?

I don't believe we do, I'll add it in

etennoQ (approved):

PR #382, opened 2021-06-15
Fix EOD not occurring when not an EOP

m-madill:
**Issue:**
Setting the end of day on the STP to a time which didn't align with an end of period resulted in the end of day function not being called. This is because the ``isendoday`` variable within [``.stplg.checkends``](https://github.com/AquaQAnalytics/TorQ/blob/master/code/segmentedtickerplant/stplog.q#L248), was only set if the input time was after an end of period.

The issue can be replicated by setting ``eodtime`` to the following and watching the ``STP`` in debug. EOD will happen 42 seconds after STP start up. 
```
\d .eodtime
rolltimeoffset:.z.p+00:00:42-"p"$.z.d+1;        // offset from default midnight roll
datatimezone:`GMT;                      // timezone for TP to timestamp data in
rolltimezone:`GMT;                      // timezone to perform rollover in
```  

**Fix:**
Remove the dependency of the EOD check also being an EOP within the ``.stplg.checkends`` function.

Also updated config files to stop overwrites of custom EOD that could be set in ``default.q``.

**To do:**

- [ ] Update unit tests to catch this case

### DIFF
--- a/config/settings/segmentedtickerplant.q
+++ b/config/settings/segmentedtickerplant.q
@@ -10,19 +10,8 @@ replayperiod:`day               // [period|day|prior]
 customcsv:hsym first .proc.getconfigfile["stpcustom.csv"];       // Location for custom logging mode csv
 
 \d .proc
-
-loadprocesscode:1b
-
-\d .eodtime
-
-datatimezone:@[value;`datatimezone;`GMT];
-rolltimezone:@[value;`rolltimezone;`GMT];
-
-  
-\d .proc
 loadcommoncode:0b               // do not load common code
 loadprocesscode:1b              // load process code
-
 logroll:0b                      // do not roll logs
 
 // Configuration used by the usage functions - logging of client interaction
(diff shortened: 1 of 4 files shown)


config/settings/segmentedtickerplant.q
```diff
@@ -15,8 +15,8 @@ loadprocesscode:1b
 
 \d .eodtime
 
-datatimezone:`GMT
-rolltimezone:`GMT
+datatimezone:@[value;`datatimezone;`GMT];
```
jonnypress, line 18: I don't think either of these definitions are required in here. They should just be inherited from the eodtime code. Can you remove these and re-run the tests?

m-madill: Sure, similar [definitions are defined on FSP](https://github.com/AquaQAnalytics/TorQ-Finance-Starter-Pack/blob/master/appconfig/settings/segmentedtickerplant.q#L20), should I submit a PR to remove these also? So that these options are handled solely by config `default.q`.

jonnypress (approved):

jonnypress:
@m-madill can you merge this now please?

PR #385, opened 2021-09-07
Move .aqrest.execute definition for the Data Access API on the gateway 

nordonez:
.aqrest.execute was defined on gateway startup and this was causing issues with using q-REST. 
Define .aqrest.execute on the gateway after Data Access API initialization instead.

### DIFF
--- a/code/common/dataaccess.q
+++ b/code/common/dataaccess.q
@@ -26,6 +26,7 @@ metainfo:([tablename:`$()]partfield:`$();metas:();proctype:`$());
 //-   - load config for checking input parameters
 //-   - write meta info for tables in current process to .dataaccess.metainfo
 init:{[tablepropertiespath]
+  if[.proc.proctype = `gateway; .aqrest.execute::{[req;props] @[value;req;{(neg .z.w)(.gw.formatresponse[0b;0b;"error: ",x])}]}];
   .lg.o[`.dataaccess.init;"running .dataaccess.init"];
   .proc.loaddir getenv[`KDBCODE],"/dataaccess";
   if[not validtablepropertiespath[];resettablepropertiespath tablepropertiespath];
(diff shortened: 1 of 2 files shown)


code/common/dataaccess.q
```diff
@@ -26,6 +26,7 @@ metainfo:([tablename:`$()]partfield:`$();metas:();proctype:`$());
 //-   - load config for checking input parameters
 //-   - write meta info for tables in current process to .dataaccess.metainfo
 init:{[tablepropertiespath]
+  if[.proc.proctype = `gateway; .aqrest.execute::{[req;props] @[value;req;{(neg .z.w)(.gw.formatresponse[0b;0b;"error: ",x])}]}];
```
jonnypress, line 29: I don't think we want to hardcode based on processtype. We shouldn't require gateways to be off type "gateway" as it is too restrictive

I think what we need to do instead is add a new configuration parameter to the gateway config, which if true resets the definition of .aqrest.execute to the new definition (this will need commented and documented because it will be tricky to find). This will give us more flexibility (e.g. it should be possible to run the GW with DA loaded but not resetting the definition of .aqrest.execute as well)

nordonez: Changed the condition to check the value (true/false) of .aqrest.loadexecute instead of the process type. The default value is set in /config/settings/gateway.q.

code/common/dataaccess.q
```diff
@@ -26,6 +26,7 @@ metainfo:([tablename:`$()]partfield:`$();metas:();proctype:`$());
 //-   - load config for checking input parameters
 //-   - write meta info for tables in current process to .dataaccess.metainfo
 init:{[tablepropertiespath]
+  if[.aqrest.loadexecute; .aqrest.execute::{[req;props] @[value;req;{(neg .z.w)(.gw.formatresponse[0b;0b;"error: ",x])}]}];
```
jonnypress, line 29: Make the definition of .aqrest.execute single colon rather than double colon definition
Also, I think you should check for existence
```suggestion
  if[@[value;.aqrest.loadexecute;0b]; .aqrest.execute:{[req;props] @[value;req;{(neg .z.w)(.gw.formatresponse[0b;0b;"error: ",x])}]}];
```

PR #388, opened 2021-10-31
Gateway .z.ws handling, always execute join function, removed .gw.placehold

DGaquaq:
Changes:
- If `.z.ws` is already defined when loading `code/processes/gateway.q`, incorporate this into the gateway's sync/async handling mechanism (`.gw.call` dictionary)
- Added boolean to `.gw.results` dictionary to mark if the result has been returned from the queried process. This replaces previous `.gw.placehold` functionality, allowing GUID atoms to be returned from gateway queries safely.
-  Changed logic in `checkresults` to always execute the `join` function regardless of the number of returned objects. For custom implementations of TorQ it will be up to the developer to ensure their custom join function can handle single item lists if necessary.

Testing:
- New join logic tested with default join (`raze`) in `.gw.asyncexec` and is working fine. Needs more extensive testing before merge.
- `.z.ws` handling tested and is working fine for calling `.gw.asyncexec` and `.gw.asyncexecjpt`
- Replacement of `.gw.placehold` is working fine with standard TorQ behaviour and basic queries, requires further testing before merge.

### DIFF
--- a/code/gateway/dataaccess.q
+++ b/code/gateway/dataaccess.q
@@ -90,7 +90,7 @@ getdata:{[o]
 // join results together if from multiple processes
 autojoin:{[options]
     // if there is only one proc queried output the table
-    if[1=count options`procs;:{::}];
+    if[1=count options`procs;:first];
     // if there is no need for map reducable adjustment, return razed results
     if[not options`mapreduce;:raze];
     :mapreduceres[options;];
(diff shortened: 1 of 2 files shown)


jonathonmcmurray (approved):
Looks good to me

jonnypress (approved):
Looks good

jonnypress:
@DGaquaq this is approved, can you squash and merge?

PR #389, opened 2021-11-02
Fix pivot function to work with optional args

pkerrAqua:


code/common/analyticslib.q
```diff
@@ -47,7 +47,7 @@ ffill:{[args]
 
 / - General pivot function
 pivot:{[args]
-	checkargs[args;`table`by`piv`var;(" sS")];
+	checkargs[args;`table`by`piv`var`f`g;(" sS")];
```
jonathonmcmurray, line 50: I'm not certain this is the correct fix - I suspect the issue may actually be in the `checkargs` function.

I think the `checkargs` function _should_ be checking that all required keys (2nd arg) are in the dictionary (1st arg). Instead it is currently checking that all keys from the dict are in the list of keys in 2nd arg.

I think making the change here will just enforce that `f` & `g` _must_ be passed (without a type check as there is nothing added to third arg here).

code/common/analyticslib.q
```diff
@@ -47,7 +47,7 @@ ffill:{[args]
...
  	/ - if user has not specified f or g set to defaults
  	if[not all `f`g in key args;
```
jonathonmcmurray, line 52: I think a change is also needed here so that if only one of `f` _or_ `g` is passed, it will be used. As it stands, if only one is passed, both will be set to their default values so the one that is passed will be overwritten.

code/common/analyticslib.q
pkerrAqua: Yeah you are right, currently it is checking if all dictionary keys (1st arg) is in the list of required keys, so I think reversing this line `not all key[args] in k` to `not all k in key[args]` should fix this and allow for the optional args to be optional if only the required keys of `table` `by` `piv` `var` are used. This then also allows the 3rd arg to remain the same as it is an `in` check, the alternative I tried would be `~` but `in` makes more sense as the `~` would need to also include the optional args types to work when they are used and that would cause the `~` to fail when they weren't.

jonathonmcmurray: Yep, sounds good to me 👍

code/common/analyticslib.q
```diff
@@ -48,10 +48,15 @@ ffill:{[args]
...
 	checkargs[args;`table`by`piv`var;(" sS")];
- 	/ - if user has not specified f or g set to defaults
- 	if[not all `f`g in key args;
-    	args[`f]:{[v;P] `$"_" sv' string (v,()) cross P};
-    	args[`g]:{[k;P;c] k,asc c}];
+ 	/ - Check for optional args f and g
+	/ - Check if f optional exists, if empty set as default else leave as input
+	if[not `f in key args;
+	args[`f]:{[v;P] `$"_" sv' string (v,()) cross P};
+	args[`f]];
```
jonathonmcmurray, line 55: Don't think the final line here is necessary/doing anything? Think this can just be

```suggestion
	if[not `f in key args;
	  args[`f]:{[v;P] `$"_" sv' string (v,()) cross P};
	];
```

(also good to indent the code within an `if` statement for readability)

Obviously same applies for `g` as well

jonathonmcmurray (approved):
LGTM

jonnypress (approved):
looks good

PR #391, opened 2021-11-15
Update pub function to properly filter subscriptions

pkerrAqua:
An issue arose when subscribing to a table via dashboards. Using normal u.q (from kx website), .u.pub would publish changes no problem. 

When using torq pubsub.q, .u.pub seemed to be publishing all records of the table being published with the filtered sym list. An example of which is, records for the table called streamSnap were being published with 2500 records at a time. However, the max records that were ever passed into the .u.pub function was 300 (all syms, changes only). Along with raised cpu usage of chrome browser tab to 150%, making the browser unresponsive, it was publishing data not passed into .u.pub.

Therefore pub function within pubsub.q needs updated, so that only filtered data passed into .u.pub is what is being published instead of the entire table.


code/common/pubsub.q
```diff
@@ -73,13 +73,14 @@ selfiltered:{[x;y]
...
-  if[t in subrequestfiltered`tbl;
-    {[t;x] data:eval(?;t;x`filts;0b;x`columns);neg[x`handle](`upd;t;data)}[t;] 
-      each select handle,filts,columns from subrequestfiltered where tbl=t
-  ];
- };
+    if[not count x;:()];
+    if[count h:subrequestall[t];-25!(h;(`upd;t;x))];
+    if[t in .stpps.subrequestfiltered`tbl;
+         {[t;x;sels] data:eval(?;x;sels`filts;0b;sels`columns);
+         //show count data;
```
jonathonmcmurray, line 80: remove this comment

code/common/pubsub.q
```diff
@@ -73,13 +73,14 @@ selfiltered:{[x;y]
...
+         if[count data;neg[sels`handle](`upd;t;data)]}[t;x;]
+    each select handle,filts,columns from .stpps.subrequestfiltered where tbl=t];
```
jonathonmcmurray, line 82: the identation seems to be changed from before - previously the lambda was idented less & the `each ...` line was indented more, is there a reason to change this? It seems clearer to me the way it was before. (Also the placement of the closing `]` from the `if` and the closing `}` of the function)

code/common/pubsub.q
```diff
@@ -73,13 +73,14 @@ selfiltered:{[x;y]
...
-  if[count h:subrequestall[t];-25!(h;(`upd;t;x))];
-  if[t in subrequestfiltered`tbl;
-    {[t;x] data:eval(?;t;x`filts;0b;x`columns);neg[x`handle](`upd;t;data)}[t;] 
-      each select handle,filts,columns from subrequestfiltered where tbl=t
-  ];
- };
+    if[not count x;:()];
+    if[count h:subrequestall[t];-25!(h;(`upd;t;x))];
+    if[t in .stpps.subrequestfiltered`tbl;
+         {[t;x;sels] data:eval(?;x;sels`filts;0b;sels`columns);
```
jonathonmcmurray, line 79: I know this is how it was before, but is there a reason to use `eval` here rather than just directly doing the functional select?

```suggestion
         {[t;x;sels] data:?[x;sels`filts;0b;sels`columns];
```

tests/stp/pubsub/pubsub.csv
```diff
@@ -37,4 +37,10 @@ run,0,0,q,psHandle(upsert;`trade;testtrade),1,,"Update trade table on pubsub"
 true,0,0,q,1~count psHandle(`trade),1,,"Check there is one row in trade table"
 run,0,0,q,neg[psHandle](`.stpps.pubclear;`trade),1,,"Call pubclear funtion"
 true,0,0,q,0~count psHandle(`trade),1,,"Check that trade table has been cleared"
-true,0,0,q,1~.tst.upd,1,,"Check that local UPD function has been called"
\ No newline at end of file
+true,0,0,q,1~.tst.upd,1,,"Check that local UPD function has been called"
+run,0,0,q,upd:{show x:show count y},1,,"Define upd for sub filter tests"
```
jonathonmcmurray, line 41: I think you can get rid of the `show`s from this test, presumably those were only for debugging during development

tests/stp/pubsub/pubsub.csv
```diff
@@ -37,4 +37,10 @@ run,0,0,q,psHandle(upsert;`trade;testtrade),1,,"Update trade table on pubsub"
...
+run,0,0,q,rdbHandle(`.u.sub;`trade;`AAPL`GOOG),1,,"Subscribe to filtered sym list"
```
jonathonmcmurray, line 42: it looks like this test is subscribing to the rdb, which seems unusual - is that the intention? (Typically rdb doesn't have subscribers, so while it probably does test the same code, it's a slightly abnormal use case)

tests/stp/pubsub/pubsub.csv
```diff
@@ -37,4 +37,10 @@ run,0,0,q,psHandle(upsert;`trade;testtrade),1,,"Update trade table on pubsub"
...
+run,0,0,q,rdbHandle".stpps.subrequestfiltered",1,,"Check filtered subscription has passed"
+run,0,0,q,rdbHandle".u.pub[`trade;-10 sublist trade]",1,,"Force publish for the filtered list"
+run,0,0,q,rdbHandle"upd:insert",1,,"Return upd back to default definition"
+true,0,0,q,rdbHandle"(upd)~(insert)",1,,"Check upd has been reset to default"
```
jonathonmcmurray, line 46: I don't think this test is important, I think it's safe to assume that `:` in q works (which is all this is really testing)

tests/stp/pubsub/pubsub.csv
```diff
@@ -37,4 +37,10 @@ run,0,0,q,psHandle(upsert;`trade;testtrade),1,,"Update trade table on pubsub"
...
 run,0,0,q,neg[psHandle](`.stpps.pubclear;`trade),1,,"Call pubclear funtion"
 true,0,0,q,0~count psHandle(`trade),1,,"Check that trade table has been cleared"
-true,0,0,q,1~.tst.upd,1,,"Check that local UPD function has been called"
\ No newline at end of file
+true,0,0,q,1~.tst.upd,1,,"Check that local UPD function has been called"
+run,0,0,q,upd:{show x:show count y},1,,"Define upd for sub filter tests"
+run,0,0,q,rdbHandle(`.u.sub;`trade;`AAPL`GOOG),1,,"Subscribe to filtered sym list"
+run,0,0,q,rdbHandle".stpps.subrequestfiltered",1,,"Check filtered subscription has passed"
+run,0,0,q,rdbHandle".u.pub[`trade;-10 sublist trade]",1,,"Force publish for the filtered list"
+run,0,0,q,rdbHandle"upd:insert",1,,"Return upd back to default definition"
```
jonathonmcmurray, line 45: I don't think `upd` on rdb was ever changed? the above test on line 41 changes the `upd` locally (which makes sense), so I don't think this is needed?

tests/stp/pubsub/pubsub.csv
```diff
@@ -37,4 +37,10 @@ run,0,0,q,psHandle(upsert;`trade;testtrade),1,,"Update trade table on pubsub"
 true,0,0,q,1~count psHandle(`trade),1,,"Check there is one row in trade table"
 run,0,0,q,neg[psHandle](`.stpps.pubclear;`trade),1,,"Call pubclear funtion"
 true,0,0,q,0~count psHandle(`trade),1,,"Check that trade table has been cleared"
-true,0,0,q,1~.tst.upd,1,,"Check that local UPD function has been called"
\ No newline at end of file
+true,0,0,q,1~.tst.upd,1,,"Check that local UPD function has been called"
+run,0,0,q,upd:{show x:show count y},1,,"Define upd for sub filter tests"
+run,0,0,q,rdbHandle(`.u.sub;`trade;`AAPL`GOOG),1,,"Subscribe to filtered sym list"
+run,0,0,q,rdbHandle".stpps.subrequestfiltered",1,,"Check filtered subscription has passed"
+run,0,0,q,rdbHandle".u.pub[`trade;-10 sublist trade]",1,,"Force publish for the filtered list"
```
jonathonmcmurray, line 44: I think you need to add another test to check that this publish worked as expected - currently these tests just make sure that the `.u.pub` function will run, but don't check that it does what it's supposed to

jonathonmcmurray (commented):
Change looks good to me, I think the tests need a little work still.

Also remember to write a useful description for PRs please (give a little context for what's currently broken & how it's being fixed)

tests/stp/pubsub/pubsub.csv
```diff
@@ -37,4 +37,10 @@ run,0,0,q,psHandle(upsert;`trade;testtrade),1,,"Update trade table on pubsub"
 true,0,0,q,1~count psHandle(`trade),1,,"Check there is one row in trade table"
 run,0,0,q,neg[psHandle](`.stpps.pubclear;`trade),1,,"Call pubclear funtion"
 true,0,0,q,0~count psHandle(`trade),1,,"Check that trade table has been cleared"
-true,0,0,q,1~.tst.upd,1,,"Check that local UPD function has been called"
\ No newline at end of file
+true,0,0,q,1~.tst.upd,1,,"Check that local UPD function has been called"
+run,0,0,q,upd:{show x:show count y},1,,"Define upd for sub filter tests"
+run,0,0,q,rdbHandle(`.u.sub;`trade;`AAPL`GOOG),1,,"Subscribe to filtered sym list"
+run,0,0,q,rdbHandle".stpps.subrequestfiltered",1,,"Check filtered subscription has passed"
```
jonathonmcmurray, line 43: this test doesn't "do" anything, it just retrieves a variable from the rdb but never checks if it is as expected etc. I think this can probably be removed tbh, I don't think your tests need to explicitly check the `.stpps.subrequestfiltered` variable - this is probably tested elsewhere, and isn't the behaviour that you're modifying so you don't need to be testing it. As long as the end result is correct (i.e. you receive filtered updates), you can just assume that all the underlying bits are working properly.

code/common/pubsub.q
pkerrAqua: Hi Jonathon, I tried changing the code to this suggestion and it was returning a `'type` error when I tried doing a force publish to check it was working, but it runs without issue using the `eval` line so I'm assuming that's why it is used and will just leave it the way it is

jonnypress (commented):
Is this good to go? Changes look good to me

jonathonmcmurray (approved):

jonathonmcmurray (approved):

PR #392, opened 2021-11-17
Add stripe function

mauricelimaquaq:


code/common/pubsub.q
```diff
@@ -157,3 +157,41 @@ init:{[t]
...
+// for subsequent requests for that symbol, it looks them up
+.ds.stripe:{[input;skey]
+    // If no updates, return
+    $[0=count input;:();];
+
+    sym:distinct input;
+    // Check if .ds.subReq (dict) exists
+    $[`subReq in key`.ds;
+        // If .ds.numSeg changes - reset .ds.subReq
+        [$[.ds.numSeg=1+max key .ds.subReq;;`.ds.subReq set ()!()];
```
jonathonmcmurray, line 188: I think as nothing is being returned from this `$` & only the "else" branch is being used, it might be more clear to use an `if` with the inverse condition

```suggestion
        [if[.ds.numSeg<>1+max key .ds.subReq;`.ds.subReq set ()!()];
```

code/common/pubsub.q
```diff
@@ -157,3 +157,41 @@ init:{[t]
...
+        // If .ds.subReq exists - check for new sym and append to .ds.subReq
+        $[any new:not sym in raze value .ds.subReq;
+            `.ds.subReq set .ds.subReq,' .ds.map[.ds.numSeg;sym where new];
+            ];];
```
jonathonmcmurray, line 192: likewise here, nothing is being returned & only one branch being used, `if` is probably a better fit. I also prefer for the closing `]` from multiline `$` & `if`s to align with the opening line i.e.

```suggestion
        if[any new:not sym in raze value .ds.subReq;
            `.ds.subReq set .ds.subReq,' .ds.map[.ds.numSeg;sym where new];
        ];
    ];
```

PR #393, opened 2021-11-17
Add unit and performance tests for striping function

mauricelimaquaq:


jonathonmcmurray:
It seems to me that this PR includes the actual changes as well as the unit tests, can we just close that other one & keep everything in one PR?

PR #394, opened 2021-11-19
Allow programmatically starting additional TorQ processes from within q

cdempsey99:
…processes in the cloud
The launchprocess.sh script:

- Runs a TorQ process in the background on the same host
- Can take any command line options for a TorQ process
- Requires at least procname and proctype to be passed
- Provides defaults for any parameters not passed
- Starts logging the output of the process as torq.sh does
- Uses exit codes as explained in the comments


launchprocess.sh
```diff
@@ -0,0 +1,110 @@
...
+        fi
+        cnt=$((cnt+1))	
+    done
+else 
+    echo "Error: First two parameters must be procname and proctype"
+    exit 1
+fi
+
+#set default values
+U_arg=${U_arg:=0}
```
jonathonmcmurray, line 56: U is meant to be path to file with usernames & passwords, so I don't think a default value of 0 is appropriate here. Possibly we don't default this here & instead default it in the wrapper q function, where it'll be easier to get the location of the calling process' -U file to use as a default?

jonathonmcmurray (commented):
Looks like a good start. As well as comments here, please run 

```bash
$ shellcheck launchprocess.sh
```

and fix any issues it flags (if you're using homer, it's already installed; on a personal machine you may have to install it)

launchprocess.sh
```diff
@@ -0,0 +1,110 @@
...
+localtime_arg=${localtime_arg:=0}
```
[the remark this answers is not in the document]
cdempsey99: Ok yeah I wasn't sure, will do.

launchprocess.sh
```diff
@@ -0,0 +1,110 @@
...
+g_arg=${g_arg:=0}
+T_arg=${T_arg:=0}
+w_arg=${w_arg:=0}
+qcmd_arg=${qcmd:="q"}
```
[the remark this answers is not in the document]
cdempsey99: No you're right, that's a typo.

launchprocess.sh
```diff
@@ -0,0 +1,110 @@
...
+other_args=${other_args:=" "}
+
+#may be able to just assume that all the relevant env vars are defined but just in case:
```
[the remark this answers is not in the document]
cdempsey99: Ok grand, I also just wanted to have that in so I could have the environment variables sourced while writing the script. But yeah I'll get rid of it.

launchprocess.sh
```diff
@@ -0,0 +1,110 @@
...
+
+
+#check process isn't already running before attempting to start it
+if [[ -z $(findproc "$procname") ]]; then
+    echo "Starting " "$procname"
+    #run process in background and redirect input & output 
+    eval "nohup $sline </dev/null >${KDBLOG}/torq${1}.txt 2>&1 &"
+
+    #after 3s check if the process is running
+    sleep 3s
```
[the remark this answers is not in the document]
cdempsey99: Ok perfect.

launchprocess.sh
```diff
@@ -0,0 +1,110 @@
...
+findproc()  {
+    pgrep -lf "$1" -u $USER | grep -ow "q"
+}
+
+
+#check process isn't already running before attempting to start it
+if [[ -z $(findproc "$procname") ]]; then
+    echo "Starting " "$procname"
+    #run process in background and redirect input & output 
+    eval "nohup $sline </dev/null >${KDBLOG}/torq${1}.txt 2>&1 &"
```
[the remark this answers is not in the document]
cdempsey99: Yes that's a typo

launchprocess.sh
cdempsey99: Ok sounds good.

launchprocess.sh
```diff
@@ -0,0 +1,110 @@
...
+		;;
+	    g )
+		g_arg=${sub_arg_arr[1]}
+		;;
+	    T )
+		T_arg=${sub_arg_arr[1]}
+		;;
+	    w )
+		w_arg=${sub_arg_arr[1]}
+		;;
```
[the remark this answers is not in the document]
cdempsey99: Ok grand

launchprocess.sh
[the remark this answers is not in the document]
jonathonmcmurray: for testing outside of q torq process, you should be able to just manually source the setenv.sh & then run the script (as long as you keep the same shell session open, you won't need to source again after the first time, can just run the script after that)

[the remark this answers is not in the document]
cdempsey99: Oh yeah that works, thanks.

lprocwrapper.q
```diff
@@ -0,0 +1,18 @@
...
+launch:{ [params]
+    /allparams is the list of all possible commmand line options, undef is the subset of these which have not been given a value yet. We add a new key value pair to the dict params for each undef:
+    allparams:`procname`proctype`U`localtime`T;
+    undef:allparams except key params;
+    {[params;x] 
+       $[(x=`T) or (x=`localtime);
+          @[params;x;:;"1"];
+         x=`U;
+          @[params;x;:; getenv[`KDBAPPCONFIG],"/passwords/accesslist.txt"]; 
+         ()]}[params;] each undef;
```
jonathonmcmurray, line 15: Might be worth looking at `.Q.def` here, examples shown here at bottom of section: https://code.kx.com/q/ref/dotz/#zx-argv

cdempsey99: Ah ok I hadn't come across that before, thanks

lprocwrapper.q
```diff
@@ -0,0 +1,18 @@
+/function which lets us call launchprocess.sh from inside a TorQ process
```
jonathonmcmurray, line 1: should this script still be here? looks like it's maybe replaced by `launchprocess.q`?

launchprocess.q
```diff
@@ -0,0 +1,45 @@
+/function which execute system commands from q, and is tailored to deal with the exit codes of lproc.sh:
```
jonathonmcmurray, line 1: I think this file should probably live in `code/common/`

launchprocess.q
```diff
@@ -0,0 +1,45 @@
...
+/cmd is a string and params are the process parameters which are passed to launch() 
+syscomm:{[params;cmd]
+
+    prevexitcode:first `$system cmd,"; echo $?";
```
jonathonmcmurray, line 5: Should cast to an int instead of a symbol given the return code will always be numeric

```suggestion
    prevexitcode:first "I"$system cmd,"; echo $?";
```

launchprocess.q
```diff
@@ -0,0 +1,45 @@
...
+    
+    /set default values with .Q.def and string the result for the startline:
+    /for now, let the set of parameters which need to be provided with defaults be:
+    /(procname,proctype),U,localtime
+
+    deflt:`procname`proctype`U`localtime!(`;`;`$getenv[`KDBAPPCONFIG],"/passwords/accesslist.txt";1);
+    params:string each .Q.def[deflt] params; 
+   
+    $[4<count params;
+          [ extras:(-1*(count params)-4)#params;
```
jonathonmcmurray, line 35: will this work if e.g. `procname` is passed as 5th arg in `params` dict? can't remember if `.Q.def` reorders things, if it does this is maybe fine

but it might be easier to just do the same for all the args, rather than excluding the default ones and then pulling them out manually below?

launchprocess.q
```diff
@@ -0,0 +1,45 @@
...
+    /set default values with .Q.def and string the result for the startline:
+    /for now, let the set of parameters which need to be provided with defaults be:
+    /(procname,proctype),U,localtime
+
+    deflt:`procname`proctype`U`localtime!(`;`;`$getenv[`KDBAPPCONFIG],"/passwords/accesslist.txt";1);
+    params:string each .Q.def[deflt] params; 
+   
+    $[4<count params;
+          [ extras:(-1*(count params)-4)#params; 
+            extraargs:{raze "-",string x[0]," ",string x[1]} each (key extras),'{`$x} each value extras];
```
jonathonmcmurray, line 36: Why cast the values to symbols and then string them?

```suggestion
            extraargs:{"-",string x," ",y}'[key extras;value extras]];
```

this is probably a little cleaner - don't think you should need the `raze`

launchprocess.q
```diff
@@ -0,0 +1,45 @@
...
+           
+          extraargs:" "];
+
+    sline:"bash ",getenv[`TORQHOME],"/launchprocess.sh -procname ",params[`procname]," -proctype ",params[`proctype]," -U ",params[`U]," -localtime ",params[`localtime],raze " ",'extraargs;
```
jonathonmcmurray, line 40:
```suggestion
    sline:"bash ",getenv[`TORQHOME],"/launchprocess.sh -procname ",params[`procname]," -proctype ",params[`proctype]," -U ",params[`U]," -localtime ",params[`localtime]," " sv extraargs;
```

(although if taking above suggestion, you'll also remove procname, proctype etc. here)

lprocwrapper.q
cdempsey99: Yeah I changed the names to launchprocess.sh and launchprocess.q (instead of lprocwrapper.q). But actually since then I've added code for the wrapper for killproccess.sh into that same file, so as it stands there are launchprocess.sh, killprocess.sh and wrappers.q

jonathonmcmurray: OK - just make sure that these changes get reflected on git :+1:

launchprocess.q
cdempsey99: This wasn't working for me originally (because I wasn't using a capital I), that's why I tried using syms instead. Will do

launchprocess.q
cdempsey99: .Q.def does reorder the dictionary in the way that I needed, fortunately.

I see what you're saying, not sure why I got caught up in handling them differently to be honest.

launchprocess.q
cdempsey99: What you suggest there is what I tried initially, but I was getting some issues with the keys being syms and the values strings, i.e the result would be something like ("-";"procname";" ";"w";"d";"b";"1";....) so I ended up needing the raze. In any case, I've reshuffled things since your suggestion to treat all params the same, so everything is much simpler (but I do still have that raze).

launchprocess.q
cdempsey99: Yeah everything here has been changed but I have used that suggestion of sv, thanks.

launchprocess.q
jonathonmcmurray: sorry my suggestion should have been

```suggestion
            extraargs:{"-",string[x]," ",y}'[key extras;value extras]];
```

without the `[]` around `x`, the join will happen first and then `string` will operate on mixed list. this way you should avoid the need to `raze`

cdempsey99: Ah of course yeah that works, thanks.

code/common/cloudutils.q
```diff
@@ -0,0 +1,59 @@
...
+        [$[0=prevexitcode;
+            .lg.o[`killprocess;"Successful execution: Terminating ",params];
+          1=prevexitcode;
+            .lg.e[`killprocess; params," was not running"];
+          2=prevexitcode;
+            .lg.e[`killprocess;"Process name ",params," was not found"];
+          3=prevexitcode;
+            .lg.e[`killprocess;"Failed to terminate "params]
+          .lg.e[`killprocess;"Unknown error encountered"]]] 
+      ]
```
jonathonmcmurray, line 27: there's a lot of duplication here, I think this can probably be simplified a little

```suggestion
    id:$[lok:99h=type params;`launchprocess;`killprocess];
    $[0=prevexitcode;
        .lg.o[id;"Successful execution: ",$[lok;"Starting ";"Terminating "],params[`procname]];
      1=prevexitcode;
        .lg.e[id;"Error: ",params[`procname]," failed to ",$[lok;"start";"terminate"]];
      2=prevexitcode;
        .lg.e[id;"Error: ",params[`procname]," already running"];
        .lg.e[id;"Unknown error encountered"];
    ];
```

Or something like that. I may have lost track of number of square brackets so that might not be quite right but hopefully you get the gist anyway. For exit code of 2, might just have to put the whole error message inside a `$` conditional, instead of only part like with some of the others.

I may also have missed an error code somewhere - looks like kill has 0-3 and launch only has 0-2; if possible, align the error codes as closely as possible to simplify things here. e.g. kill has an error code of 3 when it failed to terminate but launch has an error code of 1 when it failed - better if they both have the same, so maybe swap 1 & 3 for the kill script, if that makes sense?

code/common/cloudutils.q
```diff
@@ -0,0 +1,59 @@
...
+kill:{[procname]
+
+    syscomm[procname;] "bash ",getenv[`TORQHOME],"/killprocess.sh ",procname   }
+
+
+
+
+
+
+
```
jonathonmcmurray, line 59: don't need all these empty lines

code/common/cloudutils.q
```diff
@@ -0,0 +1,59 @@
...
+    f_args:{"-",string[x]," ",y}'[key params;value params];
+    sline: "bash ",getenv[`TORQHOME],"/launchprocess.sh "," " sv f_args;
+    syscomm[params;] sline     }
+
+
+/this function to terminate a given TorQ process from within another process, 
+/takes a single parameter, a string procname
+kill:{[procname]
+
+    syscomm[procname;] "bash ",getenv[`TORQHOME],"/killprocess.sh ",procname   }
```
jonathonmcmurray, line 52: with a func this short, I'd just define on one line & get rid of some whitespace

```suggestion
kill:{[procname]syscomm[procname;]"bash ",getenv[`TORQHOME],"/killprocess.sh ",procname}
```

code/common/cloudutils.q
```diff
@@ -0,0 +1,59 @@
...
+            .lg.e[`killprocess;"Failed to terminate "params]
+          .lg.e[`killprocess;"Unknown error encountered"]]] 
+      ]
+   }
+
+/function which lets us call launchprocess.sh from inside a TorQ process
+/it takes a dictionary of parameters which will be passed to launchprocess.sh, i.e "-procname rdb1 -proctype rdb" etc.
+launch:{ [params]
+    /exit immediately if process name and type aren't provided
+    if[not 11b~`procname`proctype in key params;0N!
```
jonathonmcmurray, line 34: I imagine the `0N!` on this line is a remnant from some debugging? Presumably this should be removed. Can also use `all` here - should do the same thing, think it's just a little more immediately clear

```suggestion
    if[not all `procname`proctype in key params;
```

code/common/cloudutils.q
```diff
@@ -0,0 +1,59 @@
...
+        .lg.e[`launchprocess;"Process name and type must be provided"];
+        :()];
+
+    /set default values with .Q.def and string the result:
+    deflt:`procname`proctype`U`localtime!(`;`;`$getenv[`KDBAPPCONFIG],"/passwords/accesslist.txt";1);
```
jonathonmcmurray, line 39: I think default for localtime should be `.proc.localtime` i.e. whatever it's set to in the parent process.

I think we should possibly also try to use the `-U` setting from the parent process, maybe falling back to what you have here if that's not present.

I think we can do something like:

```
first`$.Q.opt[.z.X]`U
```

to try and get the `-U` arg of the current process (`.proc.params` that is defined in TorQ uses `.z.x` rather than `.z.X`, so it doesn't include kdb's single letter args like `-U`)

I'd consider inlcuding default for port & qcmd here as well, just to simplify the launch bash script a little? (Unless there's a reason to do it there instead, of course)

code/common/cloudutils.q
```diff
@@ -0,0 +1,59 @@
...
+    params:string each .Q.def[deflt] params;
+    
+    /format the params dictionary as a series of command line args
+    f_args:{"-",string[x]," ",y}'[key params;value params];
+    sline: "bash ",getenv[`TORQHOME],"/launchprocess.sh "," " sv f_args;
```
jonathonmcmurray, line 44:
```suggestion
    sline:"bash ",getenv[`TORQHOME],"/launchprocess.sh "," " sv f_args;
```

killprocess.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
jonathonmcmurray, line 1: I think it'd be neater if we made a `bin/` directory at the top level of TorQ directory to hold executable scripts, like this one & the launchprocess one (just so we're not cluttering the top level of the repo with these scripts & any others in future)

killprocess.sh
```diff
@@ -0,0 +1,43 @@
...
+            exit 3
+        fi
+    else
+        exit 2
+    fi
+fi
+
+
+
+
```
jonathonmcmurray, line 43: Remove blank lines (can leave one at the end of file, but more than that is excessive)

launchprocess.sh
```diff
@@ -30,27 +31,40 @@ do
...
 #define the startline
-sline="q ${TORQHOME}/torq.q -stackid ${KDBBASEPORT} -proctype $proctype -procname $procname -U $U_arg -localtime $localtime_arg "
+sline="q ${TORQHOME}/torq.q -stackid ${KDBBASEPORT} -proctype $proctype -procname $procname -U $U_arg -localtime $localtime_arg -p $p_arg $extra_arg_arr"
 
 #fn to check if a process is running, returns a non-zero search result for success, returns null for failure
 findproc()  {
     pgrep -lf "$1" -u "$USER" | grep -ow "q"
 }
 
+echo "$sline"
```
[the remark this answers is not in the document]
jonathonmcmurray: Make sure this gets removed please

jonathonmcmurray (commented):
On the whole, this looks good. I've left a few comments, mostly these are styling/formatting things.

I think next couple of things that are needed are (i) unit tests and (ii) documentation.

For unit tests, I think it's probably something like

1. Start a new process (probably with some extra command line args)
2. Wait a second or two for it to start
3. Attempt to connect to new process
4. Query new process to verify that all command line args got passed properly (i.e. they're in `.proc.params` as appropriate) and get the PID to check later it's been killed
5. Kill the process
6. Confirm that the handle to that process got invalidated & process is killed

And probably do the same with minimal args passed to start process to make sure defaulting all works properly as well.

For docs, I think most sensible place to add is likely https://github.com/AquaQAnalytics/TorQ/blob/master/docs/utilities.md and it should cover what the code does & how to use it etc.

code/common/cloudutils.q
```diff
@@ -0,0 +1,41 @@
+/a function to execute system commands and return a log message depending on the resulting exit code:
+/used for both launchprocess.sh and killprocess.sh
+syscomm:{[params;cmd]
+    /params is (i) a dictionary if syscomm has been called by launch, and (ii) a string if it has been called by killproc
+    /cmd is the command to be executed, as a string
+    
+    prevexitcode:first "I"$system cmd,"; echo $?"; 
+    $[lok:99h=type params;[id:`launchprocess;pname:params[`procname]];[id:`killprocess;pname:params]];
```
jonathonmcmurray, line 8: I think it's a little clearer to do this as two conditionals:

```suggestion
    id:$[lok:99h=type params;`launchprocess;`killprocess];
    pname:$[lok;params`procname;params];
```

IMO the code-block-within-$ syntax gets confusing very quickly and this just makes it a little clearer at a glance where these two variables are being defined

cdempsey99:
Hi Jonathon,

Hope you had a nice Christmas break! First of all, thanks for the tip on skype regarding the TorQ on homer thing, but I was off on leave before I got your message unfortunately.

Anyway I'm writing these unit tests and the first case of launching a process with lots of cmd line args has gone fine, but I've run into a bit of an issue in the second case. In this case I'm trying to launch a process with minimal cmd line args, to test the defaulting. This means that the port is defaulted to 0W, and is randomly selected.

As far as I can see, this presents a difficulty, as I need to connect to the process to check the command line parameters passed, but I don't see how to do this without knowing the port number. For the moment I'm trying to search through all possible port numbers in the range 32768-60999 but I was wondering if there were maybe another way that I'm missing. No worries if not, I think my method will work, it just seems a bit long winded.

Thanks,
Callum

________________________________

tests/cloudlaunchprocess/test.csv
```diff
@@ -0,0 +1,27 @@
...
+run,0,0,q,kill "wdb1",,"Kill the process"
+true,0,0,q,10h=type @[{h x};".wdb.writedownmode";{"Error: failure to connect"}],,"Check that the handle to the process is no longer open"
+true,0,0,q,not first "q" in' b:.proc.sys "pgrep -lf wdb1",,"Check that the PID of the old process is no longer in use"
+
+comment,0,,"Test 2 - Launch and then kill a process with no extra cmd line args, to test defaulting"
+run,0,0,q,launch input2,,,"Start a new process with no extra command line args, hence relying on defaulting in cloudutils.q"
+run,0,0,q,.proc.sys "sleep 2",,,"Wait for process to start"
+run,0,0,q,"h2:hopen `$""::"",string[discport],"":admin:admin""",,,"Connect to discovery1"
+run,0,0,q,"proc_port:""I""$(neg count string pid)#.Q.s1 h2""exec hpup from .servers.SERVERS where not null w,proctype=`wdb""",,,"Find the process port via discovery"
+run,0,0,q,"h3:hopen `$""::"",string[proc_port],"":admin:admin""",,,"Attempt to connect to the new process"
```
jonathonmcmurray, line 19: think this can be simplified to

```suggestion
run,0,0,q,"h3:hopen `$string[proc_port],"":admin:admin""",,,"Attempt to connect to the new process"
```

if you don't remove the hostname from the hpup to begin with

tests/cloudlaunchprocess/test.csv
```diff
@@ -0,0 +1,27 @@
...
+run,0,0,q,pid:"I"$-2 _ raze raze string a where "q" in' a:.proc.sys "pgrep -lf wdb1",,"Find the PID of the launched process"
+run,0,0,q,kill "wdb1",,"Kill the process"
+true,0,0,q,10h=type @[{h x};".wdb.writedownmode";{"Error: failure to connect"}],,"Check that the handle to the process is no longer open"
+true,0,0,q,not first "q" in' b:.proc.sys "pgrep -lf wdb1",,"Check that the PID of the old process is no longer in use"
+
+comment,0,,"Test 2 - Launch and then kill a process with no extra cmd line args, to test defaulting"
+run,0,0,q,launch input2,,,"Start a new process with no extra command line args, hence relying on defaulting in cloudutils.q"
+run,0,0,q,.proc.sys "sleep 2",,,"Wait for process to start"
+run,0,0,q,"h2:hopen `$""::"",string[discport],"":admin:admin""",,,"Connect to discovery1"
+run,0,0,q,"proc_port:""I""$(neg count string pid)#.Q.s1 h2""exec hpup from .servers.SERVERS where not null w,proctype=`wdb""",,,"Find the process port via discovery"
```
jonathonmcmurray, line 18: I think here you can use `.servers.querydiscovery` https://github.com/AquaQAnalytics/TorQ/blob/6cec4609ca7931983b527987202d9855df98ba87/code/handlers/trackservers.q#L205

So something like

```suggestion
run,0,0,q,proc:.servers.querydiscovery`test,,,"Find the process port via discovery"
```

this will return a table of all processes discovery knows about with proctype `` `test`` - you can then do e.g. ``exec hpup from proc where procname=`test3`` to get the hpup for the proc in question

I think your code to extract just the port number is also slightly flawed (pid could be a different length to port number) and unnecessary - see comment on line below

tests/cloudlaunchprocess/test.csv
```diff
@@ -0,0 +1,27 @@
...
+run,0,0,q,"h3:hopen `$""::"",string[proc_port],"":admin:admin""",,,"Attempt to connect to the new process"
+true,0,0,q,`default=h3".wdb.writedownmode",,,"Check that connection has been made"
+true,0,0,q,all value input2 in 2 _ (2 cut h3".z.X")[;1],,,"Check that correct process name and type are present in the cmd line args of the process"
```
jonathonmcmurray, line 21: similar comments as above regarding these checks

tests/cloudlaunchprocess/test.csv
```diff
@@ -0,0 +1,27 @@
...
+true,0,0,q,"all (getenv[`KDBAPPCONFIG],""/passwords/accesslist.txt"";string .proc.localtime;""0W"") in 2 _ (2 cut h3"".z.X"")[;1]",,,"Check that the expected default parameters for the U,localtime and p flags are present in the cmd line args"
+run,0,0,q,pid:"I"$-2 _ raze raze string a where "q" in' a:.proc.sys "pgrep -lf wdb1",,,"Find the PID of the launched process"
+run,0,0,q,kill "wdb1",,,"Kill the process"
+true,0,0,q,@[{h4 x};".wdb.writedownmode";{1b}],,,"Check that the handle to the process is no longer open"
```
jonathonmcmurray, line 25: probably easier to check against `.z.W` e.g.

```suggestion
false,0,0,q,h4 in key .z.W,,,"Check that the handle to the process is no longer open"
```

or using TorQ function to do the same

```suggestion
false,0,0,q,.dotz.liveh h4,,,"Check that the handle to the process is no longer open"
```

tests/cloudlaunchprocess/test.csv
```diff
@@ -0,0 +1,27 @@
...
+true,0,0,q,not first "q" in' b:.proc.sys "pgrep -lf wdb1",,,"Check that the PID of the old process is no longer in use"
```
jonathonmcmurray, line 26: I might be missing something here, but this doesn't explicitly check the PID is no longer in use?

jonathonmcmurray (commented):
Got a few comments on the unit tests - for the most part the logic looks sound, just have a few suggestions to make more use of TorQ functionality etc. to simplify things

tests/cloudlaunchprocess/test.csv
```diff
@@ -0,0 +1,27 @@
+action,ms,bytes,lang,code,repeat,minver,comment
+comment,0,,"Test 1 - Launch and then kill a process with all possible cmd line args"
+before,0,0,q,discport:1+"I"$getenv[`KDBBASEPORT] ,,"Find the port of the discovery process"
```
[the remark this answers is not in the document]
cdempsey99: Ok I see what you're saying here but when I try to look at .servers.SERVERS directly from the test process it's empty, so I'm not sure how to find the port on which the connection to discovery1 has been made? (Apart from just relying on the fact that discovery1 is configured to be on the first port after KDBBASEPORT, as above. )

I know .z.W[] shows the open handles but I don't know how to use those ints to actually send messages down the handle.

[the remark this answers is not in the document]
jonathonmcmurray: Yeah, but my point is more that you shouldn't _need_ to know the port - you should be able to just use the functions TorQ provides to query the discovery process and let it worry about the details in the background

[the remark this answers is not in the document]
cdempsey99: Ok I see, thanks

tests/cloudlaunchprocess/test.csv
```diff
@@ -0,0 +1,27 @@
...
+run,0,0,q,launch input1,,"Start a new process with extra command line args"
+run,0,0,q,system "sleep 3",,"Wait for process to start"
+run,0,0,q,h:hopen `::7124:admin:admin ,,"Attempt to connect to the new process"
+true,0,0,q,`default=h".wdb.writedownmode",,"Check that connection has been made"
+true,0,0,q,110011011b~(2 _ (2 cut h".z.X")[;1]) in value input1,,"Check that the cmd line args are as expected "
```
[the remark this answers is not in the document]
cdempsey99: I agree that it is a bit of an ugly solution but the problem is that some values don't agree exactly between input1 and the cmd line parameters of the process. 

For example variables like ${KDBAPPCONFIG}, which has its variable name in input1, but the actual value of the variable in the cmd line parameters. Also some values in input1 end up being enlisted in the cmd line parameters and so the test isn't recognising that "1" and ,"1" are equal. 

That's why I ended up resorting to that binary string.

I will definitely use .Q.opt h".z.X" instead of my other way though.

tests/cloudlaunchprocess/test.csv
```diff
@@ -1,27 +1,25 @@
...
+run,0,0,q,.sys.launch input2,,,"Start a new process with no extra command line args, hence relying on defaulting in cloudutils.q"
+run,0,0,q,system "sleep 2",,,"Wait for process to start"
+run,0,0,q,proc_port:first exec hpup from .servers.querydiscovery`test where procname=`test3,,,"Find process port from discovery services "
+run,0,0,q,"h2:hopen `$string[proc_port],"":admin:admin""",,,"Attempt to connect to the process"
+true,0,0,q,`test3=@[h2;".proc.procname";`],,,"Check that connection has been made"
+true,0,0,q,110b~(value input2)in' value key[input2]#.Q.opt h2".z.X",,,"To be reworked:Check that correct process name and type are present in the cmd line args of the process"
+true,0,0,q,"all (getenv[`KDBAPPCONFIG],""/passwords/accesslist.txt"";string .proc.localtime;""0W"")in' value `U`localtime`p#.Q.opt h2"".z.X""",,,"Check that the expected default parameters for the U,localtime and p flags are present in the cmd line args"
+run,0,0,q,pid2:h2".z.i",,,"Find the PID of the launched process"
+run,0,0,q,.sys.kill "test3",,,"Kill the process"
+true,0,0,q,not h2 in key[.z.W],,,"Check that the handle to the process is no longer open"
```
cdempsey99, line 23: I've made all the recommended changes and everything is working except this line (and line 10, which is very similar).
 
The test gives an error when I run the run.sh script but if I test it manually in the same debugger session, just after the tests are finished, it works fine.

The last time something like this happened it was because there was a comma in the code cell, meaning that the csv was being misinterpreted. But this line is so simple I don't know what could be causing it.

jonathonmcmurray: It gives an error, or the test fails? If it gives an error, I think if you add `-stop` to the command line when starting the test process (i.e. add that inside run.sh), it should drop you into debug mode at the point where it errors and you can poke around.

If it's just the test failing, you might just need to add a sleep after the kill to give time for the other proc to die

code/common/cloudutils.q
jonathonmcmurray: Can you have another look at this comment?

code/common/cloudutils.q
jonathonmcmurray: Can you have another look at this comment?

tests/cloudlaunchprocess/test.csv
```diff
@@ -0,0 +1,25 @@
+action,ms,bytes,lang,code,repeat,minver,comment
+comment,0,,"Test 1 - Launch and then kill a process with all possible cmd line args"
+run,0,0,q,.sys.launch input1,,,"Start a new process with extra command line args"
+run,0,0,q,system "sleep 2",,,"Wait for process to start"
```
jonathonmcmurray, line 4: torq actually has a function for this

```suggestion
run,0,0,q,.os.sleep 2,,,"Wait for process to start"
```

(same below on line 15 and anywhere else there's a sleep)

tests/cloudlaunchprocess/test.csv
```diff
@@ -0,0 +1,25 @@
...
+run,0,0,q,h:hopen `::7124:admin:admin,,,"Attempt to connect to the new process"
+true,0,0,q,`test2=@[h;".proc.procname";`],,,"Check that connection has been made"
+true,0,0,q,11011111010b~(value input1) in' value key[input1]#.Q.opt (h)".z.X",,,"Tentatively? - Check that the cmd line args are as expected"
```
jonathonmcmurray, line 7:
```suggestion
true,0,0,q,11011111010b~(value input1) in' value key[input1]#.Q.opt h".z.X",,,"Tentatively? - Check that the cmd line args are as expected"
```

don't need those parentheses

also the comment says "Tentatively?" - does that need to be updated?

tests/cloudlaunchprocess/test.csv
```diff
@@ -0,0 +1,25 @@
...
+true,0,0,q,not h in key[.z.W],,,"Check that the handle to the process is no longer open"
+true,0,0,q,not pid1 in "I"$1 _ .proc.sys "ps -u $USER | awk '{print $1}'",,,"Check that the PID of the old process is no longer in use"
+
+comment,0,,"Test 2 - Launch and then kill a process with no extra cmd line args, to test defaulting"
+run,0,0,q,.sys.launch input2,,,"Start a new process with no extra command line args, hence relying on defaulting in cloudutils.q"
+run,0,0,q,system "sleep 2",,,"Wait for process to start"
+run,0,0,q,proc_port:first exec hpup from .servers.querydiscovery`test where procname=`test3,,,"Find process port from discovery services "
+run,0,0,q,"h2:hopen `$string[proc_port],"":admin:admin""",,,"Attempt to connect to the process"
+true,0,0,q,`test3=@[h2;".proc.procname";`],,,"Check that connection has been made"
+true,0,0,q,110b~(value input2)in' value key[input2]#.Q.opt h2".z.X",,,"To be reworked:Check that correct process name and type are present in the cmd line args of the process"
```
jonathonmcmurray, line 19: as per comment, is this still to be reworked or should comment be updated to remove that?

jonathonmcmurray (commented):
couple of older comments still outstanding, please check if you want to make any changes based on them etc.

tests/cloudlaunchprocess/test.csv
cdempsey99: It gives an error, rather than just failing to return true. And yes I tried adding in a sleep statement after the .sys.kill. Ok I'll use that -stop command thanks

cdempsey99: I wasn't sure before but looking at the error in debug mode it seems that when the test2/test3 processes are killed the handles h/h2 stay in .z.W, even if they are no longer valid. 

hclose must be used to remove them from.z.W, but this doesn't reflect whether or not the test processes are still runnning. 

As a result I think I'll just keep the previous test: @[h;".proc.procname";{1b}] if that's ok?

jonathonmcmurray: Hmm, that seems strange. OK sure

docs/utilities.md
```diff
@@ -1400,4 +1400,43 @@ One final important variable is ```.grafana.del```, this dictates the delimeter
...
+The .sys.kill function is passed a single argument: the name of the process to
+be terminated, as a string. 
+
+If the default value of "0W" is used for the port, a random port will be chosen
+on which to launch the process. In this case the process will neeed to register
+itself with discovery in order for other processes to be able to connect to it.
+This is a standard behaviour for TorQ processes on startup (for more information
+see [Connection Management](http://aquaqanalytics.github.io/TorQ/conn/#connections)).  
+
+
```
cdempsey99, line 1441: I know the formatting is off, as I'm not 100% on how to format .md files yet, but I thought I'd just submit the content for the moment at least.

docs/utilities.md
```diff
@@ -1400,4 +1400,43 @@ One final important variable is ```.grafana.del```, this dictates the delimeter
 between options in the drop down menus. This has significant repercussions if 
 one of your columns includes full stops, eg. email adresses. As a result we have 
 left this as definable so that the user can alter this to a non-disruptive value 
-for their data eg./.
\ No newline at end of file
+for their data eg./.
+
+## cloudutils.q
```
jonathonmcmurray, line 1405: I think this should be further up, before the stuff about API table etc.

docs/utilities.md
```diff
@@ -1400,4 +1400,43 @@ One final important variable is ```.grafana.del```, this dictates the delimeter
...
+The cloud utilities allow other processes to be programmatically launched, or 
+terminated, from inside a TorQ process. The two main functions here are 
+.sys.launch and .sys.kill. 
+
+.sys.launch is the q function which takes a dictionary of input parameters which
+are then passed to a bash script. This bash script functions like torq.sh in how
+it starts processes. It is important to note that the cloud utilities are only
+supported on Linux as a result. 
+
+    `q)input:`procname`proctype`localtime`p!("hdb1";"hdb";"0";"1234");
```
jonathonmcmurray, line 1416:
````suggestion
```q
q)input:`procname`proctype`localtime`p!("hdb1";"hdb";"0";"1234");
```
````

think this should be a code block

File: docs/utilities.md [hunk 39]
```diff
@@ -1400,4 +1400,43 @@ One final important variable is ```.grafana.del```, this dictates the delimeter
...
+
+The input parameter dictionary, as shown above, should contain symbols as keys 
+and strings as values. Any standard or custom TorQ process can be launched 
+using .sys.launch, and as such the function can accept any desired command line
+parameters in the input dictionary. The minimum required are `procname and 
+`proctype. In the case that only these two are used the other arguments will be
```
jonathonmcmurray, line 1422:
```suggestion
parameters in the input dictionary. The minimum required are `` `procname`` and 
`` `proctype``. In the case that only these two are used the other arguments will be 
```

think you need to do this to prevent inline code formatting messing up here

docs/utilities.md
```diff
@@ -1400,4 +1400,43 @@ One final important variable is ```.grafana.del```, this dictates the delimeter
...
+given default values. 
+
+|    Parameter   |                                                    Default Value                                                          |
+| :------------: | :-----------------------------------------------------------------------------------------------------------------------: |
+|      U         | The password file used for the parent launching process, if none exists ${KDBAPPCONFIG}/passwords/accesslist.txt is used  |
+|  localtime     |                                 The .proc.localtime value of the parent launching process                                 |
+|      p         |                                             0W - a random port will be chosen                                             |
+|     qcmd       |                                                          q                                                                |
+
+The .sys.kill function is passed a single argument: the name of the process to
```
jonathonmcmurray, line 1432:
```suggestion
The .sys.kill function is passed a single argument: the procname of the process to
```

just to make it completely clear

docs/utilities.md
```diff
@@ -1400,4 +1400,43 @@ One final important variable is ```.grafana.del```, this dictates the delimeter
...
+be terminated, as a string. 
+
+If the default value of "0W" is used for the port, a random port will be chosen
+on which to launch the process. In this case the process will neeed to register
+itself with discovery in order for other processes to be able to connect to it.
+This is a standard behaviour for TorQ processes on startup (for more information
+see [Connection Management](http://aquaqanalytics.github.io/TorQ/conn/#connections)).  
+
+
+
```
jonathonmcmurray, line 1442: get rid of a couple of these blank lines at the end (leave 1)

jonathonmcmurray (commented):
Few minor tweaks to documentation but otherwise looks good.

One other point - I think maybe we should rename this stuff. It's not necessarily cloud specific (that's just the use case that prompted us to add this), it'll work and could be useful elsewhere. Maybe something like "launchbg" (given that it launches background processes)?

docs/utilities.md
cdempsey99: Sorry just to be clear do you mean move the whole section to just before the Full API section, or add the new namespace/ fns to the Full API list and API table or both?

jonathonmcmurray: The first

docs/utilities.md
```diff
@@ -1128,6 +1128,46 @@ this, we’ve modified u.q. This can be turned off by setting .u.broadcast
...
+.sys.bglaunch and .sys.bgkill.
+
+.sys.bglaunch is the q function which takes a dictionary of input parameters which
+are then passed to a bash script. This bash script functions like torq.sh in how
+it starts processes. It is important to note that the background launch utilities
+are only supported on Linux as a result.
+
+```q
+
+    `q)input:`procname`proctype`localtime`p!("hdb1";"hdb";"0";"1234");
```
jonathonmcmurray, line 1144:
```suggestion
q)input:`procname`proctype`localtime`p!("hdb1";"hdb";"0";"1234");
```

jonathonmcmurray (approved):
Formatting on the code block, but otherwise looks good

PR #395, opened 2021-11-23
Striping data in a TorQ Installation

mauricelimaquaq:


code/common/pubsub.q
```diff
@@ -157,3 +157,42 @@ init:{[t]
...
+// Get the number of rdb processes from process.csv
+if[.ds.numSeg=0;.ds.numSeg:sum `rdb=((.proc`readprocs).proc`file)`proctype]
+
+// Striping function which stores the mappings for any symbols that it has already computed and 
+// for subsequent requests for that symbol, it looks them up
+.ds.stripe:{[input;skey]
+    // If no updates, return
+    $[0=count input;:();];
+
+    sym:distinct input;
```
jonathonmcmurray, line 184: condense this a little & use `if` instead of `$`:

```suggestion
    if[0=count sym:distinct input;:()];
```

code/common/pubsub.q
```diff
@@ -157,3 +157,42 @@ init:{[t]
...
+    // Check if .ds.subReq (dict) exists
+    $[`subReq in key`.ds;
+        // If .ds.numSeg changes - reset .ds.subReq
+        [if[.ds.numSeg<>1+max key .ds.subReq;`.ds.subReq set ()!()];
+        // If .ds.subReq exists - check for new sym and append to .ds.subReq
+        if[any new:not sym in raze value .ds.subReq;
+            `.ds.subReq set .ds.subReq,' .ds.map[.ds.numSeg;sym where new];
+        ];
+    ];
```
jonathonmcmurray, line 193: could do with a slight layout change to make this a little more clear IMO

```suggestion
        [
            if[.ds.numSeg<>1+max key .ds.subReq;`.ds.subReq set ()!()];
            // If .ds.subReq exists - check for new sym and append to .ds.subReq
            if[any new:not sym in raze value .ds.subReq;
                `.ds.subReq set .ds.subReq,' .ds.map[.ds.numSeg;sym where new];
            ];
        ];
```

code/common/pubsub.q
```diff
@@ -157,3 +157,42 @@ init:{[t]
...
+        // Initialize .ds.subReq if does not exists
+        `.ds.subReq set .ds.map[.ds.numSeg;sym]
+        ];
```
jonathonmcmurray, line 196: if I'm following this right, I think this is overly indented

```suggestion
    ];
```

code/common/pubsub.q
```diff
@@ -157,3 +157,42 @@ init:{[t]
   val:.u.sub[`$tab;1!enlist `tabname`filters`columns!(`$tab;filters;columns)];
   $[10h~type last val;'last val;val]
  };
+
+// Striping data in a TorQ Installation
+// Limitation: only max 16 (hexidecimal digits 0-9 A-F) processes using this logic
+// Can expand beyond 16 processes if required - Modify lookup and modMd5 function to use more digits e.g. "8b" for a single map
+.ds.lookup:{[numSeg;maxProc]
```
jonathonmcmurray, line 164: we generally don't use camelCase in torq, just all lowercase for variables etc.

tests/stp/stripe/run.sh
```diff
@@ -0,0 +1,69 @@
...
+# Start a new session (kdbProcess) detached (-d)
+tmux new-session -d -s $session
+# Wait till session and window is created
+while ! tmux has-session -t $session:bash; do sleep 1; done
+# qcon into stp proc
+stpport=$(cat ${testpath}/process.csv | grep segmentedtickerplant | awk -F',' '{print $2}')
+tmux send-keys -t kdbProcess:bash -l qcon=\'/usr/bin/rlwrap\ /opt/kdb/qcon\'
+tmux send-keys -t kdbProcess:bash Enter
+tmux send-keys -t kdbProcess:bash stpport=\'$stpport\' Enter
+tmux send-keys -t kdbProcess:bash -l '$qcon homer.aquaq.co.uk:$stpport:admin:admin'
```
jonathonmcmurray, line 45: I think this should probably use `localhost` or something, otherwise I don't think this test will run on any machine other than homer

tests/stp/stripe/run.sh
```diff
@@ -0,0 +1,69 @@
...
+  #$debug $stop $write $quiet
+
+# k4unit test
+k4unit=$(find "$(cd ..; pwd)" -name "k4unit.q")
+# Performance test from stp
+cd $perfpath
+# Specify session name
+session=kdbProcess
+# Kill remaining session first
+tmux kill-session -t $session
```
jonathonmcmurray, line 35: is it typical in other TorQ tests to run things in a tmux (I haven't checked other tests)? not necessarily a problem, only issue is tmux isn't necessarily installed by default on all linux servers so it seems maybe a little strange to rely on it to run the tests

tests/stp/stripe/run.sh
```diff
@@ -0,0 +1,69 @@
...
+tmux send-keys -t kdbProcess:bash stpport=\'$stpport\' Enter
+tmux send-keys -t kdbProcess:bash -l '$qcon homer.aquaq.co.uk:$stpport:admin:admin'
+tmux send-keys -t kdbProcess:bash Enter
+# Load the script file
+tmux send-keys -t kdbProcess:bash \\l\ $k4unit Enter
+# Load settings
+settings=$testpath/settings.q
+tmux send-keys -t kdbProcess:bash \\l\ $settings Enter
+# Run performance test and save results
+tmux send-keys -t kdbProcess:bash \\l\ $perfpath/run_perf_test.q Enter
```
jonathonmcmurray, line 53: I'm a bit confused about all this stuff, why can't we just run the TorQ process with the `-test {whatever}` arg and let TorQ load k4unit etc. automatically?

tests/stp/stripe/stripe.csv
```diff
@@ -0,0 +1,45 @@
...
+run,0,0,q,row:100000,1,,"Rows to publish"
+run,0,0,q,"stpHandle @/: `.u.upd ,/: ((`trade;t row);(`quote;q row);(`heartbeat;`error))",1,,"Publish table updates"
+run,0,0,q,system "sleep 1",1,,"Wait for update to publish"
+run,0,0,q,![-11;] each exec logname from currlog where tbl in `trade`quote`segmentederrorlogfile,1,,"Replay logs locally"
+true,0,0,q,"all not raze value exec null logname,null handle from currlog",1,,"Check lognames and handles generated"
+true,0,0,q,currlog[`trade;`logname]~exec last logname from met where `trade in/: tbls,1,,"Check lognames the same in log and meta"
+true,0,0,q,currlog[`quote;`logname]~exec last logname from met where `quote in/: tbls,1,,"Check lognames the same in log and meta"
+true,0,0,q,(0#trade)~(exec first schema from met where `trade in/: tbls)`trade,1,,"Check trade schema saved to meta table"
+true,0,0,q,(0#quote)~(exec first schema from met where `quote in/: tbls)`quote,1,,"Check quote schema saved to meta table"
+true,0,0,q,"(row+sum[t1],sum q1)~sum(value rdbHandles) @\: ""count[trade],count quote""",1,,"Check trade and quote updates were correctly published"
```
jonathonmcmurray, line 26: If I'm reading this test right, it will check that a total of 100000 rows have been added to the tables across all the RDBs - this doesn't necessarily mean that the updates have been published correctly though, right?

e.g. I think we have 4 rdbs & syms should be roughly evenly distributed between them so we'd expect ~25000 rows added per rdb, but if all 100000 went to rdb1 and 0 to the other 3, I think this test would still pass?

Might be better to look at the distinct syms present in each rdb & check there's no overlap or something like that?

tests/stp/stripe/stripe.csv
```diff
@@ -0,0 +1,45 @@
...
+true,0,0,q,"(row+q2,t2)~count each (trade;quote)",1,,"Check updates were correctly saved to disk"
+run,0,0,q,rdbDown:`rdb1`rdb3,1,,"RDB(s) process to stop"
+run,0,0,q,"system""$HOME/TorQ/deploy/TorQ/latest/torq.sh stop "",("" ""sv string[(),rdbDown]),"" -csv $HOME/TorQ/deploy/TorQ/latest/tests/stp/stripe/process.csv""",1,,"Stop RDB(s) proccess"
```
jonathonmcmurray, line 29: Can we use `$TORQHOME` here instead of `$HOME`? (I think by the time these are running we must have `$TORQHOME` defined anyway). That way it is not dependent on a specific install location to run these tests

You can maybe even use `startorstopproc` from https://github.com/AquaQAnalytics/TorQ/blob/master/tests/helperfunctions.q here instead

tests/stp/stripe/stripe.csv
```diff
@@ -0,0 +1,45 @@
...
+run,0,0,q,"stpHandle @/: `.u.upd ,/: ((`trade;t row);(`quote;q row);(`heartbeat;`error))",1,,"Publish table updates while RDB(s) process is down"
+run,0,0,q,system "sleep 1",1,,"Wait for update to publish"
+run,0,0,q,"system""$HOME/TorQ/deploy/TorQ/latest/torq.sh start "",("" ""sv string[(),rdbDown]),"" -csv $HOME/TorQ/deploy/TorQ/latest/tests/stp/stripe/process.csv""",1,,"Restart RDB(s) proccess"
+run,0,0,q,rdbHandles:rdbNames!gethandle each rdbNames:exec procname from (.proc`readprocs).proc`file where proctype=`rdb,1,,"Reopen handle to RDBs"
+run,0,0,q,system "sleep 1",1,,"Wait for RDBs handles to open"
+run,0,0,q,"{(rdbHandles rdbDown) @\: x}each""-11!`"",/:string exec logname from currlog where tbl in `trade`quote",1,,"Replay logfiles for the restarted RDB(s) process"
+run,0,0,q,"(rdbHandles rdbDown) @\: "".rdb.applyfilters[;.rdb.subscribesyms]each .rdb.subtables""",1,,"Apply filters to the restarted RDB(s) process"
+true,0,0,q,"((2*row)+sum[t1],sum q1)~sum(value rdbHandles) @\: ""count[trade],count quote""",1,,"Check trade and quote data were replayed and filtered correctly"
```
jonathonmcmurray, line 37: same as above, probably better to do a check on distinct syms

jonathonmcmurray (commented):
This looks pretty good to me. A few minor style points on the code and a couple of thoughts about tests.

tests/stp/stripe/run.sh
mauricelimaquaq: I had to run the performance test on the stp and when I tried to run the test using the torq.q method, the rdbs are not able to connect to it before the test runs. I.e. the test runs on the stp port before the starting up the stp in debug mode. As a result, all the test runs and fail before it starts up in debug mode. Only then the connections to the rdbs are made. 

Which is why I had to rely on such an unorthodox method to run the performance test on the stp itself. Do you have any suggestion on how I can maybe change it?

tests/stp/stripe/run.sh
mauricelimaquaq: This is related to the problem above. Because the `-test` did not work as I expected, I had to use the run script to 

- qcon into the stp
- load the specific k4unit test files etc. 
- run the test

which is why I relied on tmux to "send the keys".

I tried to minimize the loading of the files by passing the k4unit.q, settings.q etc into the **load** column but it breaks the stp so I had to load it after it starts up.

tests/stp/stripe/stripe.csv
mauricelimaquaq: The additional check for updates have been published correctly is in line 38, i.e. there is no overlap of syms.

- `correct` checks that the sym updates went to the correct rdb 
- `notwrong` checks that there is not any sym update to the specific rdb that belongs to the other (wrong) rdbs

Line 39 iterates through a list of different rows of update, and checks everything is correct based on the accumation of the appropriate previous tests.
i.e. sum is correct, and syms are correct.

https://github.com/AquaQAnalytics/TorQ/blob/0d1e6a828fb1dccc625ca135324efdcfa2f18932/tests/stp/stripe/stripe.csv#L38-L39

Will work on the additional test for the distribution count

tests/stp/stripe/stripe.csv
mauricelimaquaq: Updated this in a643f77080175a4ad59ecbe41ccb070c4ef5e208

tests/stp/stripe/stripe.csv
mauricelimaquaq: Added the additional test for the distribution count in 08fa7c6e8289ff69a67d84bcf2b3a7c0dc0236b9

tests/stp/stripe/run.sh
jonathonmcmurray: Ah OK, I see why this is an issue... I'll have a think about it and see if I can think of any suggestions that could make this a bit simpler etc.

docs/stripe.md
```diff
@@ -0,0 +1,145 @@
...
+
+And with a series of symbols:
+
+```q
+q)f:{first string first md5 string x}each
+q)f`AMD`AIG`AAPL`DELL`DOW`GOOG`HPQ`INTC`IBM`MSFT
+"40809edfcb"
+```
+
+A hash map is based on the **`sym`** column is created like so:
```
jonathonmcmurray, line 38:
```suggestion
A hash map based on the **`sym`** column is created like so:
```

docs/stripe.md
```diff
@@ -0,0 +1,145 @@
...
+2| `SRPI`JGGQ`UJXC`TAFZ`DNBV
+c| `JKHN`PKBH`UOE
+f| `OZMV`KFKN`TDDC`ZSYF`ISAG`NJHL`KNKI`QSMN`VJYY
+6| `SRXA`DKKJ`FGXC`LGHO`OBCR`XXYB`TBKR
+d| `RYVW`SNKZ`PGBO`EPSV
+0| `QRXI`LOY`MADY
+9| `CWXD`KZJL`SYMB`EJDG
+```
+
+The hex keys will be divided across the number of striped processes, example 4 RDBs, like so:
```
jonathonmcmurray, line 62:
```suggestion
The hex keys will be divided across the number of striped processes, for example 4 RDBs, like so:
```

jonathonmcmurray (approved):
Few minor typos in docs, otherwise LGTM

code/common/pubsub.q
```diff
@@ -157,3 +157,41 @@ init:{[t]
...
+    hexdg!maxproc#til numseg
+    }[;16]
+
+// Hash function
+.ds.modmd5:{first each string first each md5'[string x]}
+
+.ds.map:{[numseg;sym] sym@/:group .ds.lookup[numseg] .ds.modmd5 sym}
+
+// Get the number of rdb processes from process.csv
+if[.ds.numseg=0;.ds.numseg:sum `rdb=((.proc`readprocs).proc`file)`proctype]
```
jonnypress, line 176: We shouldn't have any hardcoded requirements on process types or names. 

I'm not sure the best way to do this, but I don't think it needs to be full automatic as processes are added to the config file. I think it's fine to define in the publisher that the data is to be be split X ways, and that the subscriber has a config number that knows which consumer it is

docs/stripe.md
```diff
@@ -0,0 +1,145 @@
+# Data striping
+
```
jonnypress, line 2: I'm not sure where this page should go - I don't think it should be its own top level page. Maybe it should be a subsection in here: http://aquaqanalytics.github.io/TorQ/utilities/#pubsubq  ?

code/common/pubsub.q
mauricelimaquaq: I have removed this in 70c9c8b833392c818eba63956cc1a024196d4226, users can specify the number in the extras column of the process.csv

docs/stripe.md
mauricelimaquaq: I have moved the documentation to the subsection of pubsub.q in 70c9c8b833392c818eba63956cc1a024196d4226. However, I am not sure how the website renders the md text. Based on the existing md - it seems like the subsection starts with ### Heading level 3.

mauricelimaquaq: I have added 7a9279a9f7f78f24730fb10297176dfe95ad2d64 for your reference which uses the same format as the existing md (starting with ### Heading level 3). In terms of visuals on the website, I am not sure if the subheaders for the subsection will be too small for reading.

jonathonmcmurray: You can use `mkdocs build` & `mkdocs serve` to build & serve the docs and view how they will be rendered, see here: https://github.com/AquaQAnalytics/TorQ#updating-the-documentation-with-mkdocs

mauricelimaquaq: Great thanks, I have fixed the formatting and added a new paragraph on how general subscribers can make use of the hash function to stripe data.

jonnypress:
Hey @mauricelimaquaq this looks good to go, can you merge please?

PR #396, opened 2021-12-13
Fixed wdb EOD failure caused by syms containing special chars or lead…

cdempsey99:
…ing/trailing spaces.

When dealing with two or more syms which are identical apart from
 special chars or leading/trailing spaces, .wdb.upserttopartition
groups them all into the same sym subpartition in the wdbhdb. This
 can lead to a u-fail when the `p attr cannot be applied to the
 mixed syms. Now, however, .wdb.merge checks whether `p# can be
applied to the contents of each subpartition before they are saved
to the hdb. If not, the data is reordered by the subpartition
column (usually sym), which prevents the EOD failure.

### DIFF
--- a/code/processes/wdb.q
+++ b/code/processes/wdb.q
@@ -313,7 +313,15 @@ merge:{[dir;pt;tableinfo;mergelimits;hdbsettings]
 	      curr[0]:curr[0],select from get segment;
 	      curr[1]:curr[1],segment;		
       	$[islast or mergemaxrows < count curr[0];
-	        [.lg.o[`merge;"upserting ",(string count curr[0])," rows to ",string dest];
+	        
+                [pattrtest:@[{@[x;y;`p#]}[curr[0];];`sym;{x}];
+                if[10h=type pattrtest;
+                  /p attribute could not be applied, data must be re-sorted by subpartition col (sym):
+                  curr[0]: xasc[.merge.getextrapartitiontype[tablename];curr[0]];
+                  .lg.o[`resort;"The p attribute can now be applied"];
+                 ];
+                
+                .lg.o[`merge;"upserting ",(string count curr[0])," rows to ",string dest];
 	        dest upsert curr[0];
 	        .lg.o[`merge;"removing segments", (", " sv string curr[1])];
 	        .os.deldir each string curr[1];


code/processes/wdb.q
```diff
@@ -313,7 +313,15 @@ merge:{[dir;pt;tableinfo;mergelimits;hdbsettings]
 	      curr[0]:curr[0],select from get segment;
 	      curr[1]:curr[1],segment;		
       	$[islast or mergemaxrows < count curr[0];
-	        [.lg.o[`merge;"upserting ",(string count curr[0])," rows to ",string dest];
+	        
+                [pattrtest:@[{@[x;y;`p#]}[curr[0];];`sym;{x}];
```
jonnypress, line 317: I think it would be better if this test returned a true or false rather than a list or a string
Also, you should have log lines around this test condition as it will not take 0 time - so it would be good to know how long it takes. 
And you have `sym hardcoded in here, but I think it should be .merge.getextrapartitiontype[tablename]

code/processes/wdb.q
```diff
@@ -313,7 +313,15 @@ merge:{[dir;pt;tableinfo;mergelimits;hdbsettings]
...
+                if[10h=type pattrtest;
+                  /p attribute could not be applied, data must be re-sorted by subpartition col (sym):
+                  curr[0]: xasc[.merge.getextrapartitiontype[tablename];curr[0]];
```
jonnypress, line 320: Would be good to have a "re-sorting data" type log line before it does the sort as well

code/processes/wdb.q
cdempsey99: For the log lines around the test, do you just mean something like "Starting test" and "Test completed" for example?

jonnypress: Yes. You can probably get away with just the "starting test" line as it will be closely followed by either "the p attribute can now be applied" or "upserting.... "

code/processes/wdb.q
```diff
@@ -313,7 +313,17 @@ merge:{[dir;pt;tableinfo;mergelimits;hdbsettings]
 	      curr[0]:curr[0],select from get segment;
 	      curr[1]:curr[1],segment;		
       	$[islast or mergemaxrows < count curr[0];
-	        [.lg.o[`merge;"upserting ",(string count curr[0])," rows to ",string dest];
+	        
+                [.lg.o[`resort;"Checking that the contents of this subpartition conform"];
+                pattrtest:10h=type @[{@[x;y;`p#]}[curr[0];];.merge.getextrapartitiontype[tablename];{x}];
```
jonnypress, line 318: Can you just make this error trap return a 1b or 0b

cdempsey99: Sorry actually I'm not sure I see what you mean. If the error trap is only called when the function {@[x;y;`p#]} fails, shouldn't it just be a 1b in there? Otherwise just the successful table is returned

jonathonmcmurray: I think what Jonny is suggesting is make that lambda return e.g. 1b if it succeeds e.g. ``{@[x;y;`p#];1b}`` and then in the error trap return 0b if it fails, so you don't need the `10h=type` bit at the start, as you already have a boolean & don't need to do another check

jonnypress (approved):
Looks good

PR #400, opened 2021-12-22
Made STP upd function work for tables that have compound columns

darrenwsun:
If a table has compound columns like of string type, pushing a single row of data into segmented tickerplant in unchained mode gives a "length" error while inserting into the table. This is because the default implementation adds the current timestamp as the first element in the format of 1-size array, which may conflict with the length of the value for the compound column. 

To illustrate with an example, the below `.u.upd` call fails silently as no data is inserted into table `t` in either tp or rdb.
```
t:([] time:`timestamp$(); sym:`symbol$(); comment:());
.u.upd[`t;(`dummy;"test")];
```

### DIFF
--- a/code/processes/segmentedtickerplant.q
+++ b/code/processes/segmentedtickerplant.q
@@ -39,7 +39,7 @@ generateschemas:{
   // Table UPD functions attach the current timestamp by default, if STP is chained these do nothing
   $[.sctp.chainedtp;
     .stplg.updtab:(.stpps.t!(count .stpps.t)#{[x;y] x}),.stplg.updtab;
-    .stplg.updtab:(.stpps.t!(count .stpps.t)#{(enlist(count first x)#y),x}),.stplg.updtab
+    .stplg.updtab:(.stpps.t!(count .stpps.t)#{(enlist $[0>type first x;y;(count first x)#y]),x}),.stplg.updtab
     ]
   }
 
--- a/code/segmentedtickerplant/stplog.q
+++ b/code/segmentedtickerplant/stplog.q
@@ -50,7 +50,9 @@ upd:@[value;`.stplg.upd;enlist[`]!enlist ()];
 zts:@[value;`.stplg.zts;enlist[`]!enlist ()];
 
 // Functions to add columns on updates
-updtab:@[value;`.stplg.updtab;enlist[`]!enlist {(enlist(count first x)#y),x}]
+// The conditional is necessary when x is a single row of data and the table contains compound (e.g. string) columns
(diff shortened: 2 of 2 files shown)


jonnypress:
Hi Darren. Thanks for this - we always appreciate contributions. For this one though, we made a deliberate choice not to support this as standard (for performance reasons - remove the conditional logic on every update). We made a note of it in the backward compatibility section - every input must be list of lists: http://aquaqanalytics.github.io/TorQ/Processes/#backward-compatibility. 

So I don't think we should add this to the main codebase, though obviously you can customise it however you like in your installation by overriding or extending the definition of .stplg.updtab.

If I've misinterpreted the code change, please let me know. Thanks, Jonny

darrenwsun:
Thanks Jonny. It turned out that I didnt read the doc carefully :)

Appreciate that this is a deliberate decision for performance consideration, I'm curious how much the performance penalty is by the type checking (I could figure it out by doing benchmarking but wish someone that has experience can share). My understanding is that the main factor for throughout is the number of rows each time upd is called, and that this type checking plays a relatively minor role. When each update is sizable, data transmission, deserialization, and memory allocation are major time-taking parts; when each update is small, TCP slow start plays a bigger role. 

Basically I value completeness more than performance for this case: STP "should" support single row update* (despite not ideal for performance) for tables with compound columns, i.e. it shouldn't be a backward compatibility gap and instinct is better than doc. For sure I can overwrite it for my case, but I feel like this should be the default, and leave it to others to optimize where performance is really critial and removing this type check makes it worth (there are other steps done by upd that could be removed, this type checking might not be the first choice).

* Precisely STP supports this now, when it does not need to populate the first column as a timestamp itself. The type checking doesnt apply here.

darrenwsun:
@jonnypress Happy new year! Any comment on the response? Like to draw a conclusion on it, be it rejected or accepted.

jonnypress:
Hi @darrenwsun - happy new year to you too! 

W.r.t. the performance hit - agreed that it is very small in comparison to the other operations. But, we don't want to support this. When we were building this internally, we decided that we didn't want to by default support the optionality that kdb+ tick supports, specifically single row (non enlisted) inserts, and to optionally supply the time or not. For the vast majority of implementations we work on, data is inserted in batch (even if it's a 1 row "batch") and we always want the tickerplant to apply a timestamp. As mentioned in your commit, there's still an unhandled edge case where the first column is an nested list. We thought better to target the simplest case. Hope that makes sense.

darrenwsun:
Cool, thanks @jonnypress for the context.

PR #403, opened 2022-02-03
Update stripe function to improve distribution of syms

mauricelimaquaq:


jonathonmcmurray (approved):
LGTM

code/common/pubsub.q
```diff
@@ -161,27 +161,23 @@ init:{[t]
...
-            ];
-        ];
-        // Initialize .ds.subreq if does not exists
-        `.ds.subreq set .ds.map[.ds.numseg;sym]
+    // Check for new sym(s)
+    if[any new:not sym in key .ds.subreq;
+        // Append to .ds.subreq
+        newsubreq:.ds.subreq, .ds.map[.ds.numseg;sym where new];
+        // Apply unique attr on key of .ds.subreq
+        `.ds.subreq set (`u#key newsubreq)!value newsubreq;
```
jonnypress, line 180: I think you should just be able to do

.ds.subreq,:.ds.map[.ds.numseg;sym where new]

i.e. don't need to create the temp variarble and then assign it back

code/common/pubsub.q
```diff
@@ -161,27 +161,23 @@ init:{[t]
...
-.ds.map:{[numseg;sym] sym@/:group .ds.modmd5[sym]mod numseg}
+.ds.map:{[numseg;sym] sym!.ds.modmd5[sym]mod numseg};
 
+// Initialise subscription request on startup
+.ds.subreq:()!();
 // Striping function which stores the mappings for any symbols that it has already computed and 
 // for subsequent requests for that symbol, it looks them up
 .ds.stripe:{[input;skey]
     // If no updates, return
     if[0=count sym:distinct input;:()];
```
jonnypress, line 174: I think you should move the distinct to the next if clause. 
If the if condition here causes the function to return (), then the distinct wasn't required.

mauricelimaquaq:
@jonnypress I have made the necessary changes and updated all the unit-test files to work with the new function. The unit and performance tests are working and passed. Let me know if it is ok.

code/common/pubsub.q
```diff
@@ -160,36 +160,22 @@ init:{[t]
...
-                `.ds.subreq set .ds.subreq,' .ds.map[.ds.numseg;sym where new];
-            ];
-        ];
-        // Initialize .ds.subreq if does not exists
-        `.ds.subreq set .ds.map[.ds.numseg;sym]
-    ];
-    .ds.subreq skey
-    }
+  // If no updates, return
+  if[0=count input;:()];
```
jonnypress, line 172: Probably better if this returns a boolean list rather than a generic empty

docs/utilities.md
```diff
@@ -721,13 +721,13 @@ The advantages of striping include performance and throughput improvements. Segm
...
 
 A common method for data striping between processes is to use an instrument (sym) filter. For example, to stripe data across 2 RDB processes, data with symbols starting with A-M and N-L will be striped to RDB1 and RDB2 respectively. However, a major problem with this method is the uneven distribution of data (a lot of symbols tend to start with A for example).
 
-#### Data hash striping with MD5
+**Data hash striping with MD5**
 
 A way to get around this problem is to stripe the data using a hash value which allows for better distribution. The hash function will store the mappings for the symbols that it has already computed and for subsequent requests for those symbols, it looks them up. It is loaded into the segmented tickerplant to use as subscription requests. For this purpose, MD5 (Message-Digest algorithm 5) hash is chosen as it is a fast and [built-in](https://code.kx.com/q/ref/md5/) hash function in kdb+. It creates a hexadecimal byte array from an input string and the first hex value from the output byte array is used to create a hash map for the data hash striping.
```
jonnypress, line 732: Need to update this bit of documentation

tests/stp/stripe/process.csv
```diff
@@ -1,8 +1,8 @@
...
 localhost,{KDBBASEPORT}+101,segmentedtickerplant,stpperf,${TORQHOME}/appconfig/passwords/accesslist.txt,1,0,,,${KDBCODE}/processes/segmentedtickerplant.q ${perfpath}/prep_test.q,0,-schemafile ${TORQHOME}/database.q,q
-localhost,{KDBBASEPORT}+102,rdb,rdb1,${TORQHOME}/appconfig/passwords/accesslist.txt,1,1,180,,${KDBCODE}/processes/rdb.q,1,-.rdb.replaylog 0b -.rdb.subfiltered 1,q
-localhost,{KDBBASEPORT}+103,rdb,rdb2,${TORQHOME}/appconfig/passwords/accesslist.txt,1,1,180,,${KDBCODE}/processes/rdb.q,1,-.rdb.replaylog 0b -.rdb.subfiltered 1,q
-localhost,{KDBBASEPORT}+104,rdb,rdb3,${TORQHOME}/appconfig/passwords/accesslist.txt,1,1,180,,${KDBCODE}/processes/rdb.q,1,-.rdb.replaylog 0b -.rdb.subfiltered 1,q
-localhost,{KDBBASEPORT}+105,rdb,rdb4,${TORQHOME}/appconfig/passwords/accesslist.txt,1,1,180,,${KDBCODE}/processes/rdb.q,1,-.rdb.replaylog 0b -.rdb.subfiltered 1,q
\ No newline at end of file
+localhost,{KDBBASEPORT}+102,rdb,rdb1,${TORQHOME}/appconfig/passwords/accesslist.txt,1,1,180,,${KDBCODE}/processes/rdb.q,1,-.rdb.replaylog 0b -.rdb.subfiltered 1 -.ds.numseg 4,q
```
jonnypress, line 5: I don't fully understand this start line - why do the RDBs need -.ds.numseg 4? 
Would they not just need to know which segment they are subscribing to? i.e. the first one needs to subscribe to segment 0, next to segment 1 etc.

docs/utilities.md
```diff
@@ -863,22 +863,22 @@ localhost,{KDBBASEPORT}+5,rdb,rdb3,${TORQAPPHOME}/appconfig/passwords/accesslist
...
-##### {subscriber}.q
+**{subscriber}.q**
 
 Add the conditions required (using the data hash striping function).
 
 ```q
 ...
 // Subscribe to all tables but with custom conditions
-conditions:1!flip`tabname`filters`columns!(`trade`quote;("sym in .ds.stripe[sym;0]";"sym in .ds.stripe[sym;1]");("";""))
+conditions:1!flip`tabname`filters`columns!(`trade`quote;(".ds.stripe[sym;0]";".ds.stripe[sym;1]");("";""))
```
jonnypress, line 881: Is this definitely a correct example? I would've thought the common case was for a subscriber to subscribe to all tables for the same segment, rather than trade=segment 0, quote=segment 1

tests/stp/stripe/process.csv
mauricelimaquaq: This is for testing purposes only. In the normal setup it is not required.
In the test, .rdb.replaylog is disabled, so during the testing when a rdb restarts, it manually replays the log and invoke the .rdb.applyfilters[;.rdb.subscribesyms] which requires a .ds.numseg variable (as .rdb.subscribesyms uses .ds.stripe to apply the filter) on the rdb.

Specifically in this line of code:
[run,0,0,q,"(rdbHandles rdbDown) @\: "".rdb.applyfilters[;.rdb.subscribesyms]each .rdb.subtables""",1,,"Apply filters to the restarted RDB(s) process"](https://github.com/AquaQAnalytics/TorQ/blob/385f8628541b9d096d98322c8f6c1cd82266aa76/tests/stp/stripe/stripe.csv#L39)

docs/utilities.md
mauricelimaquaq: Yes this is a correct example. I was trying to show that it is possible to have custom conditions where a single general subscriber has the flexibility to subscribe to different tables in different segments.

jonnypress (approved):
Think this looks good

PR #410, opened 2022-02-17
Tcs 20 add segmented subscriptions filter and only calculate symbol segments for new symbols when mapping

AdamBrown2:
Adding segmented subscription


rian-morgan (approved):

rian-morgan (changes requested):
please move this to segmentedtickerplant/datastripe.q

jwrno1 (changes requested):

code/segmentedtickerplant/subsegement.q
```diff
@@ -0,0 +1,24 @@
+\d .stpps
+
+// Function to extract where clause from config table this need to be finished
+segmentfilter:{[tbl;segid]
+  segmentfilterfunc:{[tbl;segid] string first .stpps.segmentfiltermap[exec wcRef from .stpps.segmentconfig where table=tbl , segmentID=segid]};
```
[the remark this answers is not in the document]
AdamBrown2: I ran into and issue when placing it inside .[] where the function would always return "" defining inside the function was the first way I found for it to work. So i will define it outside of the function and test that.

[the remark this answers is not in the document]
rian-morgan: segmentfilterfunc lets get out of the habit of appending func to things, this needs changed

code/segmentedtickerplant/datastripe.q
```diff
@@ -22,3 +22,29 @@ initdatastripe:{
...
+  // Function to subscribe to particular segment using segmentID based on .u.sub
+subsegment:{[tbl;segid];
+//tablename and segmentid used to get filters
+//filteroutput needs to be tested
+  if[tbl~`;:.z.s[;segid] each .stpps.t];
+   if[not tbl in .stpps.t;
+     .lg.e[`sub;m:"Table ",string[tbl]," not in list of stp pub/sub tables"];
+     :(tbl;m)
+  ];
+  $[segid~`;[filter:`]; [filter: segmentfilter[tbl;segid]]];
```
rian-morgan, line 46: segid is not a symbol, its an int. better to use `null segid` for this conditional. Its more general.

AdamBrown2:
```
  filter:$[null segid; `;segmentfilter[tbl;segid]];
  [filter~`;.stpps.suball[tbl]; .ps.subtablefiltered[string[tbl];filter;""]]`
```
Made the changes requested but when you use .stpps.subsegment[`quote;`] it gives a type error. `segid ~ `  would allow the function to act .u.sub[`quote;`] where you sub to quote with no filter. At the moment the only way I think you can sub to a table without a filter is to trigger the error trap in `.stpps.segmentfilter[quote;quote]`.

AdamBrown2:
Added all the changes requested and its nearly ready. The same issue occurred when segmenting.csv and filtermap.csv dont exist.
```
q).stpps.segmentfilter[`trade;1]

q)type ,stpps.segmentfilter[`trade;1]
0h
```
when the csv's exist but are empty this isnt a problem the function return ""
This doesn't affect .stpps.subsegment but could cause issues for other processes.

code/segmentedtickerplant/datastripe.q
```diff
@@ -22,3 +22,29 @@ initdatastripe:{
...
+\d .stpps
+//function to grab where clause from config tables
+segmentfilterfunc:{[tbl;segid] 
+  string first .stpps.segmentfiltermap[exec wcRef from .stpps.segmentconfig where table=tbl , segmentID=segid]};
+// Function to use segmentfilterfunc and return "" incase of error
+segmentfilter:{[tbl;segid]
+  output:.[segmentfilterfunc;
+  (tbl;segid) ;
+  "" ];
+  $[count output>0;output;""]};
```
jonathonmcmurray, line 35: what's the point of this check? seems to me like it wouldn't change the output?

also not sure that the condition is doing what might be expected, assume it should be `count[output]>0` (or `0<count output` to avoid square brackets)

code/segmentedtickerplant/datastripe.q
```diff
@@ -22,3 +22,29 @@ initdatastripe:{
...
+
+\d .stpps
+//function to grab where clause from config tables
+segmentfilterfunc:{[tbl;segid] 
+  string first .stpps.segmentfiltermap[exec wcRef from .stpps.segmentconfig where table=tbl , segmentID=segid]};
+// Function to use segmentfilterfunc and return "" incase of error
+segmentfilter:{[tbl;segid]
+  output:.[segmentfilterfunc;
+  (tbl;segid) ;
+  "" ];
```
jonathonmcmurray, line 34:
```suggestion
  output:.[segmentfilterfunc;(tbl;segid);""];
```

think this is short enough that it can/should be a single line

AdamBrown2: Thanks, I have now made these changes

AdamBrown2:
I think I made a mistake in the process of pushing this and github is showing the changes made as if they where mine. My changes are lines 53 to 79.

rian-morgan (changes requested):

rian-morgan (approved):

code/segmentedtickerplant/datastripe.q
```diff
@@ -48,4 +48,32 @@ segmentedsubdetails: {[tabs;instruments;id] (!). flip 2 cut (
...
+subsegment:{[tbl;segid];
+//tablename and segmentid used to get filters
+  if[tbl~`;:.z.s[;segid] each .stpps.t];
+   if[not tbl in .stpps.t;
+     .lg.e[`sub;m:"Table ",string[tbl]," not in list of stp pub/sub tables"];
+     :(tbl;m)
+  ];
+  filter:segmentfilter[tbl;segid];
+  if[filter~"";
+    .lg.o[`sub;m:"Incorrect pairing of table ",string[tbl]," and segmentID ",string[segid]," not found in .stpps.segmentconfig"];
```
rian-morgan, line 70: can we change this to .lg.e, to match original .u.sub functionality. .proc.stop=0b on stp so wont kill process.

rian-morgan (approved):

code/segmentedtickerplant/datastripe.q
```diff
@@ -48,4 +48,32 @@ segmentedsubdetails: {[tabs;instruments;id] (!). flip 2 cut (
...
+    .lg.e[`sub;m:"Incorrect pairing of table ",string[tbl]," and segmentID ",string[segid]," not found in .stpps.segmentconfig"];
+    :((tbl;segid);m)
+  ];
+ .ps.subtablefiltered[string[tbl];filter;""]
+ };
+
+\d .
+
 if[.ds.datastripe;.proc.addinitlist[(`initdatastripe;`)]];
+
```
MichaelTodorov, line 79: could we move the filtermap after the subsegment code and the segmentedsubdetails just before the final init to avoid jumping back and forth between namespaces?

MichaelTodorov (approved):

PR #411, opened 2022-02-21
TCS-31

eoinnagle:
added error trapping to configload function and added empty table/mapping for use in reverting to default mode

### DIFF
--- a/code/segmentedtickerplant/datastripe.q
+++ b/code/segmentedtickerplant/datastripe.q
@@ -9,5 +9,11 @@
 configload:{
      .stpps.segmentconfig:("SIS";enlist",")0: hsym first .proc.getconfigfile["segmenting.csv"];
      .stpps.segmentfiltermap:(!/) (("SS";enlist",")0: hsym first .proc.getconfigfile["filtermap.csv"])`wcRef`filter;
-     }
+     };
 
+initdatastripe:{
+     .lg.o[`init;"init datastriping"];
+     configload[];
+     };
+
+if[.ds.datastripe;.proc.addinitlist[(`initdatastripe;`)]];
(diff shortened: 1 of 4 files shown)


code/segmentedtickerplant/datastripe.q
```diff
@@ -3,12 +3,17 @@
...
+//creating empty versions of segmenting table and filter mapping so we can revert to default mode if issues with either csv
+
+.stpps.segmentconfig:([] table:();segmentID:();wcRef:());
+.stpps.segmentfiltermap:(!/)([] wcRef:();filter:())`wcRef`filter;
 
 configload:{
-     .stpps.segmentconfig:("SIS";enlist",")0: hsym first .proc.getconfigfile["segmenting.csv"];
-     .stpps.segmentfiltermap:(!/) (("SS";enlist",")0: hsym first .proc.getconfigfile["filtermap.csv"])`wcRef`filter;
+     @[{.stpps.segmentconfig:{("SIS";enlist",")0: hsym first .proc.getconfigfile[x]}[x]};"segmenting.csv";{.lg.o[`stp1;"Failed to load segmenting.csv, revert to defaut mode and no segmenting."]}];
+     @[{.stpps.segmentfiltermap:{(!/)(("SS";enlist",")0: hsym first .proc.getconfigfile[x])`wcRef`filter}[x]};"filtermap.csv";{.lg.o[`stp1;"Failed to load filtermap.csv, revert to defaut mode and no segmenting."]}];
```
jwrno1, line 16: Will need to change segmenting.csv and filtermap.csv to variables at somepoint. I'll approve for now so that adam can test it with his changes

jwrno1 (approved):

code/segmentedtickerplant/datastripe.q
rian-morgan: like josh said, these nfilename need to be variables, read in from command line. Also the logging shows that system will revert to default mode, but there is no logic to do this. I suppose this can be added in later but it will need to be noted.

The logging should use .lg.e instead of .lg.o, since its an error here. Also remove the reference to stp1, the proc name shouldnt be hard coded in here. What if this isnt the process name?

rian-morgan (changes requested):
The logging messages need work

PR #412, opened 2022-02-21
TCS-31  added error trapping to configload function and added empty table/mapping for use in reverting to default mode

eoinnagle:
noted Josh and Rian's comments on previous PR

-doing PR so that changes can be merged and Adam can test if his function will revert the TP to default mode when empty table or mapping is passed
-using .lg.o at this stage as when .lg.e is passed when starting a proc the session is exited, can revisit but kept at this for time being for Adams tests
-reference to stp1 has been removed

### DIFF
--- a/code/segmentedtickerplant/datastripe.q
+++ b/code/segmentedtickerplant/datastripe.q
@@ -9,5 +9,11 @@
 configload:{
      .stpps.segmentconfig:("SIS";enlist",")0: hsym first .proc.getconfigfile["segmenting.csv"];
      .stpps.segmentfiltermap:(!/) (("SS";enlist",")0: hsym first .proc.getconfigfile["filtermap.csv"])`wcRef`filter;
-     }
+     };
 
+initdatastripe:{
+     .lg.o[`init;"init datastriping"];
+     configload[];
+     };
+
+if[.ds.datastripe;.proc.addinitlist[(`initdatastripe;`)]];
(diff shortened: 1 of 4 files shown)


code/segmentedtickerplant/datastripe.q
```diff
@@ -3,12 +3,17 @@
...
 
-//function transfoms segmenting.csv into table format so it can be accessed
-//function transforms filtermap.csv into table format then into a mapping of wcRef to filter which can be accessed and applied to data
+//configload transfoms segmenting.csv into table format so it can be accessed
+//configload transforms filtermap.csv into table format then into a mapping of wcRef to filter which can be accessed and applied to data
+
+//creating empty versions of segmenting table and filter mapping so we can revert to default mode if issues with either csv
+
+.stpps.segmentconfig:([] table:();segmentID:();wcRef:());
+.stpps.segmentfiltermap:(!/)([] wcRef:();filter:())`wcRef`filter;
```
jonathonmcmurray, line 12:
```suggestion
.stpps.segmentfiltermap:exec wcRef!filter from ([] wcRef:();filter:());
```

IMO this is clearer - though for an empty version could also just define an empty dict, would give the same result:

```suggestion
.stpps.segmentfiltermap:()!();
```

rian-morgan (approved):
ok looks better. This will still need more work but can be merged now.

jwrno1 (approved):

PR #413, opened 2022-02-22
wdb reloadproc improvement 

### LINKED ISSUE #406: wdb reloadproc improvement
Currently when .wdb.eodwaittime>0, the async callback reload to the underlying processes does not log error should the reload failed and continue silently. This creates a blind spot for reload issues of the underlying processes.

Suggestion is to log the error during try catch. A summary of all failed reload on wdb will be good also.

eointeague:
Catch and log error if async callback reload fails (when .wdb.eodwaittime>0).
Once .wdb.reloadproc has been called for each process add summary logging to log the number of processes which have successfully reloaded.

### DIFF
--- a/code/processes/wdb.q
+++ b/code/processes/wdb.q
@@ -390,7 +390,8 @@ reloadproc:{[h;d;ptype]
         syncreloadfunc:{[h;d;ptype] r:@[h;(`reload;d);{[ptype;e] .lg.e[`reloadproc;"failed to reload the ",string[ptype],". The error was : ",e];e}[ptype]];
         .lg.o[`reloadproc;"the ", string[ptype]," ", $[10h~type r; "failed to reload with error ",r;"successfully reloaded"]]}; 
         $[eodwaittime>0;
-                 sendfunc[(reloadfunc;d;ptype);h;ptype];     
+                 (.lg.o[`reloadproc;"reload call has been sent to ", string[ptype]];
+                         sendfunc[(reloadfunc;d;ptype);h;ptype]);     
 		 syncreloadfunc[h;d;ptype]
         ];
         }
--- a/code/processes/wdb.q
+++ b/code/processes/wdb.q
@@ -208,7 +208,7 @@ handler:{
 	/-insert process reload outcome into .wdb.reloadsummary 
                  .wdb.reloadsummary[.z.w]:x;
         /-log result of reload in wdb out log 
-                .lg.o[`reload;"the ", string[.wdb.reloadsummary[.z.w]`process]," process ", string[.wdb.reloadsummary[.z.w]`result]];
+                .lg.o[`reloadproc;"the ", string[.wdb.reloadsummary[.z.w]`process]," process ", string[.wdb.reloadsummary[.z.w]`result]];
         if[(.proc.cp[]>.wdb.timeouttime) or (count[.wdb.reloadsummary]=.wdb.countreload);
                 .lg.o[`handler;"releasing processes"];
                 .lg.o[`reload;string[count select from .wdb.reloadsummary where status=1]," out of ", string[count .wdb.reloadsummary]," processes successfully reloaded"];
@@ -386,9 +386,12 @@ reloadproc:{[h;d;ptype]
         /-and call .wdb.handler with failed reload message. If reload is successful call .wdb.handler with successful reload message.
         reloadfunc:{[d;ptype] r:@[`. `reload;d;{.lg.e[`reloadproc;"failed to reload from .wdb.reloadproc call. The error was : ",x];x}];
         (neg .z.w)(`.wdb.handler;(ptype;10h<>type r;$[10h~type r;`$"reload failed with error ",r;`$"reloaded successfully"]));(neg .z.w)[]};
+        /-reload function to be executed if eodwaitime = 0 - sync message processes to reload and log if reload was successful or failed
+        syncreloadfunc:{[h;d;ptype] r:@[h;(`reload;d);{[ptype;e] .lg.e[`reloadproc;"failed to reload the ",string[ptype],". The error was : ",e];e}[ptype]];
+        .lg.o[`reloadproc;"the ", string[ptype]," ", $[10h~type r; "failed to reload with error ",r;"successfully reloaded"]]}; 
         $[eodwaittime>0;
                  sendfunc[(reloadfunc;d;ptype);h;ptype];     
-		 @[h;(`reload;d);{[ptype;e] .lg.e[`reloadproc;"failed to reload the ",string[ptype],".  The error was : ",e]}[ptype]]
+		 syncreloadfunc[h;d;ptype]
         ];
         }
 
--- a/code/processes/wdb.q
+++ b/code/processes/wdb.q
@@ -384,11 +384,11 @@ reloadproc:{[h;d;ptype]
         sendfunc:{[x;y;ptype].[{neg[y]@x};(x;y);{[ptype;x].lg.e[`reloadproc;"failed to reload the ",string[ptype]];'x}[ptype]]};
         /-reload function sent to processes by sendfunc in order to call process to reload. If process fail to reload log error 
         /-and call .wdb.handler with failed reload message. If reload is successful call .wdb.handler with successful reload message.
-        reloadfunc:{[d;ptype] r:@[`. `reload;d;{x}];
+        reloadfunc:{[d;ptype] r:@[`. `reload;d;{.lg.e[`reloadproc;"failed to reload from .wdb.reloadproc call. The error was : ",x];x}];
 	$[10h = type r;
-                 ((neg .z.w)(`.wdb.handler;(ptype;0b;`$"reload failed with error ",r));(neg .z.w)[];
-                          .lg.e[`reloadproc;"failed to reload ",string[ptype]," from .wdb.reloadproc call. The error was : ",r]);
-                 ((neg .z.w)(`.wdb.handler;(ptype;1b;`$"reloaded successfully"))(neg .z.w)[])]};
+                 (reloadstatus:(ptype;0b;`$"reload failed with error ",r));
+                 (reloadstatus:(ptype;1b;`$"reloaded successfully"))];
+                 (neg .z.w)(`.wdb.handler;reloadstatus);(neg .z.w)[]};
         $[eodwaittime>0;
                  sendfunc[(reloadfunc;d;ptype);h;ptype];     
 		 @[h;(`reload;d);{[ptype;e] .lg.e[`reloadproc;"failed to reload the ",string[ptype],".  The error was : ",e]}[ptype]]
--- a/code/processes/wdb.q
+++ b/code/processes/wdb.q
@@ -206,17 +206,16 @@ endofdaysave:{[dir;pt]
 /- add entries to dictionary of callbacks. if timeout has expired or d now contains all expected rows then it releases each waiting process
 handler:{
 	/-insert process reload outcome into .wdb.reloadstatus 
-        if[not .z.w in  key .wdb.reloadstatus;
-                .wdb.reloadstatus[.z.w]:x];
-        if[(not .wdb.reloadstatus[.z.w]`status);
-                .lg.o[`reload;"the ", string[.wdb.reloadstatus[.z.w]`process]," process reload has ", string[.wdb.reloadstatus[.z.w]`result]]];
+                .wdb.reloadstatus[.z.w]:.wdb.reloadsum;
+        /-log result of reload in wdb out log 
+                .lg.o[`reload;"the ", string[.wdb.reloadstatus[.z.w]`process]," process ", string[.wdb.reloadstatus[.z.w]`result]];
         if[(.proc.cp[]>.wdb.timeouttime) or (count[.wdb.reloadstatus]=.wdb.countreload);
                 .lg.o[`handler;"releasing processes"];
                 .lg.o[`reload;string[count select from .wdb.reloadstatus where status=1]," out of ", string[count .wdb.reloadstatus]," processes successfully reloaded"];
-                .wdb.flushend[]];
         /-delete contents from .wdb.reloadstatus when reloads completed
-        if[.wdb.reloadcomplete;delete from `.wdb.reloadstatus];
-	};
+                .wdb.flushend[];
+                delete from `.wdb.reloadstatus];
+       	};
 
 /- evaluate contents of d dictionary asynchronously
 /- notify the gateway that we are done
@@ -381,16 +380,23 @@ endofdaysort:{[dir;pt;tablist;writedownmode;mergelimits;hdbsettings]
 
 /-function to send reload message to rdbs/hdbs
 reloadproc:{[h;d;ptype]
+        /-default process message to be inserted into reload status table (.wdb.reloadstatus) will be changed if reload fails
+        .wdb.reloadsum:(ptype,1b,`$"successfully reloaded");
+        /-count of processes to be reloaded 
         .wdb.countreload:count[raze .servers.getservers[`proctype;;()!();1b;0b]each reloadorder];
+        /-defining lambdas to be executed in conditional statement 
+        /-async call back function executed when eodwaittime>0
+        sendfunc:{[x;y;ptype].[{neg[y]@x};(x;y);{[ptype;x].lg.e[`reloadproc;"failed to reload the ",string[ptype]];'x}[ptype]]};
+        /-reload function sent to processes by sendfunc in order to call process to reload. Calls process to reload, if process fails to reload
+        /-execute lambda to set .wdb.reloadsum to an error message. Then message wdb to execute wdb.handler function.
+        reloadfunc:{[d;ptype]@[`. `reload;d;
+        /-lambda to execute if reload fails, sets .wdb.reloadsum to error message and logs error in process error log
+                         {[ptype;e](neg .z.w)(set;`.wdb.reloadsum;(ptype;0b;`$"reload failed with error: ",e));
+                         .lg.e[`reloadproc;"failed to reload ",string[ptype]," from .wdb.reloadproc call. The error was : ",e]}[ptype]];
+        /-send message to wdb to execute .wdb.handler function
+                         (neg .z.w)(`.wdb.handler;`); (neg .z.w)[]};
         $[eodwaittime>0;
-                {[x;y;ptype].[{neg[y]@x};(x;y);{[ptype;x].lg.e[`reloadproc;"failed to reload the ",string[ptype]];'x}[ptype]]}
-                [({@[`. `reload;x;
-	/-error trap lambda to execute if process fails to reload error added to .wdb.reloadstatus and logged in process error log
-                        {[ptype;e](neg .z.w)(`.wdb.handler;(ptype;0b;`$"failed with error: ",e));
-                        .lg.e[`reloadproc;"failed to reload ",string[ptype]," from .wdb.reloadproc call. The error was : ",e]}[y]];
-        /-successful reload message to be sent to handler and entered into .wdb.reloadstatus for end of reload summary logging
-        /-this message will alway be sent to wdb however if process has failed .wdb.handler will not log this successful reload 
-	                 (neg .z.w)(`.wdb.handler;(y;1b;`$"successfully reloaded")); (neg .z.w)[]};d;ptype);h;ptype];       
+                 sendfunc[(reloadfunc;d;ptype);h;ptype];       
 		 @[h;(`reload;d);{[ptype;e] .lg.e[`reloadproc;"failed to reload the ",string[ptype],".  The error was : ",e]}[ptype]]
         ];
         .lg.o[`reload;string[ptype]," reload has finished"];
--- a/code/processes/wdb.q
+++ b/code/processes/wdb.q
@@ -67,8 +67,6 @@ gc:@[value;`gc;1b];                                                        /-gar
 
 eodwaittime:@[value;`eodwaittime;0D00:00:10.000];                          /-length of time to wait for async callbacks to complete at eod
 
-/failedreloadsummary:@[value;`failedreloadsummary;([id:()]time:();process:();error:())];
-reloadfailed:0b;
 
 / - settings for the common save code (see code/common/save.q)
 .save.savedownmanipulation:@[value;`savedownmanipulation;()!()];           /-a dict of table!function used to manipulate tables at EOD save
@@ -207,18 +205,13 @@ endofdaysave:{[dir;pt]
 
 /- add entries to keyed table of callbacks. if timeout has expired or d now contains all expected rows then it releases each waiting process
 handler:{
-        `.wdb.d insert .z.w,x;
-	$[(last .wdb.d)[`status];
-	.lg.o[`reload;"the ",string[(last .wdb.d)`process]," has been successfully reloaded"];
-	.lg.o[`reload;"the ",string[(last .wdb.d)`process]," reload has failed check ",string[(last .wdb.d[])`process]," error log"]];
+	 if[not .z.w in  key .wdb.d;
+        .wdb.d[.z.w]:x];
 	if[(.proc.cp[]>.wdb.timeouttime) or (.wdb.countreload=count .wdb.d);
-		$[(count select from .wdb.d  where status=1)=count .wdb.d;
-        	.lg.o[`reload;"all processes successfully reloaded"];
-        	.lg.o[`reload;string[count select from .wdb.d where status=1]," out of ", string[count .wdb.d]," processes successfully reloaded"]];
+        	.lg.o[`reload;string[count select from .wdb.d where status=1]," out of ", string[count .wdb.d]," processes successfully reloaded"];
         	.lg.o[`handler;"releasing processes"];
-		.lg.o[`handler;"releasing processes"];
-		.wdb.flushend[];
-		.wdb.d:([id:()]time:();process:();status:();result:())];
+		.wdb.flushend[]];
+		if[.wdb.reloadcomplete;.wdb.d:([handle:()]process:();status:();result:())];
 	};
 
 /- evaluate contents of d dictionary asynchronously
@@ -232,7 +225,10 @@ flushend:{
 	};
 
 /- initialise d
-d:([id:()]time:();process:();status:();result:());
+d:([handle:()]process:();status:();result:());
+
+/-initialise reload complete
+reloadcomplete:0b;
 
 doreload:{[pt]
 	.wdb.reloadcomplete:0b;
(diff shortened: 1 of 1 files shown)


code/processes/wdb.q
```diff
@@ -203,27 +203,32 @@ endofdaysave:{[dir;pt]
...
-/- add entries to dictionary of callbacks. if timeout has expired or d now contains all expected rows then it releases each waiting process
+/- add entries to keyed table of callbacks. if timeout has expired or d now contains all expected rows then it releases each waiting process
 handler:{
-	.wdb.d[.z.w]:x;
-	if[(.proc.cp[]>.wdb.timeouttime) or (count[.wdb.d]=.wdb.countreload);
-		.lg.o[`handler;"releasing processes"];
-		.wdb.flushend[];
-		.wdb.d:()!()];
+	 if[not .z.w in  key .wdb.d;
+        .wdb.d[.z.w]:x];
```
jonnypress, line 209: indentation

code/processes/wdb.q
```diff
@@ -203,27 +203,32 @@ endofdaysave:{[dir;pt]
...
+	if[(.proc.cp[]>.wdb.timeouttime) or (.wdb.countreload=count .wdb.d);
+        	.lg.o[`reload;string[count select from .wdb.d where status=1]," out of ", string[count .wdb.d]," processes successfully reloaded"];
```
jonnypress, line 211: Would be nice to print the errors of those that failed here for completeness

code/processes/wdb.q
```diff
@@ -203,27 +203,32 @@ endofdaysave:{[dir;pt]
...
-	 @[{neg[x]"";neg[x][]};;()] each key d;
+	 @[{neg[x]"";neg[x][]};;()] each value flip key d;
 	 informgateway(`reloadend;`);
 	 .lg.o[`sort;"end of day sort is now complete"];
 	 .wdb.reloadcomplete:1b];
 	};
 
 /- initialise d
-d:()!()
+d:([handle:()]process:();status:();result:());
```
jonnypress, line 228: Add types to the columns of this table

code/processes/wdb.q
```diff
@@ -203,27 +203,32 @@ endofdaysave:{[dir;pt]
...
-		.lg.o[`handler;"releasing processes"];
-		.wdb.flushend[];
-		.wdb.d:()!()];
+	 if[not .z.w in  key .wdb.d;
+        .wdb.d[.z.w]:x];
+	if[(.proc.cp[]>.wdb.timeouttime) or (.wdb.countreload=count .wdb.d);
+        	.lg.o[`reload;string[count select from .wdb.d where status=1]," out of ", string[count .wdb.d]," processes successfully reloaded"];
+        	.lg.o[`handler;"releasing processes"];
+		.wdb.flushend[]];
+		if[.wdb.reloadcomplete;.wdb.d:([handle:()]process:();status:();result:())];
```
jonnypress, line 214: probably better to do a "delete from" as it means if the structure of the table were to change you only have to change it in one place

code/processes/wdb.q
```diff
@@ -252,9 +257,9 @@ movetohdb:{[dw;hw;pt]
       not any a[dw]in(a:{key hsym`$x}) hw;
       [{[y;x] 
         $[not(b:`$last"/"vs x)in key y;
-          [.[.os.ren;(x;y);{[x;y;e].lg.e[`mvtohdb;"Table ",string[x]," has failed to copy to ",string[y]," with error: ",e]}[b;y;]];
-           .lg.o[`mvtohdb;"Table ",string[b]," has been successfully moved to ",string[y]]];
-          .lg.e[`mvtohdb;"Table ",string[b]," was skipped because it already exists in ",string[y]]];
+          [.[.os.ren;(x;y);{[x;y;e].lg.e[`mvtohdb;"Table ",string[x]," has failed to copy to ",y," with error: ",e]}[b;y;]];
```
jonnypress, line 260: should these changes be in here?

jonnypress (commented):
I think you've lost a lot of previous changes that we would like to keep. I'm not sure how that has happened - but can you fix it?
You need to review the PR after creating it to ensure that nothing unexpected has crept in

code/processes/wdb.q
```diff
@@ -38,7 +38,7 @@ tpconnsleepintv:@[value;`tpconnsleepintv;10];                              /-num
 tpcheckcycles:@[value;`tpcheckcycles;0W];                                  /-number of attempts to connect to tp before process is killed 
 
 sorttypes:@[value;`sorttypes;`sort];                                       /-list of sort types to look for upon a sort		
-sortworkertypes:@[value;`sortworkertypes;`sortworker];                     /-list of sort types to look for upon a sort being called with worker process
+sortslavetypes:@[value;`sortslavetypes;`sortslave];                        /-list of sort types to look for upon a sort being called with slave process
```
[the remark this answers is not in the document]
eointeague: Think I have managed to revert back for up to date version and get rid of unintended changes.

code/processes/wdb.q
eointeague: Have added formatting and comments

code/processes/wdb.q
eointeague: Have added an if statement to print out process which failed to reload and the error. Would it be better to print out the failed processes and errors in the summary at the end - rather than what I have added now which should print out the error after the reload message is sent?

code/processes/wdb.q
eointeague: Have added types to the columns in d

code/processes/wdb.q
eointeague: Have edited this to a delete. However, would this be an issue in the scenario where the last process to reload fails? As the .wdb.handler is called twice re:
```
  [({@[`. `reload;x;
        /-if process fails to reload message with error entered into .wdb.d and  output error to process error log
                                {[ptype;e](neg .z.w)(`.wdb.handler;(ptype;0b;`$"failed with error: ",e));
                                .lg.e[`reloadproc;"failed to reload ",string[ptype]," from .wdb.reloadproc call. The error was : ",e]}[y]];
        /-Successful reload message to be sent to handler and entered into .wdb.d for end of reload summary logging
                                 (neg .z.w)(`.wdb.handler;(y;1b;`$"successfully reloaded")); (neg .z.w)[]};d;ptype);h;ptype];
```
As the error trap message will cause the reload end process to run i.e. calling .wdb.flushend[] and setting .wdb.reloadcomplete to 1b and deleting .wdb.d. The second .wdb.handler call will execute while .wdb.d has been deleted. Which I think will an output print to the err_wdb1.log file as the initial if statements in .wdb.handler reference .wdb.d.

code/processes/wdb.q
```diff
@@ -205,11 +205,17 @@ endofdaysave:{[dir;pt]
 
 /- add entries to dictionary of callbacks. if timeout has expired or d now contains all expected rows then it releases each waiting process
 handler:{
-	.wdb.d[.z.w]:x;
+	/-insert process reload outcome into .wdb.d
+	if[not .z.w in  key .wdb.d;
+        	.wdb.d[.z.w]:x];
+	if[(not .wdb.d[.z.w]`status);
+		.lg.o[`reload;string[.wdb.d[.z.w]`process]," reload ", string[.wdb.d[.z.w]`result]]];
```
jonnypress, line 212: I think it's correct to log this as .lg.o rather than .lg.e
But it would be better to make the error more explicit i.e. 

"process equityrdb failed to reload with error ",error

code/processes/wdb.q
```diff
@@ -205,11 +205,17 @@ endofdaysave:{[dir;pt]
...
 	if[(.proc.cp[]>.wdb.timeouttime) or (count[.wdb.d]=.wdb.countreload);
 		.lg.o[`handler;"releasing processes"];
-		.wdb.flushend[];
-		.wdb.d:()!()];
+		.lg.o[`reload;string[count select from .wdb.d where status=1]," out of ", string[count .wdb.d]," processes successfully reloaded"];
+		.wdb.flushend[]];
+	/-delete .wdb.d when reloads completed
+	if[.wdb.reloadcomplete;delete d from `.wdb];
```
jonnypress, line 218: This will delete the whole thing which we don't want - we just want to delete the contents of .wdb.d
delete from `.wdb.d

code/processes/wdb.q
```diff
@@ -222,8 +228,11 @@ flushend:{
 	 .wdb.reloadcomplete:1b];
 	};
 
-/- initialise d
-d:()!()
+/- initialise d, keyed tale to track status of local reloads
+d:([`int$handle:()]`symbol$process:();`boolean$status:();`symbol$result:());
```
jonnypress, line 232: Can you come up with a better name for this table
I know this is what it was previously called, and it may well have been me that called it that, but it's awful

jonnypress: Like "reloadstatus" or something...

code/processes/wdb.q
```diff
@@ -372,13 +381,20 @@ endofdaysort:{[dir;pt;tablist;writedownmode;mergelimits;hdbsettings]
...
 /-function to send reload message to rdbs/hdbs
 reloadproc:{[h;d;ptype]
-	.wdb.countreload:count[raze .servers.getservers[`proctype;;()!();1b;0b]each reloadorder];
-	$[eodwaittime>0;
-		{[x;y;ptype].[{neg[y]@x};(x;y);{[ptype;x].lg.e[`reloadproc;"failed to reload the ",string[ptype]];'x}[ptype]]}[({@[`. `reload;x;()]; (neg .z.w)(`.wdb.handler;1b); (neg .z.w)[]};d);h;ptype];
-		@[h;(`reload;d);{[ptype;e] .lg.e[`reloadproc;"failed to reload the ",string[ptype],".  The error was : ",e]}[ptype]]
-	];
-	.lg.o[`reload;"the ",string[ptype]," has been successfully reloaded"];
-	}
+        .wdb.countreload:count[raze .servers.getservers[`proctype;;()!();1b;0b]each reloadorder];
```
jonnypress, line 384: Could you add some more comments around this as to what is happening

code/processes/wdb.q
```diff
@@ -372,13 +381,20 @@ endofdaysort:{[dir;pt;tablist;writedownmode;mergelimits;hdbsettings]
...
+        $[eodwaittime>0;
+                {[x;y;ptype].[{neg[y]@x};(x;y);{[ptype;x].lg.e[`reloadproc;"failed to reload the ",string[ptype]];'x}[ptype]]}
+                        [({@[`. `reload;x;
+	/-if process fails to reload message with error entered into .wdb.d and  output error to process error log
```
jonnypress, line 388: Are you using spaces or tabs for whitespace? Can you use spaces

code/processes/wdb.q
```diff
@@ -372,13 +381,20 @@ endofdaysort:{[dir;pt;tablist;writedownmode;mergelimits;hdbsettings]
...
-	.wdb.countreload:count[raze .servers.getservers[`proctype;;()!();1b;0b]each reloadorder];
-	$[eodwaittime>0;
-		{[x;y;ptype].[{neg[y]@x};(x;y);{[ptype;x].lg.e[`reloadproc;"failed to reload the ",string[ptype]];'x}[ptype]]}[({@[`. `reload;x;()]; (neg .z.w)(`.wdb.handler;1b); (neg .z.w)[]};d);h;ptype];
-		@[h;(`reload;d);{[ptype;e] .lg.e[`reloadproc;"failed to reload the ",string[ptype],".  The error was : ",e]}[ptype]]
-	];
-	.lg.o[`reload;"the ",string[ptype]," has been successfully reloaded"];
-	}
+        .wdb.countreload:count[raze .servers.getservers[`proctype;;()!();1b;0b]each reloadorder];
+        $[eodwaittime>0;
+                {[x;y;ptype].[{neg[y]@x};(x;y);{[ptype;x].lg.e[`reloadproc;"failed to reload the ",string[ptype]];'x}[ptype]]}
```
jonnypress, line 386: this is a really complex function. I think you need to separately define some of the lambdas on different lines to try to make it less complicaated e.g. 

sendfunc:{[x;y;ptype].[{neg[y]@x};(x;y);{[ptype;x].lg.e[`reloadproc;"failed to reload the ",string[ptype]];'x}[ptype]]}
reloadfunc:({@[`. `reload;x; ....

// then invoke
sendfunc[reloadfunc;h;ptype]

code/processes/wdb.q
```diff
@@ -372,13 +381,20 @@ endofdaysort:{[dir;pt;tablist;writedownmode;mergelimits;hdbsettings]
...
+                        [({@[`. `reload;x;
+	/-if process fails to reload message with error entered into .wdb.d and  output error to process error log
+                                {[ptype;e](neg .z.w)(`.wdb.handler;(ptype;0b;`$"failed with error: ",e));
+                                .lg.e[`reloadproc;"failed to reload ",string[ptype]," from .wdb.reloadproc call. The error was : ",e]}[y]];
```
jonnypress, line 390: Is this function definitely correct? It looks to me like if the reload fails then it will 

a) print that the reload has failed and send a error result back to the wdb
b) because the failure is caught, it will then send a pass result as well

jonnypress (changes requested):

code/processes/wdb.q
eointeague: Have renamed to .wdb.reloadstatus

code/processes/wdb.q
eointeague: Have edited the structure of this so the log in .lg.o looks like
```
2022.02.23D09:28:33.441218000|satoshi|wdb|wdb1|INF|reload|the rdb process reload has failed with error: eodtabcount
```
However my error message in the result column of my .wdb.reloadstatus is quite lengthy - looks something like this:
```
handle| process status result
------| ---------------------------------------------
8     | rdb     0      failed with error: eodtabcount
```
Would it be better to shorten that message that's added to .wdb.reloadstatus result column and then edit the .lg.o statement?

code/processes/wdb.q
eointeague: Apologies misread your last comment have changed this to a delete from

code/processes/wdb.q
eointeague: So at the moment it does as you have stated above - I have edited the .wdb.handler function to try and mitigate the problems with this. The handler function currently looks like this:
```
handler:{
        /-insert process reload outcome into .wdb.reloadstatus
        if[not .z.w in  key .wdb.reloadstatus;
                .wdb.reloadstatus[.z.w]:x];
        if[(not .wdb.reloadstatus[.z.w]`status);
                .lg.o[`reload;"the ", string[.wdb.reloadstatus[.z.w]`process]," process reload has ", string[.wdb.reloadstatus[.z.w]`result]]];
        if[(.proc.cp[]>.wdb.timeouttime) or (count[.wdb.reloadstatus]=.wdb.countreload);
                .lg.o[`handler;"releasing processes"];
                .lg.o[`reload;string[count select from .wdb.reloadstatus where status=1]," out of ", string[count .wdb.reloadstatus]," processes successfully reloaded"];
                .wdb.flushend[]];
        /-delete contents from .wdb.reloadstatus when reloads completed
        if[.wdb.reloadcomplete;delete from `.wdb.reloadstatus];
        };
```
I added this if statement at the start of the function when inserting the success/fail message that is passed as an argument when the handler function is called from .wdb.reloadproc:
```
 if[not .z.w in  key .wdb.reloadstatus;
                .wdb.reloadstatus[.z.w]:x];
```
In the case where a reload fails the .wdb.handler function will be first called with the failed message and that if statement will execute adding the message to .wdb.reloadstatus. When the following success message is called it shouldn't execute this if statement as the process handle is already a key in the table. 

The issue then if the final final process to reload fails:
The error message will be added to .wdb.reloadstatus and the count of the .wdb.reloadstatus will = the count of the number of processes to execute and this if statement will execute:
```
   if[(.proc.cp[]>.wdb.timeouttime) or (count[.wdb.reloadstatus]=.wdb.countreload);
                .lg.o[`handler;"releasing processes"];
                .lg.o[`reload;string[count select from .wdb.reloadstatus where status=1]," out of ", string[count .wdb.reloadstatus]," processes successfully reloaded"];
                .wdb.flushend[]];
```
as will this final if statement which will clear out the table
```
  if[.wdb.reloadcomplete;delete from `.wdb.reloadstatus];
```
and if the final reload fails the following success message will be inserted into the .wdb.reloadstatus - however as the reload has been completed the final if statement will execute again clearing out the table. This approach is not ideal and I'm working on a simpler approach now.

code/processes/wdb.q
jonnypress: Both of those look fine

code/processes/wdb.q
jonnypress: looks fine

code/processes/wdb.q
```diff
@@ -203,27 +203,32 @@ endofdaysave:{[dir;pt]
...
+/- add entries to table of callbacks. if timeout has expired or d now contains all expected rows then it releases each waiting process
 handler:{
-	.wdb.d[.z.w]:x;
-	if[(.proc.cp[]>.wdb.timeouttime) or (count[.wdb.d]=.wdb.countreload);
-		.lg.o[`handler;"releasing processes"];
-		.wdb.flushend[];
-		.wdb.d:()!()];
-	};
+	/-insert process reload outcome into .wdb.reloadsummary 
+                .wdb.reloadsummary[.z.w]:.wdb.reloadstatus;
```
jonnypress, line 209: what is .wdb.reloadstatus? 
Surely this should be the inbound parameter?

code/processes/wdb.q
```diff
@@ -372,13 +377,28 @@ endofdaysort:{[dir;pt;tablist;writedownmode;mergelimits;hdbsettings]
...
+        /-count of processes to be reloaded 
+        .wdb.countreload:count[raze .servers.getservers[`proctype;;()!();1b;0b]each reloadorder];
+        /-defining lambdas to be executed in conditional statement 
+        /-async call back function executed when eodwaittime>0
+        sendfunc:{[x;y;ptype].[{neg[y]@x};(x;y);{[ptype;x].lg.e[`reloadproc;"failed to reload the ",string[ptype]];'x}[ptype]]};
+        /-reload function sent to processes by sendfunc in order to call process to reload. Calls process to reload, if process fails to reload
+        /-execute lambda to set .wdb.reloadstatus to an error message. Then message wdb to execute wdb.handler function.
+        reloadfunc:{[d;ptype]@[`. `reload;d;
+        /-lambda to execute if reload fails, sets .wdb.reloadstatus to error message and logs error in process error log
+                         {[ptype;e](neg .z.w)(set;`.wdb.reloadstatus;(ptype;0b;`$"reload failed with error: ",e));
```
jonnypress, line 391: You don't need to set a variable here
You should be able to just do 

```(neg .z.w)(`.wdb.handler; (ptype;0b;`$"reload failed with error: ",e)```

I think and then you can get rid of the follow up async call

File: code/processes/wdb.q [hunk 16]
```diff
@@ -372,13 +377,28 @@ endofdaysort:{[dir;pt;tablist;writedownmode;mergelimits;hdbsettings]
...
+                         .lg.e[`reloadproc;"failed to reload ",string[ptype]," from .wdb.reloadproc call. The error was : ",e]}[ptype]];
+        /-send message to wdb to execute .wdb.handler function
+                         (neg .z.w)(`.wdb.handler;`); (neg .z.w)[]};
+        $[eodwaittime>0;
+                 sendfunc[(reloadfunc;d;ptype);h;ptype];       
+		 @[h;(`reload;d);{[ptype;e] .lg.e[`reloadproc;"failed to reload the ",string[ptype],".  The error was : ",e]}[ptype]]
+        ];
+        .lg.o[`reload;string[ptype]," reload has finished"];
```
jonnypress, line 399: I'm not sure this log message is correct (I know it's the same as it was previously). 

I think that if eodwaittime is > 0 then it should print that the reload has been sent
And if the reload is synchronous, it should print something depending on whether the reload was successful

code/processes/wdb.q
eointeague: .wdb.reloadstatus is defined in the .wdb.reloadproc function at the top
https://github.com/AquaQAnalytics/TorQ/blob/137521af9d0cf051fa8be64b9032fbdc12a3dfef/code/processes/wdb.q#L380-L381
If the process reload fails I have then used an async set here:
https://github.com/AquaQAnalytics/TorQ/blob/137521af9d0cf051fa8be64b9032fbdc12a3dfef/code/processes/wdb.q#L380-L381

code/processes/wdb.q
eointeague: https://github.com/AquaQAnalytics/TorQ/blob/137521af9d0cf051fa8be64b9032fbdc12a3dfef/code/processes/wdb.q#L389-L394

If I replace the async set message with this
```
(neg .z.w)(`.wdb.handler; (ptype;0b;`$"reload failed with error: ",e)
```
get rid of the following async call would the handler function then not only get called when the reload fails?

code/processes/wdb.q
eointeague: Then the handler is called here:
https://github.com/AquaQAnalytics/TorQ/blob/137521af9d0cf051fa8be64b9032fbdc12a3dfef/code/processes/wdb.q#L389-L392

And .wdb.reloadstatus should be set the correct values to be inserted to .wdb.reloadsummary here:
https://github.com/AquaQAnalytics/TorQ/blob/137521af9d0cf051fa8be64b9032fbdc12a3dfef/code/processes/wdb.q#L209

I pushed these changes in order to get away from the multiple handler calls if a process reload fails discussed earlier

code/processes/wdb.q
eointeague: Will look into this now

code/processes/wdb.q
```diff
@@ -372,13 +377,23 @@ endofdaysort:{[dir;pt;tablist;writedownmode;mergelimits;hdbsettings]
...
-	.lg.o[`reload;"the ",string[ptype]," has been successfully reloaded"];
-	}
+        /-count of processes to be reloaded 
+        .wdb.countreload:count[raze .servers.getservers[`proctype;;()!();1b;0b]each reloadorder];
+        /-defining lambdas to be in asynchronously calling processes to reload 
+        /-async call back function executed when eodwaittime>0
+        sendfunc:{[x;y;ptype].[{neg[y]@x};(x;y);{[ptype;x].lg.e[`reloadproc;"failed to reload the ",string[ptype]];'x}[ptype]]};
+        /-reload function sent to processes by sendfunc in order to call process to reload. If process fail to reload log error 
+        /-and call .wdb.handler with failed reload message. If reload is successful call .wdb.handler with successful reload message.
+        reloadfunc:{[d;ptype] r:@[`. `reload;d;{x}];
```
drgdavies, line 387: Ok so for this - you could actually log out your error message below inside and still return x.
{.lg.e[`reloadproc;"failed too....",x;x}

Could also then consolidate your neg .z.w down to one line.
By working out 0b/1b and success reload based on the if conditional

code/processes/wdb.q
```diff
@@ -372,13 +377,23 @@ endofdaysort:{[dir;pt;tablist;writedownmode;mergelimits;hdbsettings]
...
+        /-count of processes to be reloaded 
+        .wdb.countreload:count[raze .servers.getservers[`proctype;;()!();1b;0b]each reloadorder];
+        /-defining lambdas to be in asynchronously calling processes to reload 
+        /-async call back function executed when eodwaittime>0
+        sendfunc:{[x;y;ptype].[{neg[y]@x};(x;y);{[ptype;x].lg.e[`reloadproc;"failed to reload the ",string[ptype]];'x}[ptype]]};
+        /-reload function sent to processes by sendfunc in order to call process to reload. If process fail to reload log error 
+        /-and call .wdb.handler with failed reload message. If reload is successful call .wdb.handler with successful reload message.
+        reloadfunc:{[d;ptype] r:@[`. `reload;d;{.lg.e[`reloadproc;"failed to reload from .wdb.reloadproc call. The error was : ",x];x}];
+	$[10h = type r;
+                 (reloadstatus:(ptype;0b;`$"reload failed with error ",r));
```
drgdavies, line 389: Could still do this as one line potentially like the below
(ptype;10h<>type r;(,/)("reload ";$[10h<>type r;"succeeded";"failed with error "];r))

jonnypress:
Hey @eointeague - as far as you are concerned, is this good to go now?

eointeague:
Hi @jonnypress - I think it's ready, but can make any changes if needed.



> Hey @eointeague - as far as you are concerned, is this good to go now?

code/processes/wdb.q
```diff
@@ -203,27 +203,32 @@ endofdaysave:{[dir;pt]
...
+/- add entries to table of callbacks. if timeout has expired or d now contains all expected rows then it releases each waiting process
 handler:{
-	.wdb.d[.z.w]:x;
-	if[(.proc.cp[]>.wdb.timeouttime) or (count[.wdb.d]=.wdb.countreload);
-		.lg.o[`handler;"releasing processes"];
-		.wdb.flushend[];
-		.wdb.d:()!()];
-	};
+	/-insert process reload outcome into .wdb.reloadsummary 
+                 .wdb.reloadsummary[.z.w]:x;
```
jonnypress, line 209: this shouldn't be indented so far to RHS

code/processes/wdb.q
```diff
@@ -203,27 +203,32 @@ endofdaysave:{[dir;pt]
...
+	 @[{neg[x]"";neg[x][]};;()] each key reloadsummary;
 	 informgateway(`reloadend;`);
 	 .lg.o[`sort;"end of day sort is now complete"];
 	 .wdb.reloadcomplete:1b];
 	};
 
-/- initialise d
-d:()!()
+/- initialise reloadsummary, keyed tale to track status of local reloads
+reloadsummary:([`int$handle:()]`symbol$process:();`boolean$status:();`symbol$result:());
```
jonnypress, line 231: this syntax isn't correct

Should be e.g. handle:`int$()

code/processes/wdb.q
```diff
@@ -372,13 +377,24 @@ endofdaysort:{[dir;pt;tablist;writedownmode;mergelimits;hdbsettings]
...
+        /-reload function sent to processes by sendfunc in order to call process to reload. If process fail to reload log error 
+        /-and call .wdb.handler with failed reload message. If reload is successful call .wdb.handler with successful reload message.
+        reloadfunc:{[d;ptype] r:@[`. `reload;d;{.lg.e[`reloadproc;"failed to reload from .wdb.reloadproc call. The error was : ",x];x}];
+                (neg .z.w)(`.wdb.handler;(ptype;10h<>type r;$[10h~type r;`$"reload failed with error ",r;`$"reloaded successfully"]));(neg .z.w)[]};
+        /-reload function to be executed if eodwaitime = 0 - sync message processes to reload and log if reload was successful or failed
+        syncreloadfunc:{[h;d;ptype] r:@[h;(`reload;d);{[ptype;e] .lg.e[`reloadproc;"failed to reload the ",string[ptype],". The error was : ",e];e}[ptype]];
+                .lg.o[`reloadproc;"the ", string[ptype]," ", $[10h~type r; "failed to reload with error ",r;"successfully reloaded"]]}; 
+        $[eodwaittime>0;
+                 (.lg.o[`reloadproc;"reload call has been sent to ", string[ptype]];
```
jonnypress, line 393: this should not be ()
It should be []
[] in an if-else defines a block of statements to execute. You have defined a list here

code/processes/wdb.q
```diff
@@ -372,13 +377,24 @@ endofdaysort:{[dir;pt;tablist;writedownmode;mergelimits;hdbsettings]
...
-	}
+        /-count of processes to be reloaded 
+        .wdb.countreload:count[raze .servers.getservers[`proctype;;()!();1b;0b]each reloadorder];
+        /-defining lambdas to be in asynchronously calling processes to reload 
+        /-async call back function executed when eodwaittime>0
+        sendfunc:{[x;y;ptype].[{neg[y]@x};(x;y);{[ptype;x].lg.e[`reloadproc;"failed to reload the ",string[ptype]];'x}[ptype]]};
+        /-reload function sent to processes by sendfunc in order to call process to reload. If process fail to reload log error 
+        /-and call .wdb.handler with failed reload message. If reload is successful call .wdb.handler with successful reload message.
+        reloadfunc:{[d;ptype] r:@[`. `reload;d;{.lg.e[`reloadproc;"failed to reload from .wdb.reloadproc call. The error was : ",x];x}];
+                (neg .z.w)(`.wdb.handler;(ptype;10h<>type r;$[10h~type r;`$"reload failed with error ",r;`$"reloaded successfully"]));(neg .z.w)[]};
```
jonnypress, line 388: What is the 10h<>type r check about
You shouldn't be relying on whether the function returns a string as to whether it is an error (as reload function might correctly return a string)
You need to catch that error explicitly

code/processes/wdb.q
eointeague: Have changed this now so that the reload call has been sent log message is before the conditional, so there it logs this whether it is an async or sync reload call. Thought it would be more complete logging to have it for either option, however can change this if needs be

PR #425, opened 2022-03-07
TCS-31

eoinnagle:
Adding unit tests for configload function

### DIFF
--- a/code/rdb/datastripe.q
+++ b/code/rdb/datastripe.q
@@ -1,3 +1,0 @@
-\d .ds
-
-segmentid: "J"$.proc.params[`segid]		// segmentid variable defined by applying key to dictionary of input values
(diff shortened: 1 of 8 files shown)


code/rdb/datastripe.q
```diff
@@ -0,0 +1,3 @@
+\d .ds
+
+segmentid: "J"$.proc.params[`segid]		// segmentid variable defined by applying key to dictionary of input values
```
jwrno1, line 3: This is out of date I'm sure, this file should no longer exist, I don't think. Can you make sure your branch is up to date?

tests/stp/configload/existTests/run.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+${TORQHOME}/torq.sh start all
```
jwrno1, line 3: Do either of your set of tests need the full stack to be started? Or do you only need the stp to be started?

eoinnagle: I don't think so, only stp. just tested without startall and stopall and tests working ok. Will remove and push

rian-morgan (approved):

jwrno1 (approved):

PR #436, opened 2022-03-14
Dataaccess improvements/fixes

eointeague:
-Add virtual column i to available columns to be used for aggregations
-Fix bug found, postprocessing and sublist keys not executing when queries sent across gateway
-Fixes to be made to documentation, eg example dictionaries throwing a length error due to mismatching key and value length

### DIFF
--- a/code/dataaccess/getdata.q
+++ b/code/dataaccess/getdata.q
@@ -39,10 +39,12 @@ getdata:{[inputparams]
   if[not 0~count (queryparams`ordering);
     table:f[table;;queryparams`ordering]/[1;last til count (queryparams`ordering)]];         
 // rename the columns  
-  result:queryparams[`renamecolumn] xcol table;                                              
-  if[10b~`sublist`procs in key inputparams;result:select [inputparams`sublist] from result];
-    // apply post-processing function  
-    if[10b~in[`postprocessing`procs;key inputparams];                                                           
+  result:queryparams[`renamecolumn] xcol table; 
+// apply sublist function if called in process or query to single process called from gateway                                             
+  if[(10b~`sublist`procs in key inputparams)or((1b~`sublist in key inputparams)and(1~count inputparams `procs));
+        result:select [inputparams`sublist] from result];
+// apply post-processing function if called in process or query to single process called from gateway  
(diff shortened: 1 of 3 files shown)


code/dataaccess/getdata.q
```diff
@@ -39,10 +39,12 @@ getdata:{[inputparams]
...
-  result:queryparams[`renamecolumn] xcol table;                                              
-  if[10b~`sublist`procs in key inputparams;result:select [inputparams`sublist] from result];
-    // apply post-processing function  
-    if[10b~in[`postprocessing`procs;key inputparams];                                                           
+  result:queryparams[`renamecolumn] xcol table; 
+// apply sublist function if called in process or query to single process called from gateway                                             
+  if[(10b~`sublist`procs in key inputparams)or((1b~`sublist in key inputparams)and(1~count inputparams `procs));
+        result:select [inputparams`sublist] from result];
+// apply post-processing function if called in process or query to single process called from gateway  
+    if[(10b~in[`postprocessing`procs;key inputparams])or((1b~`postprocessing in key inputparams)and(1~count inputparams `procs));
```
jonnypress, line 47: I know this isn't  your change... but I think it makes more sense to apply sublist *after* post processing
For example - if I want to post process a result to re-aggregate it, and only show the first 10 rows of the result
Then we need to run the aggregation first, and then the sublist
If we do it the other way we are extracting the first 10 rows and then aggregating over that which is wrong

code/gateway/dataaccess.q
```diff
@@ -134,23 +147,6 @@ attributesrouting:{[options;procdict]
     :types;
     };
 
-// mixture of all the post processing functions in gw
-returntab:{[input;tab;reqno]
```
jonnypress, line 138: So this function isn't used anymore anywhere?

jonnypress (approved):

code/gateway/dataaccess.q
eointeague: No, it seems to only be defined here and not used anywhere

code/dataaccess/getdata.q
eointeague: Will address this and change the order

jonnypress (approved):

PR #442, opened 2022-04-05
Tcs 35 functionality to clear data from data store after it has been saved down by the tailer process

jdhalligan:


code/common/datastripe.q
```diff
@@ -1,3 +1,26 @@
...
 
 segmentid: "J"$.proc.params[`segid]		// segmentid variable defined by applying key to dictionary of input values
+
+\d .
+
+.rdb.datastripeendofperiod:{[currp;nextp;data]
+    .lg.o[`reload;"reload command has been called remotely"];
+    
+    // remove periods of data from tables
+    t:tables[`.] except .rdb.ignorelist;
```
drgdavies, line 13: Why are we calling this like tables[`.]

code/common/datastripe.q
```diff
@@ -1,3 +1,26 @@
...
+    lasttime:currp-.ds.periodstokeep*(nextp-currp);
+    tabs:{![x;enlist (<;y;z);0b;0#`]}'[t;`time;lasttime];
```
drgdavies, line 15: I think this should be a separate function as opposed to a lambda.  It seems useful and could potentially be reused?

code/common/datastripe.q
```diff
@@ -1,3 +1,26 @@
...
+    .lg.o[`reload;"Kept ",string[.ds.periodstokeep]," period",$[.ds.periodstokeep>1;"s";""]," of data from : ",", " sv string[tabs]];
+    };
+
+
+initdatastripe:{
+	// update endofday and endofperiod functions
+    endofday::.rdb.endofday;
```
drgdavies, line 22: whats the reason for setting these like this?  are they vars required by the datastriping?

code/wdb/datastripe.q
```diff
@@ -0,0 +1,66 @@
...
+
+\d .
+
+.wdb.datastripeendofperiod:{[currp;nextp;data]
+    .lg.o[`reload;"reload command has been called remotely"];
+    
+    // remove periods of data from tables
+    t:tables[`.] except .wdb.ignorelist;
+    lasttime:currp-.ds.periodstokeep*(nextp-currp);
+    tabs:{![x;enlist (<;y;z);0b;0#`]}'[t;`time;lasttime];
```
drgdavies, line 15: Is this the same as the lambda I flagged in common? If so looks like it should be spearated.

code/wdb/datastripe.q
```diff
@@ -0,0 +1,66 @@
...
+    .lg.o[`reload;"Kept ",string[.ds.periodstokeep]," period",$[.ds.periodstokeep>1;"s";""]," of data from : ",", " sv string[tabs]];
+    };
+
+
+initdatastripe:{
+	// update endofday and endofperiod functions
+    //endofday::endofday;
```
drgdavies, line 22: Why is endofday not needed?

code/wdb/datastripe.q
```diff
@@ -0,0 +1,66 @@
...
+    endofperiod::.wdb.datastripeendofperiod;
+    };
+
+if[.ds.datastripe;.proc.addinitlist[(`initdatastripe;`)]];
```
drgdavies, line 26: Is this logic not already in the common?

code/wdb/datastripe.q
```diff
@@ -0,0 +1,66 @@
...
+	/- upsert select data matched on partition to specific directory
+	.[upsert;
+	  (directory;enumdata);
+	  {[e] .lg.e[`upserttopartition;"Failed to save table to disk : ",e];'e}
+	];
+	};	
+
+savetablesoverperiod:{[dir;tablename;nextp]
+	/- function to get keycol for table from access table
+	keycol:`sym; / do we need to consider access table in this??
```
drgdavies, line 47: Yes - I think you want this to be interchangeable based on the segment configuration

config/settings/wdb.q
```diff
@@ -23,7 +23,7 @@ mode:`save
                                                                                             //                     save mode process.  When this is triggered it will sort the
                                                                                             //                     data on disk, apply attributes and the trigger a reload on the
                                                                                             //                     rdb and hdb processes
-writedownmode:`default                                                                      // the wdb process can periodically write data to disc and sort at EOD in two ways:
+writedownmode:`partbyattr                                                                      // the wdb process can periodically write data to disc and sort at EOD in two ways:
```
drgdavies, line 26: What does this mean for standard torq setups if we are changing this? Does it impact how they work?

If we have one called default which is not the default....

code/common/datastripe.q
jdhalligan: this function is in the .ds namespace so tables[] would get all tables in .ds namespace. need to call tables[`.] to get the tables in the root namespace instead

drgdavies (approved):

AdamBrown2 (approved):

PR #448, opened 2022-04-27
Wdb partbyattr copy data column by column

eointeague:
The standard partbyattr writedown method merges data by entire partition, this is causing issues where certain partitions have a large amount of data and reading the whole partition in causes memory limit issues. To approach this problem 3 configurable merge methods have been created, standard partbyattr, column by column and a hybrid method (that uses the standard partbyattr or column by column method depending on the row count of the partition being merged). 

Added 3 configurable methods to merge data - this involved removing the existing lambda used to merge data in the .wdb.merge function and separating each configurable method into it's own function. The .wdb.merge lambda has been replaced with a conditional which calls the configured option set by variable .wdb.mergemethod.

Added table that keeps track of table row counts and bytesize when data is periodically upserted to temporary storage, this table is used in the hybrid option. Where an exec is done on the table to get the table partitions where the row count/bytesize is over the row limit. Those partitons over the limit are merged with the column by column function. And the remaining partitions (.i.e. those under the limit) are merged with the merge by partition method.

The original merge by entire partition method used a scan iterator over a list of temp storage partitions an if a limit was exceeded, a writedown was triggered. This has been changed. When merging data by entire partition, the .merge.getpartchunks function is called - this uses the .merge.partsizes table to do a running sum on the row count/byte size and partitions as a list/nested list that can be read in in batch without exceeding the configurable limit.


code/processes/wdb.q
```diff
@@ -21,12 +21,17 @@ mode:@[value;`mode;`saveandsort];                                          /-the
                                                                            /-                                              data on disk, apply attributes and the trigger a reload on the
                                                                            /-                                              rdb and hdb processes
 
-writedownmode:@[value;`writedownmode;`default];                            /-the wdb process can periodically write data to disc and sort at EOD in two ways:
+writedownmode:@[value;`writedownmode;`partbyattr];                         /-the wdb process can periodically write data to disc and sort at EOD in two ways:
```
jonnypress, line 24: Don't change this default value

eointeague: Have reset to default

code/processes/wdb.q
```diff
@@ -295,54 +304,100 @@ endofdaysortdate:{[dir;pt;tablist;hdbsettings]
...
-      .lg.w[`merge;"no records found for ",(string tableinfo[0]),", merging empty table"];
-      dest set @[.Q.en[hdbsettings[`hdbdir];tableinfo[1]];.merge.getextrapartitiontype[tableinfo[0]];`p#];
-      //.lg.o[`merge;"setting attributes"];
-      //@[dest;;`p#] each .merge.getextrapartitiontype[tableinfo[0]];
-      //.lg.o[`merge;"merge complete"];
+    [.lg.w[`merge;"no records found for ",(string tabname),", merging empty table"];
+     (` sv dest,`) set @[.Q.en[hdbsettings[`hdbdir];tableinfo[1]];.merge.getextrapartitiontype[tabname];`p#];
+     //.lg.o[`merge;"setting attributes"];
+     //@[dest;;`p#] each .merge.getextrapartitiontype[tabname];
+     //.lg.o[`merge;"merge complete"];
```
jonnypress, line 383: Prob better to just remove this code if these lines aren't used

code/processes/wdb.q
```diff
@@ -27,8 +27,13 @@ writedownmode:@[value;`writedownmode;`default];                            /-the
                                                                            /- 2. partbyattr                -       the data is partitioned by [ partitiontype ] and the column(s) assigned the parted attributed in sort.csv
                                                                            /-                                      at EOD the data will be merged from each partiton before being moved to hdb
 
-mergenumrows:@[value;`mergenumrows;100000];                                /-default number of rows for merge process
-mergenumtab:@[value;`mergenumtab;`quote`trade!10000 50000];                /-specify number of rows per table for merge process
```
jonnypress, line 31: We need to maintain backward compatibilty, so we can't remove or rename these variables. Surely we still need them?

code/common/merge.q
```diff
@@ -1,4 +1,7 @@
 \d .merge
+.merge.mergebybytesize:0b;                                                    /- merge limit configuration - default is 0b row count limit 1b is byte size limit
```
jonnypress, line 2: Should this be mergebybytelimit ?

code/processes/wdb.q
```diff
@@ -295,67 +308,77 @@ endofdaysortdate:{[dir;pt;tablist;hdbsettings]
...
+               if[gc;.gc.run[]]};`);(neg x)(::)} each .z.pd[];
+    ];	
     [.lg.o[`merge;"merging on main"];
      reloadsymfile[.Q.dd[hdbsettings `hdbdir;`sym]];
-     merge[dir;pt;;mergelimits;hdbsettings] each flip (key tablist;value tablist)]];
+     merge[dir;pt;;mergelimits;hdbsettings;mergemethod] each flip (key tablist;value tablist);
+     .lg.o[`eod;"Delete from partsizes"];
+     delete from `.merge.partsizes;
+     /- run a garbage collection (if enabled)
+     if[gc;.gc.run[]];
```
jonnypress, line 374: We want to do this right at the end do we not?

config/settings/wdb.q
```diff
@@ -28,9 +29,14 @@ writedownmode:`default
...
+                                                                                            // 1. part                      -       the entire partition is merged to the hdb 
+                                                                                            // 2. col                       -       each column in the temporary partitions are merged individually 
+                                                                                            // 3. hybrid                    -       partitions merged by column or entire partittion based on byte limit  
 												
 mergenumrows:100000                                                                         // default number of rows for merge process
 mergenumtab:`quote`trade!10000 50000                                                        // specify number of rows per table
+mergenumbytes:5000000                                                                        // default partition bytesize for merge limit in merge process (only used when .merge.mergebybytelimit=1b)
```
jonnypress, line 39: Can we default this to a bigger value ... 5MB is pretty small 
Let's say 500 MB (still quite small)

jonnypress (approved):
Looks good

PR #449, opened 2022-05-03
Update chainedtp.q

jtennyson724:
This is not a problem in segmented version, endofday sent down not .u.end

### DIFF
--- a/code/processes/chainedtp.q
+++ b/code/processes/chainedtp.q
@@ -174,7 +174,7 @@ end:{[d]
   /- roll over the log you need a new log for next days data 
   .ctp.refreshtp[d+1];
   /- push endofday messages to subscribers
-  (neg union/[w[;;0]])@\:(`.u.end;d)
+  (neg union/[.stpps.allsubhandles[]])@\:(`.u.end;d)
   }
 
 \d .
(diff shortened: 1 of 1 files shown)


code/processes/chainedtp.q
```diff
@@ -174,7 +174,7 @@ end:{[d]
   /- roll over the log you need a new log for next days data 
   .ctp.refreshtp[d+1];
   /- push endofday messages to subscribers
-  (neg union/[w[;;0]])@\:(`.u.end;d)
+  (neg union/[.stpps.allsubhandles[]])@\:(`.u.end;d)
```
jonnypress, line 177: Could / should this be both lists of handles? 

i.e. w[;;0],.stpps.allsubhandles[]

jtennyson724: Idea would work except .u.w isn't defined straight off the bat in sctp. Sctp would need a .u.init[] or similar to have the .u.w table defined off the bat

jtennyson724: But in checking this, causes a type error

jonnypress: Could you make it 

neg union[@[value;(`.stpps.allsubhandles;`);()]; @[{union/[(value x)[;;0]};`.u.w;()]]

Basically we need to maintain backward compatibility here as well as making sure it works for new. There are probably a number of different ways that people could have this deployed - so if we change to this I think it catches most cases - whether they are using new or old subscription logic (or some mixture) I think it should work. It shouldn't need to have .u.w defined as that should be error trapped

jonnypress (approved):

PR #451, opened 2022-05-04
Tcs 111 updating config table

AdamBrown2:
Combined all previous csv's into one segmenting.csv in TorQ/config
Updated TorQ/code/segmentedtickerplant/datastripe.q to use the one csv
Unit test currently being worked on

### DIFF
--- a/code/segmentedtickerplant/datastripe.q
+++ b/code/segmentedtickerplant/datastripe.q
@@ -7,7 +7,6 @@
 
 //if statement checks segmenting.csv and filtermap.csv exist. If not, process exited and message sent to error logs.
 //configload transfoms segmenting.csv into table format so it can be accessed.
-//configload transforms filtermap.csv into table format then into a mapping of wcRef to filter which can be accessed and applied to data.
 //checks csv files load correctly. If not, process exited and message sent to error logs.
 
 configload:{
@@ -36,6 +35,7 @@ segmentfilter:{[tbl;segid]
      };
 
 // Subscribe to particular segment using segmentID based on .u.sub
+// Uses .stpps.segmentconfig to handle subscriptions for each table
 subsegment:{[tbl;segid];
      //tablename and segmentid used to get filters
      if[tbl~`;:.z.s[;segid] each .stpps.t];
--- a/code/segmentedtickerplant/datastripe.q
+++ b/code/segmentedtickerplant/datastripe.q
@@ -1,12 +1,9 @@
 //configload loads configurable .csv files into memory so relevant filters can be applied to data.
-//loads segmenting.csv which defines how many segments there will be.
-//loads filtermap.csv which defines what filters should be applied to each segment.
 
 //.ds.segmentconfig/segmentfiltermap are variables which can take different csv files. These files can be chosen on in process.csv.
 //error trap these .ds variables incase this file is loaded stand alone.
 
-.ds.segmentconfig:@[value;`.ds.segmentconfig;`segmenting.csv];
-.ds.filtermap:@[value;`.ds.filtermap;`filtermap.csv];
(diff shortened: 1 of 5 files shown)


code/segmentedtickerplant/datastripe.q
```diff
@@ -1,24 +1,20 @@
 //configload loads configurable .csv files into memory so relevant filters can be applied to data.
-//loads segmenting.csv which defines how many segments there will be.
-//loads filtermap.csv which defines what filters should be applied to each segment.
 
 //.ds.segmentconfig/segmentfiltermap are variables which can take different csv files. These files can be chosen on in process.csv.
 //error trap these .ds variables incase this file is loaded stand alone.
 
-.ds.segmentconfig:@[value;`.ds.segmentconfig;`segmenting.csv];
-.ds.filtermap:@[value;`.ds.filtermap;`filtermap.csv];
+.ds.segmentconfig:@[value;`.ds.segmentconfig;`segmenting1.csv];
```
jonathonmcmurray, line 6: should this be `segmenting.csv` or is `segmenting1.csv` intentional?

code/segmentedtickerplant/datastripe.q
```diff
@@ -30,29 +26,33 @@ initdatastripe:{
...
      filter:segmentfilter[tbl;segid];
-     if[filter~"";
-          .lg.e[`sub;m:"Incorrect pairing of table ",string[tbl]," and segmentID ",string[segid]," not found in .stpps.segmentconfig"];
-          :();
-     ];
-     
-     .ps.subtablefiltered[string[tbl];filter;""]
+     $[tbl in .stpps.fullsubscriptions;[.stpps.suball[tbl]];
+     [$[tbl in .stpps.ignoredtables;
+                    [.lg.o[`sub;m:"Table ",string[tbl]," is in ignoredtables.csv and will not be subscribed to"];
```
jonathonmcmurray, line 49: I think this log message probably needs updated?

code/segmentedtickerplant/datastripe.q
```diff
@@ -30,29 +26,33 @@ initdatastripe:{
...
+                    :();];
+       [$[filter~"";             
+                    [.lg.e[`sub;m:"Incorrect pairing of table ",string[tbl]," and segmentID ",string[segid]," not found in .stpps.segmentconfig"];
+                    :(tbl;`err`msg!(`segmentid;m));]
+                    ;
+                    [.ps.subtablefiltered[string[tbl];filter;""]]]]]]]
```
jonathonmcmurray, line 55: I think this nested conditional is a bit confusing and hard to follow - could it be either re-written as separate `if` statements, or re-formatting in some way?

e.g. I think something like the following would have the same functionality but is easier to follow IMO using early exits so that each subsequent if statement only gets executed if the previous ones were false:

```suggestion
     if[tbl in .stpps.fullsubscriptions;:.stpps.suball[tbl]];
     if[tbl in .stpps.ignoredtables;
        .lg.o[`sub;m:"Table ",string[tbl]," is in ignoredtables.csv and will not be subscribed to"];
        :();
     ];
     if[filter~"";          
        .lg.e[`sub;m:"Incorrect pairing of table ",string[tbl]," and segmentID ",string[segid]," not found in .stpps.segmentconfig"];
        :(tbl;`err`msg!(`segmentid;m));
     ];
     .ps.subtablefiltered[string[tbl];filter;""]
```

code/segmentedtickerplant/datastripe.q
AdamBrown2: Yes i missed that thanks

code/segmentedtickerplant/datastripe.q
AdamBrown2: Did not know it could be done like that and it all works Thanks

PR #455, opened 2022-05-18
Tcs 115 ensure list of tables align with subscription and extend for other timestamps

CReid98:


code/common/datastripe.q
```diff
@@ -4,25 +4,15 @@
...
-    t:tables[`.] except .rdb.ignorelist;
-    lasttime:currp-.ds.periodstokeep*(nextp-currp);
-    tabs:{![x;enlist (<;y;z);0b;0#`]}'[t;`time;lasttime];
-    .lg.o[`reload;"Kept ",string[.ds.periodstokeep]," period",$[.ds.periodstokeep>1;"s";""]," of data from : ",", " sv string[tabs]];
+loadtablekeycols:{[]                                                          // loading the tablekeycols config as a dictionary
+    keypath:first .proc.getconfigfile[string .ds.tablekeycolsconfig];
+    @[{tablekeycols:(!/)(("SS";enlist",")0: hsym x)`tablename`keycol};keypath;{.lg.e[`init;"Failure in loading ",string y]}[;keypath]]
     };
 
+getstarttime:{[x] first x[`time]};
```
drgdavies, line 15: Does this need to be min and max instead of first and last?

code/segmentedtickerplant/datastripe.q
```diff
@@ -47,12 +51,13 @@ subsegment:{[tbl;segid];
           :();
      ];
      filter:segmentfilter[tbl;segid];
-     if[filter~"";
+     $[tbl in .stpps.ignoredtables;[.stpps.suball[tbl]];
```
drgdavies, line 54: Why is there square brackets around the true part?
Also would look better with extra line as:
$[conditional;
    true;
    false]

code/wdb/datastripe.q
```diff
@@ -45,30 +76,29 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
...
-	/-upsert table to partition
-	upserttopartition[dir;tablename;keycol;;nextp] each enumdata;
-	/- delete data from last period
-	.[.ds.deletetablebefore;(tablename;`time;nextp)];
-	/ .[{![x;enlist(<;`time;y);0b;0#`]};(tablename;nextp)];
-	/- run a garbage collection (if enabled)
-	.gc.run[];
-	};
+    /- function to get keycol for table from access table
+    keycol:$[.wdb.tablekeycols[tablename]=`;`sym;.wdb.tablekeycols[tablename]];
```
drgdavies, line 83: would a null check be better here?

code/common/datastripe.q
CReid98: Yes, it will need to be in case data arrives out of order.

config/settings/rdb.q
```diff
@@ -31,9 +31,9 @@ subfiltered:0b      //allows subscription filters to be loaded and applied in th
 connectonstart:1b   //rdb connects to tickerplant as soon as it is started
 
 \d .ds
-datastripe:1b
+datastripe:0b
 period:0D01                                    
-periodstokeep:4
+/periodstokeep:4
```
drgdavies, line 36: Are these supposed to be commented out (same for next file

code/wdb/datastripe.q
```diff
@@ -45,30 +76,29 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
...
+    /- get distint values to partition table on
+    partitionlist:raze value each ?[[`.]tablename;();1b;enlist[keycol]!enlist keycol];
```
jonathonmcmurray, line 84: I think you could use a functional exec instead of a select here & simplify a bit

```q
q)t:([]abc:1 1 2 3 4 4 4 5)
q)?[t;();();(distinct;`abc)]
1 2 3 4 5
```

(wouldn't need the `raze value each` and the functional query itself should be a bit simpler I think)

code/wdb/datastripe.q
```diff
@@ -45,30 +76,29 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
...
+    /- enumerate table to be upserted and get each table by sym
```
jonathonmcmurray, line 85: "get each table by sym" doesn't sound quite right to me, maybe "split each table by keycol" (seeing as that isn't necessarily sym)

code/wdb/datastripe.q
CReid98: yes, I've made this change

config/settings/rdb.q
CReid98: Yes, this has been moved to the default.q file. I will delete these lines

code/wdb/datastripe.q
```diff
@@ -45,30 +76,29 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
...
+    enumdata:{[dir;tablename;keycol;nextp;s] .Q.en[dir;0!?[[`.]tablename;((<;`time;nextp);(=;keycol;enlist s));0b;()]]}[dir;tablename;keycol;nextp]'[partitionlist];
```
jonathonmcmurray, line 86: have you tested to see if it's more efficient to enumerate first & then split vs split & then enumerate as is done here?

imagine it may not make much difference, but on the off chance there is an efficiency gain to be made here

code/wdb/datastripe.q
```diff
@@ -45,30 +76,29 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
...
+    /-upsert table to partition
+    upserttopartition[dir;tablename;keycol;;nextp] each enumdata;
```
jonathonmcmurray, line 88: I might be missing something here, but will this upsert each of the split tables to same partition? I don't see `partitionlist` being used here, so it seems like we just split it & then we're upserting it all to same partition - in which case, why do we need to split it?

code/wdb/datastripe.q
```diff
@@ -45,30 +76,29 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
...
+    /- delete data from last period
+    .[.ds.deletetablebefore;(tablename;`time;nextp)];
+    .[{![x;enlist(<;`time;y);0b;0#`]};(tablename;nextp)];
```
jonathonmcmurray, line 91: another instance were we could be using the defined version of `deletetablebefore` I think

actually that's what's in the line above - is this line a duplicate?

jonathonmcmurray (commented):
a few comments here - I haven't looked at the tests yet but will try get a look at them later as well

code/wdb/datastripe.q
```diff
@@ -2,26 +2,57 @@
...
+
     // remove periods of data from tables
     t:tables[`.] except .wdb.ignorelist;
-    lasttime:currp-.ds.periodstokeep*(nextp-currp);
+    lasttime:nextp-.ds.periodstokeep*(nextp-currp);
+
+    // update the access table in the wdb
+    // on first save down we need to replace the null valued start time in the access table
+    // using the first value in the saved data
+    .wdb.access:update start:(lasttime^(.ds.getstarttime each (key .wdb.tablekeycols)))^start from .wdb.access;
```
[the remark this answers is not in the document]
CReid98: yes, these can be removed

code/wdb/datastripe.q
```diff
@@ -2,26 +2,57 @@
 
 segmentid: "J"$.proc.params[`segid]		// segmentid variable defined by applying key to dictionary of input values
 
-td:hsym `$getenv`KDBTAIL
+td:hsym `$"/"sv (getenv`KDBTAIL;string .z.d)
```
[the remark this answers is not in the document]
CReid98: I've added .z.d to the directory path so that when I am loading the access table from disk I only have the access table for one day. I believe changes will need to be made to the end of day function to move and clear the access table on disk. I will make a note of this. td is the tail directory the tables are saved to interday. I will add a comment to this.

code/common/datastripe.q
```diff
@@ -4,25 +4,15 @@
...
+getendtime:{[x] last x[`time]};
 
-initdatastripe:{
-	// update endofday and endofperiod functions
-    endofday::.rdb.endofday;
-    endofperiod::.rdb.datastripeendofperiod;
-    };
-
-if[.ds.datastripe;.proc.addinitlist[(`initdatastripe;`)]];
+deletetablebefore:{![x;enlist (<;y;z);0b;0#`]}
```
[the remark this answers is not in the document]
CReid98: You're right, some of these functions are unclear or repeated. I'll clarify this.

code/rdb/datastripe.q
```diff
@@ -0,0 +1,38 @@
+\d .rdb
+
+datastripeendofperiod:{[currp;nextp;data]
+    .lg.o[`reload;"reload command has been called remotely"];
+
+    // remove periods of data from tables
+    t:tables[`.] except ignorelist;
+
+    lasttime:nextp-.ds.periodstokeep*(nextp-currp);
+    tabs:{![x;enlist (<;y;z);0b;0#`]}'[t;`time;lasttime];
```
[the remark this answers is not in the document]
CReid98: I've swapped this out for the `deletetablebefore` function

code/rdb/datastripe.q
```diff
@@ -0,0 +1,38 @@
...
+
+ext:{[accesstab]
+
+    .rdb.access:accesstab;
+
+    }
+
+initdatastripe:{
+    // update endofday and endofperiod functions
+    endofday::.rdb.endofday;
```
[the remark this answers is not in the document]
CReid98: removed redundant code

code/wdb/datastripe.q
```diff
@@ -2,26 +2,57 @@
...
 
 \d .
 
+// user definable function to modify the access table, call with access table as input
```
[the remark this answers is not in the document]
CReid98: access table is designed to be expandable for the user, if they want to add new columns or modify existing columns. It is possible that some modifications could break things, particularly when they modify the existing columns. Would adding guidlines to the documentation resolve this or should I be thinking about running additional checks?

code/wdb/datastripe.q
```diff
@@ -2,26 +2,57 @@
...
+    // call the savedown function
+    .ds.savealltablesoverperiod[.ds.td;nextp];
+
     tabs:.ds.deletetablebefore'[t;`time;lasttime];
-    / tabs:{![x;enlist (<;y;z);0b;0#`]}'[t;`time;lasttime];
     .lg.o[`reload;"Kept ",string[.ds.periodstokeep]," period",$[.ds.periodstokeep>1;"s";""]," of data from : ",", " sv string[tabs]];
-    };
+    
+    // update the access table on disk
+    atab:get(` sv(.ds.td;.proc.procname;`access));
```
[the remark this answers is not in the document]
CReid98: removed these brackets

code/wdb/datastripe.q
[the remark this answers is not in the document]
jonathonmcmurray: I guess I'm not quite understanding what would necessitate adding new columns or modifying existing columns - is there an example of something that a user might need to modify?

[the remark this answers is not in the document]
CReid98: They might want to change the timezone the start and end times are recorded for or add a new column showing the time the end of period function was called

code/wdb/datastripe.q
CReid98: changed

code/wdb/datastripe.q
CReid98: yes, I think this does work and is a lot simpler

code/wdb/datastripe.q
CReid98: this wasn't part that I worked on. I'll speak to Jamie and see if I can get some clarification.

code/wdb/datastripe.q
CReid98: yes, this is a duplicate. it has been removed now

code/wdb/datastripe.q
CReid98: This is handled in upsert to partition function, function has a line to take each unique value from the keycolumn and upsert to unique partition

jonathonmcmurray: OK - so why do we need to split it before sending to that function then? (Could be a valid reason for this that I'm missing)

code/wdb/datastripe.q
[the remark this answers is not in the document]
jonathonmcmurray: Hmm, I'm still not really getting it. My understanding of the access table is that it's essentially an internal TorQ implementation detail that will be used by data access API etc. to retrieve data. Why does the end user need to modify the timezones etc., will that type of stuff not be handled by the API?

It seems to me that if the user starts making modifications to the access table, they're quite likely going to need to make changes to the API as well to account for those changes in the access table? Is that what we'd expect here?

I think if we're having this, we definitely need some robust documentation around it. Seems like a potential [footgun](https://en.wiktionary.org/wiki/footgun)] to me.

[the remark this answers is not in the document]
CReid98: I'll get to work on documentation and try some test cases.

code/wdb/datastripe.q
CReid98: Yes, there is an efficiency gain here. Change has been made

code/wdb/datastripe.q
CReid98: Still looking over this but upserttopartition takes each split table, determines which syms it contains and then upserts it to the correct partition.

CReid98: The upserttopartition function was recalculating the partitionlist and then upserting to that partition, by taking the partitionlist as a variable in the upserttopartition function we can remove this redundant code and hopefully make this more clear.

drgdavies (approved):

code/segmentedtickerplant/datastripe.q
```diff
@@ -1,24 +1,14 @@
...
-     fmpath:first .proc.getconfigfile[string .ds.filtermap];
-     {if[()~key hsym x;.lg.e[`init;"The following file can not be found: ",string x]]} each (scpath;fmpath);
-     @[{.stpps.segmentconfig:("SIS";enlist",")0: hsym x};scpath;{.lg.e[`init;"Failure in loading ",string y]}[;scpath]];
-     @[{.stpps.segmentfiltermap:(!/)(("S*";enlist",")0: hsym x)`wcRef`filter};fmpath;{.lg.e[`init;"Failure in loading ",string y]}[;fmpath]];
+     scpath:first .proc.getconfigfile[string .ds.stripeconfig];
+     if[()~key hsym scpath;.lg.e[`init;"The following file can not be found: ",string scpath]];
+     .stpps.stripeconfig:.j.k read1 scpath;
+     defaults:{first (flip .stpps.stripeconfig[x])[`subscriptiondefault]}each key .stpps.stripeconfig;
+     errors:1+ where {[x] not ("ignore"~x) or ("all"~x)}each defaults;
+     {if[0<count x;.lg.e[`sub;m:"subscriptiondefault not defined as \"ignore\" or \"all\" for segment ",string[x]," "]]}each errors;
```
emansanti, line 11: Is this Adam's code that you added in so that you could make your code work with the striping JSON? Just asking not saying it's wrong or anything

CReid98: I think that is Adam's code

code/wdb/datastripe.q
```diff
@@ -2,73 +2,100 @@
 
 segmentid: "J"$.proc.params[`segid]		// segmentid variable defined by applying key to dictionary of input values
 
-td:hsym `$getenv`KDBTAIL
+td:hsym `$"/"sv (getenv`KDBTAIL;string .z.d)
```
emansanti, line 5: I think the latest structure we are going with is 
data/taildir/IDB1/2022.06.09
                  /IDB2/2022.06.10
                  
I think in other parts of torq (e,g, wdb.q) they use pt or the current partition variable as the partition to look at, instead of using .z.d.  Then at EOD this value is increased by one once the processes are fine to move onto the next day/partition. Just thinking that at the EOD there might be a period where we want the access tables from yesterday whilst the write down to HDB is taking place.

config/filtermap.csv
```diff
@@ -1 +1,3 @@
 wcRef,filter
+wca,sym in `AAPL
+wcb,ex="O"
```
emansanti, line 3: Is this file needed anymore, as striping.json makes it redundant?

config/ignoredtables.csv
```diff
@@ -0,0 +1,5 @@
+tables,
+logmsg,
+packets,
+quote_iex,
+trade_iex,
```
emansanti, line 5: Is this file needed anymore, as striping.json makes it redundant?

config/segmenting.csv
```diff
@@ -1 +1,5 @@
 table,segmentID,wcRef
+trade,1,wca
+quote,1,wca
+trade,2,wcb
+quote,2,wcb
```
emansanti, line 5: Is this file needed anymore, as striping.json makes it redundant?

config/settings/rdb.q
```diff
@@ -31,9 +31,8 @@ subfiltered:0b      //allows subscription filters to be loaded and applied in th
 connectonstart:1b   //rdb connects to tickerplant as soon as it is started
 
 \d .ds
-datastripe:1b
-period:0D01                                    
-periodstokeep:4
+datastripe:0b
```
emansanti, line 34: why is datastripe set to 0b here?

tests/tailer/access/appconfig/filtermap.csv
```diff
@@ -0,0 +1,3 @@
+wcRef,filter
```
emansanti, line 1: Is this file needed for unit tests?

config/tablekeycols.csv
```diff
@@ -0,0 +1,6 @@
+tablename,keycol
+packets,
+quote,sym
+quote_iex,
+trade,sym
+trade_iex,
```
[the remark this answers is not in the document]
CReid98: that's the mapping of tablenames to keycolumns

tests/tailer/access/appconfig/filtermap.csv
CReid98: this was used in an older version

tests/tailer/access/tests.csv
```diff
@@ -0,0 +1,15 @@
...
+before,0,0,q,.rdb.subscribe[],1,,"Subscribe to all data from stp"
+before,0,0,q,system "sleep 5",1,,"Wait for tables to populate"
+before,0,0,q,rdbh:gethandle[`rdb1],1,,"Get rdb handle"
+before,0,0,q,stph:gethandle[`stp1],1,,"Get stp handle"
+true,0,0,q,.ds.datastripe=1b,1,,"Check datastriping is set to 1"
+true,0,0,q,2=count .wdb.access,1,,"Check length of access table in wdb"
+true,0,0,q,rdbh"2=count .rdb.access",1,,"Check length of access table in rdb"
+run,0,0,q,stph".stplg.endofperiod[.z.d+11:50;.z.d+12:00;.stplg.endofdaydata]",1,,"Run end of period function from stp"
+true,0,0,q,.wdb.access[`quote][`end]=.z.d+12:00,1,,"check end time in wdb access table"
+true,0,0,q,rdbh".rdb.access[`quote][`start]>=.z.d+11:20",1,,"check start time in rdb access table"
```
emansanti, line 12: Are these checks on the start and end time on rdb and wdb checked from the data values in the accesstables?

code/wdb/datastripe.q
emansanti: I'm not sure where I was going with this either now

PR #456, opened 2022-05-18
Tcs 112 combining all configuration for segmenting into one table in the stp and update any functions that previously read from the multiple tables to run of one table

AdamBrown2:
added striping.json with two empty segments to show how the json is formatted
updated datastripe.q to work with the json
added loggin to configload in datastripe.q so that if subscriptiondefault is not defined or is defined as anything but "ignore" or "all" the tickerplant will error
when .ds.datastripe is enabled .rdb.subtables is updated to allow the reload function to run. This is only temporary fix until the root of problem is found

### DIFF
--- a/code/processes/rdb.q
+++ b/code/processes/rdb.q
@@ -234,7 +234,4 @@ $[.rdb.connectonstart;
   (`.rdb.timeoutreset;`);"Set rdb timeout to 0 for EOD writedown"];
 
 //Updating .rdb.subtables to only include tables that actually subscribed
-//if[.ds.datastripe;
-//  .rdb.subtables:.rdb.subtables[where 98h=type each value each .rdb.subtables];
-//  ];
 .rdb.subtables:$[.ds.datastripe;tables[] except `heartbeat;.rdb.subtables];
--- a/code/processes/rdb.q
+++ b/code/processes/rdb.q
@@ -234,6 +234,7 @@ $[.rdb.connectonstart;
   (`.rdb.timeoutreset;`);"Set rdb timeout to 0 for EOD writedown"];
 
 //Updating .rdb.subtables to only include tables that actually subscribed
-if[.ds.datastripe;
-  .rdb.subtables:.rdb.subtables[where 98h=type each value each .rdb.subtables];
-  ];
+//if[.ds.datastripe;
+//  .rdb.subtables:.rdb.subtables[where 98h=type each value each .rdb.subtables];
+//  ];
+.rdb.subtables:$[.ds.datastripe;tables[] except `heartbeat;.rdb.subtables];
--- a/code/processes/rdb.q
+++ b/code/processes/rdb.q
@@ -232,3 +232,8 @@ $[.rdb.connectonstart;
 /-GMT offset rounded to nearest 15 mins and added to roll time
 .timer.repeat[.eodtime.nextroll-00:01+{00:01*15*"j"$(`minute$x)%15}(.proc.cp[]-.z.p);0W;1D;
   (`.rdb.timeoutreset;`);"Set rdb timeout to 0 for EOD writedown"];
+
+//Updating .rdb.subtables to only include tables that actually subscribed
+if[.ds.datastripe;
+  .rdb.subtables:.rdb.subtables[where 98h=type each value each .rdb.subtables];
+  ];
--- a/code/segmentedtickerplant/datastripe.q
+++ b/code/segmentedtickerplant/datastripe.q
@@ -6,11 +6,11 @@
 
 configload:{
      scpath:first .proc.getconfigfile[string .ds.stripeconfig];
-     {if[()~key hsym x;.lg.e[`init;"The following file can not be found: ",string scpath]]};
-     .stpps.stripeconfig:.j.k raze read0 scpath;
+     if[()~key hsym scpath;.lg.e[`init;"The qollowing file can not be found: ",string scpath]];
+     .stpps.stripeconfig:.j.k read1 scpath;
      defaults:{first (flip .stpps.stripeconfig[x])[`subscriptiondefault]}each key .stpps.stripeconfig;
      errors:1+ where {[x] not ("ignore"~x) or ("all"~x)}each defaults;
-     {if[0<count errors;.lg.e[`sub;m:"subscriptiondefault not defined for segment/segments ",string[x]," "]]}each errors;
+     {if[0<count x;.lg.e[`sub;m:"subscriptiondefault not defined as \"ignore\" or \"all\" for segment ",string[x]," "]]}each errors;
      };
 
 initdatastripe:{
@@ -22,11 +22,9 @@ initdatastripe:{
 \d .stpps
 
 //makes a dictionary of tables and there filters for segmentedsubdetails
-
 filtermap:{[tabs;id] if[tabs~`;tabs:.stpps.t]; ((),tabs)!.stpps.segmentfilter\:[(),tabs;id]}
 
 //grabs filters from the config files and for the "ignoretable" filter converts to "" to allow segmentedsudetails to run
-
 segmentfilter:{[tbl;segid]
      id:`$string segid;
      filter:first (flip .stpps.stripeconfig[id])[tbl];
@@ -45,19 +43,18 @@ subsegment:{[tbl;segid];
      if[tbl~`;:.z.s[;segid] each .stpps.t];
      stripedtables:inter [key flip .stpps.stripeconfig[id];.stpps.t];
 //if the defualt is "all" tables not mentioned in striping.json will be subscribed unfiltered
-     if[default~"all";suballtabs: except[.stpps.t;stripedtables];
+     if[default~"all";suballtabs: .stpps.t except stripedtables;
      .lg.o[`sub;m:"Table ",string[tbl]," is to be subscribed unfiltered for segment ",string[segid],""]];
 //if default is ignore creates a list to of tables to ignore
-     if[default~"ignore"; ignoredtables: except[.stpps.t;stripedtables]];
+     if[default~"ignore"; ignoredtables: .stpps.t except stripedtables];
      filter:.stpps.segmentfilter[tbl;segid];
 //for case when filter is "ignoretable" adds that table to ignoredtables list
      if[(first (flip .stpps.stripeconfig[id])[tbl])~"ignoretable";ignoredtables:ignoredtables,tbl];
      if[tbl in ignoredtables;
       .lg.o[`sub;m:"Table ",string[tbl]," is to be ignored for segment ",string[segid],""];
- :()];
+      :()];
      .ps.subtablefiltered[string[tbl];filter;""]
      };
-
 
 \d .
 
--- a/code/segmentedtickerplant/datastripe.q
+++ b/code/segmentedtickerplant/datastripe.q
@@ -1,5 +1,8 @@
+//Allows config file to be overwritten in process.csv
+
 .ds.stripeconfig:@[value;`.ds.stripeconfig;`striping.json];
 
+//Loads the striping.json config file checks if each subscriptiondefault is set for each segment and errors if not defined
 
 configload:{
      scpath:first .proc.getconfigfile[string .ds.stripeconfig];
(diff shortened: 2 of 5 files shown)


code/segmentedtickerplant/datastripe.q
```diff
@@ -1,60 +1,64 @@
...
+//Loads the striping.json config file checks if each subscriptiondefault is set for each segment and errors if not defined
 
 configload:{
-     scpath:first .proc.getconfigfile[string .ds.segmentconfig];
-     fmpath:first .proc.getconfigfile[string .ds.filtermap];
-     {if[()~key hsym x;.lg.e[`init;"The following file can not be found: ",string x]]} each (scpath;fmpath);
-     @[{.stpps.segmentconfig:("SIS";enlist",")0: hsym x};scpath;{.lg.e[`init;"Failure in loading ",string y]}[;scpath]];
-     @[{.stpps.segmentfiltermap:(!/)(("S*";enlist",")0: hsym x)`wcRef`filter};fmpath;{.lg.e[`init;"Failure in loading ",string y]}[;fmpath]];
+     scpath:first .proc.getconfigfile[string .ds.stripeconfig];
+     {if[()~key hsym x;.lg.e[`init;"The following file can not be found: ",string scpath]]};
```
jonathonmcmurray, line 9: I don't think this should be a lambda (atm it won't execute)

```suggestion
     if[()~key hsym scpath;.lg.e[`init;"The following file can not be found: ",string scpath]];
```

code/segmentedtickerplant/datastripe.q
```diff
@@ -1,60 +1,64 @@
...
+     .stpps.stripeconfig:.j.k raze read0 scpath;
```
jonathonmcmurray, line 10:
```suggestion
     .stpps.stripeconfig:.j.k read1 scpath;
```

you can skip the `raze` if you read the file as bytes with `read1` & `.j.k` will still parse it

code/segmentedtickerplant/datastripe.q
```diff
@@ -1,60 +1,64 @@
...
+     defaults:{first (flip .stpps.stripeconfig[x])[`subscriptiondefault]}each key .stpps.stripeconfig;
+     errors:1+ where {[x] not ("ignore"~x) or ("all"~x)}each defaults;
+     {if[0<count errors;.lg.e[`sub;m:"subscriptiondefault not defined for segment/segments ",string[x]," "]]}each errors;
```
jonathonmcmurray, line 13: I don't think `errors` is defined within this lambda, so I don't think this will work as-is. Might be simpler to just combine into one error message something like

```suggestion
     if[0<count errors;.lg.e[`sub;m:"subscriptiondefault not defined for segment(s) ",", " sv string[errors]];
```

(also don't think the space after the segment ids is needed, removed here)

code/segmentedtickerplant/datastripe.q
```diff
@@ -1,60 +1,64 @@
...
      .lg.o[`init;"init datastriping"];
      configload[];
      };
 
+
 \d .stpps
 
-// Function to map the where clause from config table extracted by .stpps.segmentfilter function to tablename
-// Allows use of ` as argument for tables
+//makes a dictionary of tables and there filters for segmentedsubdetails
```
jonathonmcmurray, line 24:
```suggestion
//makes a dictionary of tables and their filters for segmentedsubdetails
```

code/segmentedtickerplant/datastripe.q
```diff
@@ -1,60 +1,64 @@
...
+
```
jonathonmcmurray, line 25:
```suggestion
```

get rid of blank line between comment and what it's describing

code/segmentedtickerplant/datastripe.q
```diff
@@ -1,60 +1,64 @@
...
 filtermap:{[tabs;id] if[tabs~`;tabs:.stpps.t]; ((),tabs)!.stpps.segmentfilter\:[(),tabs;id]}
 
-// Find where clause from config tables
+//grabs filters from the config files and for the "ignoretable" filter converts to "" to allow segmentedsudetails to run
+
```
jonathonmcmurray, line 29:
```suggestion
```

code/segmentedtickerplant/datastripe.q
```diff
@@ -1,60 +1,64 @@
...
+     id:`$string segid;
+     filter:first (flip .stpps.stripeconfig[id])[tbl];
+     $[filter~"ignoretable";filter:"";filter]
      };
 
-// Subscribe to particular segment using segmentID based on .u.sub
+//subscribe to a table using segmentID to determine filtering
 subsegment:{[tbl;segid];
-     //tablename and segmentid used to get filters
+//casting segid to an symbol as json is restrictive
```
jonathonmcmurray, line 38:
```suggestion
     //casting segid to an symbol as json is restrictive
```

match indentation of the code

code/segmentedtickerplant/datastripe.q
```diff
@@ -1,60 +1,64 @@
...
+     id:`$string segid;
+     if[not (id in (key .stpps.stripeconfig));
+       .lg.e[`sub;m:"Segment ",string[segid]," is not defined in striping.json"];:()];
+     ignoredtables:`$();
+//setting the default for non-configured tables
```
jonathonmcmurray, line 43:
```suggestion
     //setting the default for non-configured tables
```

code/segmentedtickerplant/datastripe.q
```diff
@@ -1,60 +1,64 @@
...
-          :();
-     ];
-     filter:segmentfilter[tbl;segid];
-     if[filter~"";
-          .lg.e[`sub;m:"Incorrect pairing of table ",string[tbl]," and segmentID ",string[segid]," not found in .stpps.segmentconfig"];
-          :();
-     ];
-     
+     stripedtables:inter [key flip .stpps.stripeconfig[id];.stpps.t];
+//if the defualt is "all" tables not mentioned in striping.json will be subscribed unfiltered
```
jonathonmcmurray, line 47:
```suggestion
     //if the default is "all" tables not mentioned in striping.json will be subscribed unfiltered
```

code/segmentedtickerplant/datastripe.q
```diff
@@ -1,60 +1,64 @@
...
+     if[default~"all";suballtabs: except[.stpps.t;stripedtables];
+     .lg.o[`sub;m:"Table ",string[tbl]," is to be subscribed unfiltered for segment ",string[segid],""]];
+//if default is ignore creates a list to of tables to ignore
```
jonathonmcmurray, line 50:
```suggestion
     //if default is ignore creates a list to of tables to ignore
```

code/segmentedtickerplant/datastripe.q
```diff
@@ -1,60 +1,64 @@
...
+     if[default~"ignore"; ignoredtables: except[.stpps.t;stripedtables]];
+     filter:.stpps.segmentfilter[tbl;segid];
+//for case when filter is "ignoretable" adds that table to ignoredtables list
```
jonathonmcmurray, line 53:
```suggestion
     //for case when filter is "ignoretable" adds that table to ignoredtables list
```

code/segmentedtickerplant/datastripe.q
```diff
@@ -1,60 +1,64 @@
...
-          .lg.e[`sub;m:"Table ",string[tbl]," not in list of stp pub/sub tables"];
-          :();
-     ];
-     filter:segmentfilter[tbl;segid];
-     if[filter~"";
-          .lg.e[`sub;m:"Incorrect pairing of table ",string[tbl]," and segmentID ",string[segid]," not found in .stpps.segmentconfig"];
-          :();
-     ];
-     
+     stripedtables:inter [key flip .stpps.stripeconfig[id];.stpps.t];
```
jonathonmcmurray, line 46:
```suggestion
     stripedtables:.stpps.t inter key flip .stpps.stripeconfig[id];
```

personal preference, but IMO the infix notation is more readable in this case?

code/segmentedtickerplant/datastripe.q
```diff
@@ -1,60 +1,64 @@
...
+//if the defualt is "all" tables not mentioned in striping.json will be subscribed unfiltered
+     if[default~"all";suballtabs: except[.stpps.t;stripedtables];
```
jonathonmcmurray, line 48:
```suggestion
     if[default~"all";suballtabs:.stpps.t except stripedtables;
```

code/segmentedtickerplant/datastripe.q
```diff
@@ -1,60 +1,64 @@
...
+     .lg.o[`sub;m:"Table ",string[tbl]," is to be subscribed unfiltered for segment ",string[segid],""]];
+//if default is ignore creates a list to of tables to ignore
+     if[default~"ignore"; ignoredtables: except[.stpps.t;stripedtables]];
```
jonathonmcmurray, line 51:
```suggestion
     if[default~"ignore"; ignoredtables:.stpps.t except stripedtables];
```

code/segmentedtickerplant/datastripe.q
```diff
@@ -1,60 +1,64 @@
...
+     filter:.stpps.segmentfilter[tbl;segid];
+//for case when filter is "ignoretable" adds that table to ignoredtables list
+     if[(first (flip .stpps.stripeconfig[id])[tbl])~"ignoretable";ignoredtables:ignoredtables,tbl];
+     if[tbl in ignoredtables;
+      .lg.o[`sub;m:"Table ",string[tbl]," is to be ignored for segment ",string[segid],""];
+ :()];
```
jonathonmcmurray, line 57:
```suggestion
      :()];
```

code/segmentedtickerplant/datastripe.q
```diff
@@ -1,60 +1,64 @@
...
      .ps.subtablefiltered[string[tbl];filter;""]
      };
 
+
```
jonathonmcmurray, line 61:
```suggestion
```

code/segmentedtickerplant/datastripe.q
```diff
@@ -1,60 +1,64 @@
...
+     filter:first (flip .stpps.stripeconfig[id])[tbl];
+     $[filter~"ignoretable";filter:"";filter]
      };
 
-// Subscribe to particular segment using segmentID based on .u.sub
+//subscribe to a table using segmentID to determine filtering
 subsegment:{[tbl;segid];
-     //tablename and segmentid used to get filters
+//casting segid to an symbol as json is restrictive
+     id:`$string segid;
```
jonathonmcmurray, line 39: I think we're doing this because the keys in a dict parsed from a JSON object will always be symbols, right? Should we just have it that segment IDs _are_ symbols, rather than numbers? i.e. in this function, expect a symbol to be passed rather than a number

Potentially this allows a bit more flexibility in giving segments meaningful names. e.g. if you had a segment that filtered all tables to ``exch=`LSE``, you could name that segment `` `LSE`` rather than having to give it a number

Just a thought, not sure if there are other places in the code were changes would have to be made to accomodate a symbol based segment name

jonathonmcmurray (commented):
mostly minor cosmetic things

code/segmentedtickerplant/datastripe.q
AdamBrown2: when i take it the {} the stp doenst spin up

AdamBrown2: The x was the issue thanks

code/segmentedtickerplant/datastripe.q
AdamBrown2: Fixed the issue with the errors couldn't get your way to work but ill keep trying at it. pushing a working version of it

code/processes/rdb.q
```diff
@@ -232,3 +232,8 @@ $[.rdb.connectonstart;
 /-GMT offset rounded to nearest 15 mins and added to roll time
 .timer.repeat[.eodtime.nextroll-00:01+{00:01*15*"j"$(`minute$x)%15}(.proc.cp[]-.z.p);0W;1D;
   (`.rdb.timeoutreset;`);"Set rdb timeout to 0 for EOD writedown"];
+
+//Updating .rdb.subtables to only include tables that actually subscribed
+if[.ds.datastripe;
+  .rdb.subtables:.rdb.subtables[where 98h=type each value each .rdb.subtables];
```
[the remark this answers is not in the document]
AdamBrown2: Made the change to that format. Also I think I found the root of the issue in subscriptions.q line 144.
``` d:(`subtables`tplogdate!(details[`schemalist][;0];(first "D" $ -10 sublist string last first details[`logfilelist])^logdate));```
For standard TorQ subtables grabs the tables from the schemalist from details but now with the functionality to ignore tables the schema will be null.
So either these nulls need to be ignored when subtables is being created or ignore the nulls when segmentedsubdetails defines its schemalist to only contain schemas for tables we want to be subscribed to.

code/segmentedtickerplant/datastripe.q
AdamBrown2: Yes no problem do not how i keep missing them

drgdavies (approved):

PR #458, opened 2022-05-20
TCS-65 

0Cal-MacK0:
TorQ has the concept of "process attributes" - this needs to be extended to allow for efficient query routing when data is distributed across multiple processes in a segmented setup.


CReid98 (commented):
tests/metadata/settings.q

is this empty file needed?

0Cal-MacK0:
> tests/metadata/settings.q
> 
> is this empty file needed?

`settings.q` and the reference to it in `run.sh` have been removed as they are redundant.

CReid98 (approved):

code/hdb/hdbstandard.q
```diff
@@ -4,4 +4,28 @@ reload:{
 	system"l ."}
 
 // Get the relevant HDB attributes
-.proc.getattributes:{`date`tables!(@[value;`date;`date$()];tables[])}
+\d .proc
+getattributes:{
+	timecolumns:1!except[(::),gettimecolumns each t:tables[`.];(::)];
+	attrtable:`date`tables`procname!(
```
jonathonmcmurray, line 10: nitpick: this variable name states that it's a table, but it's a dict

code/hdb/hdbstandard.q
```diff
@@ -4,4 +4,28 @@ reload:{
 	system"l ."}
 
 // Get the relevant HDB attributes
-.proc.getattributes:{`date`tables!(@[value;`date;`date$()];tables[])}
+\d .proc
+getattributes:{
+	timecolumns:1!except[(::),gettimecolumns each t:tables[`.];(::)];
```
jonathonmcmurray, line 9: why is the generic null stuff required here? (assuming it is necessary, could do with a comment above to clarify reason I think)

code/hdb/hdbstandard.q
```diff
@@ -4,4 +4,28 @@ reload:{
...
+	attrtable:`date`tables`procname!(
+		/ date reporting makes some assumptions
```
jonathonmcmurray, line 11: this comment seems a bit vague - what are the assumptions?

code/hdb/hdbstandard.q
```diff
@@ -4,4 +4,28 @@ reload:{
...
+		@[value;`date;`date$()] union first[d]+til 1+last deltas d:exec(min;max)@\:distinct`date$last each raze value each timecolumns from timecolumns;
+		tables[];
+		.proc.procname);
+	attrtable[`dataaccess]:`segid`tablename!(first .ds.segmentid;(key timecolumns)!select wc,timecolumns from update wc:{""}'[i],timecolumns from timecolumns);
```
jonathonmcmurray, line 15: why is this added to the `attrtable` dictionary separately when the other three keys are all added together above? (I prefer this one as it has the key on the same line as the value so it's clearer IMO, but would be best to keep things consistent - ideally change the other keys to match this)

code/hdb/hdbstandard.q
```diff
@@ -4,4 +4,28 @@ reload:{
...
+    	attrtable}
+// Get temporal cols and the time windows they contain
+gettimecolumns:{
+	tcols:exec c from meta t:$[`date in cols x;
+    	/ select table using max date if date col exists
+	select from x where date=max date;
+                value x] where t in "pdz";
```
jonathonmcmurray, line 22: this layout is really confusing IMO. I think you should define t on a separate line first e.g.

```suggestion
    / select table using max date if date col exists
    t:$[`date in cols x;select from x where date=max date;value x];
    tcols:exec c from meta t where t in "pdz";
```

(with appropriate indentation etc.)

code/hdb/hdbstandard.q
```diff
@@ -4,4 +4,28 @@ reload:{
...
+	if[not count tcols;:(::)];
```
jonathonmcmurray, line 23: do you need to return a generic null? more common would be to return an empty list i.e.

```suggestion
	if[not count tcols;:()];
```

code/hdb/hdbstandard.q
```diff
@@ -4,4 +4,28 @@ reload:{
...
+        / functional exec to get the max value within each timebased column (defaults to -1+`timestamp$.z.d)
+	d:?[t;();();
+		tcols!enlist,/:-0Wp,/:
+				enlist each(?),/:(enlist each(in[;(-0Wd;-0Wp;-0wz)]),/:mtcols),'(enlist(+;-1;($;ets;`.z.d))),/:enlist each($;ets:enlist`timestamp),/:mtcols:enlist each max,/:tcols];
```
jonathonmcmurray, line 27: weird indentation here and this also looks very complex - could you include in a comment what the equivalent q-SQL would be?

code/rdb/rdbstandard.q
```diff
@@ -1,6 +1,32 @@
...
+getattributes:{
+    timecolumns:1!except[(::),gettimecolumns each tables[`.];(::)];
+    attrtable:`date`tables`procname!
+	({[timecolumns]
+	  d:exec(min;max)@\:distinct`date$first each raze value each timecolumns from timecolumns;
+	  first[d]+til 1+last deltas d}[timecolumns];
+	tables[];
+	.proc.procname);
+    if[.ds.datastripe;
+	filtermap:flip @[.j.k read1 hsym`$getenv[`KDBAPPCONFIG],"/striping.json";`$string first .ds.segmentid];
```
jonathonmcmurray, line 12: is this not already read into some global variable or something? (possibly sent by TP when subscription occurs?) feels strange to read from an on-disk json when this function is called (presumably the GW will be calling this periodically to get the up-to-date attributes etc.)

code/rdb/rdbstandard.q
```diff
@@ -1,6 +1,32 @@
...
+	filtermap:([]tablename:key filtermap)!([]wc:value filtermap);
+	/- Match tables (and their timewindows) against the whereclause they've used to subscribe
+	tablename:((key timecolumns)!select wc,timecolumns from update wc:{""}'[i],timecolumns from timecolumns)
+		    lj
+                      filtermap;
+	dataaccess:enlist[`dataaccess]!enlist`segid`tablename!(first .ds.segmentid;tablename);
+        attrtable,:dataaccess;
```
jonathonmcmurray, line 19: why not

```suggestion
	attrtable[`dataaccess]:`segid`tablename!(first .ds.segmentid;tablename);
```

or if that doesn't work for some reason

```suggestion
	attrtable,:enlist[`dataaccess]!enlist`segid`tablename!(first .ds.segmentid;tablename);
```

code/rdb/rdbstandard.q
```diff
@@ -1,6 +1,32 @@
...
+	  first[d]+til 1+last deltas d}[timecolumns];
+	tables[];
+	.proc.procname);
+    if[.ds.datastripe;
+	filtermap:flip @[.j.k read1 hsym`$getenv[`KDBAPPCONFIG],"/striping.json";`$string first .ds.segmentid];
+	filtermap:([]tablename:key filtermap)!([]wc:value filtermap);
+	/- Match tables (and their timewindows) against the whereclause they've used to subscribe
+	tablename:((key timecolumns)!select wc,timecolumns from update wc:{""}'[i],timecolumns from timecolumns)
+		    lj
+                      filtermap;
```
jonathonmcmurray, line 17:
```suggestion
	tablename:((key timecolumns)!select wc,timecolumns from update wc:{""}'[i],timecolumns from timecolumns) lj filtermap;
```

don't think this is long enough to justify three lines

code/rdb/rdbstandard.q
```diff
@@ -1,6 +1,32 @@
...
+    timecolumns:1!except[(::),gettimecolumns each tables[`.];(::)];
+    attrtable:`date`tables`procname!
+	({[timecolumns]
+	  d:exec(min;max)@\:distinct`date$first each raze value each timecolumns from timecolumns;
+	  first[d]+til 1+last deltas d}[timecolumns];
+	tables[];
+	.proc.procname);
+    if[.ds.datastripe;
+	filtermap:flip @[.j.k read1 hsym`$getenv[`KDBAPPCONFIG],"/striping.json";`$string first .ds.segmentid];
+	filtermap:([]tablename:key filtermap)!([]wc:value filtermap);
```
jonathonmcmurray, line 13: I think you're doing this so that you can use `lj` below, but maybe that's unnecessary?

You should be able to use the dictionary directly something like

```q
update wc:filtermap tablename from timecolumns
```

jonathonmcmurray (commented):
in addition to other comments, make sure that indentation is consistent - atm there is a mix of spaces and tabs, stick to one or other for indenting and be consistent with what's already in the file you're editing if possible

PR #462, opened 2022-06-10
TCS-117-changing-wdb-to-tailer-process

dmitchard:
change tailer to save each tail as IDB+segid rather than procname

### DIFF
--- a/code/wdb/datastripe.q
+++ b/code/wdb/datastripe.q
@@ -33,7 +33,7 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
 	/- get unique sym from table
 	s:first raze value'[?[enumdata;();1b;enlist[keycol]!enlist keycol]];
 	/- get process specific taildir location
-	dir:` sv dir,.proc.procname,`;
+	dir:` sv .ds.td,(`$raze string `IDB,.ds.segmentid),`;
 	/- get symbol enumeration
 	partitionint:`$string (where s=value [`.]`sym)0;
 	/- create directory location for selected partition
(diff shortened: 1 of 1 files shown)


SeanW17 (approved):
Will need to be updated if segid changes format in the future.

SeanW17 (approved):

code/wdb/datastripe.q
```diff
@@ -33,7 +33,7 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
 	/- get unique sym from table
 	s:first raze value'[?[enumdata;();1b;enlist[keycol]!enlist keycol]];
 	/- get process specific taildir location
-	dir:` sv dir,.proc.procname,`;
+	dir:` sv .ds.td,(`$raze string `TDB,.ds.segmentid),`;
```
jonathonmcmurray, line 36:
```suggestion
	dir:` sv .ds.td,(`$"TDB",string .ds.segmentid),`;
```

think this will be the same but less messy

is there a reason to switch away from using procname? also this no longer uses the `dir` that gets passed into function, is that intentional? should that arg be removed?

dmitchard: Sorry, it should be using dir instead of .ds.td. 
Your line to create the directory however creates a list of syms like so:
``q)`$"TDB",string .ds.segmentid``
 `` `T`D`B`1``

dmitchard: It is a long

dmitchard: It is a list of longs
``q)type .ds.segmentid``
``7h``

jonathonmcmurray: why is it a list? surely there's only one id for a given process? seems strange

PR #464, opened 2022-06-10
Tcs 115 access table create and update final

CReid98:


SeanW17 (approved):

CReid98:
Following a comment from Jonathon McMurray on TCS106 I checked that when the keycol was left blank or was some other value the savedown and partition functions would default to sym or use the defined keycol value accordingly.

drgdavies (approved):

PR #466, opened 2022-06-15
bugfix to ensure data is saved down with the correct directory structure

CReid98:
Ensuring directory where data is saved down to has the correct structure, correcting a mistake when running the upserttopartition function which was made when merging with the previous commit

### DIFF
--- a/code/wdb/datastripe.q
+++ b/code/wdb/datastripe.q
@@ -85,13 +85,13 @@ savetablesoverperiod:{[dir;tablename;nextp;lasttime]
     partitionlist: ?[tablename;();();(distinct;keycol)];
 
     /- enumerate and then split by keycol
-    symdir:` sv .ds.td,.proc.procname;
+    symdir:` sv dir,.proc.procname;
     enumkeycol: .Q.en[symdir;?[tablename;enlist (<;`time;nextp);0b;()]];
     splitkeycol: {[enumkeycol;keycol;s] ?[enumkeycol;enlist (=;keycol;enlist s);0b;()]}[enumkeycol;keycol;] each partitionlist;
 
     /-upsert table to partition
     upserttopartition[dir;tablename;keycol;;nextp] each splitkeycol;
-    
+
     /- delete data from last period
     .[.ds.deletetablebefore;(tablename;`time;lasttime)];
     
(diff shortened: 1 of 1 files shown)


code/wdb/datastripe.q
```diff
@@ -85,12 +85,13 @@ savetablesoverperiod:{[dir;tablename;nextp;lasttime]
     partitionlist: ?[tablename;();();(distinct;keycol)];
 
     /- enumerate and then split by keycol
-    enumkeycol: .Q.en[dir;?[tablename;enlist (<;`time;nextp);0b;()]];
+    symdir:` sv .ds.td,.proc.procname;
```
jonathonmcmurray, line 88: should this be `.ds.td` or `dir`? (I see `dir` is passed to the function & used below in call to `upserttopartition`, just want to make sure this is correct)

code/wdb/datastripe.q
```diff
@@ -85,12 +85,13 @@ savetablesoverperiod:{[dir;tablename;nextp;lasttime]
...
+    enumkeycol: .Q.en[symdir;?[tablename;enlist (<;`time;nextp);0b;()]];
     splitkeycol: {[enumkeycol;keycol;s] ?[enumkeycol;enlist (=;keycol;enlist s);0b;()]}[enumkeycol;keycol;] each partitionlist;
 
     /-upsert table to partition
-    @[upserttopartition[dir;tablename;keycol;;nextp;partitionlist] ; splitkeycol ; ];
-
+    upserttopartition[dir;tablename;keycol;;nextp] each splitkeycol;
+
```
jonathonmcmurray, line 94: remove added whitespace

```suggestion

```

code/wdb/datastripe.q
jonathonmcmurray: I just wanted to clarify this - if it was correct as it was, that's fine. I don't know the context enough to know which is correct

CReid98: It should have been dir to ensure that the directory structure is consistent even if you pass in a different directory when running the savedown. I think it's unlikely that this would come up but it did need corrected.

jonathonmcmurray (approved):
LGTM

SeanW17 (approved):

PR #468, opened 2022-06-17
Extend merge functions to tickerlog

eointeague:
Extending new merge functionality to tickerlogreplay process.
- Merge function was being called after every table was replayed, this was causing issues for replaying empty tables. So tables will be merged after all logs have been replayed and data saved to temporary storage.
- Have added the option to merge by a bytesize estimate rather than just row counts - set by .merge.mergebybytesizelimit
- New mergefunctions can be configured by setting .replay.mergemethod - default is part
- updated config
- added unit tests


config/settings/tickerlogreplay.q
```diff
@@ -1,5 +1,7 @@
 // default configuration for the tickerplant replay
 
+.merge.mergebybytelimit:1b              // merge limit configuration - default is 0b row count limit 1b is byte size limit
```
jonnypress, line 3: I think the default here should be 0b

jonnypress (approved):

jonnypress:
Looks good to me, please squash and merge

PR #469, opened 2022-06-21
Add bug fix to .z.exit

ThomasGillespie:
There was an issue where on set up of a segmented tickerplant, if an error was thrown an infinite loop was caused. When an error was thown, .z.exit called another error, calling the exit function again.

### DIFF
--- a/code/segmentedtickerplant/stplog.q
+++ b/code/segmentedtickerplant/stplog.q
@@ -299,6 +299,7 @@ init:{[dbname]
 
 // Close logs on clean exit
 .z.exit:{
+  .proc.initialised:1b;
   if[not x~0i;.lg.e[`stpexit;"Bad exit!"];:()];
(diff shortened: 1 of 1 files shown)


SeanW17 (approved):

dmitchard (approved):

code/segmentedtickerplant/stplog.q
```diff
@@ -299,6 +299,7 @@ init:{[dbname]
 
 // Close logs on clean exit
 .z.exit:{
+  .proc.initialised:1b;
```
jonathonmcmurray, line 302: Can you add an explanatory comment above this line? It may not be obvious why this is necessary when someone else looks at this function

SeanW17 (approved):

dmitchard (approved):

PR #481, opened 2022-07-11
Tcs 38 providing metadata for gateway routing

CReid98:
creating access table in gateway which is updated when the endofperiod function is run in the rdb and wdb and contains all information from rdb and wdb


code/gateway/datastripe.q
```diff
@@ -0,0 +1,25 @@
+\d .ds
+
+// create a function which will retrieve the access tables from the subscribers
```
0Cal-MacK0, line 3:
```suggestion
// retrieve the access tables from the subscribers
```

drgdavies (approved):

SeanW17 (approved):

PR #482, opened 2022-07-12
Minor Bug fixes

ThomasGillespie:
Added quotations around $QCMD in if on line 21, solves common "too many arguments" error

Added where the port is found for summary, added -w flag to grep, this makes it less likely to have multiple ports but not impossible, Lowers the number of ports shown when this does happen

### DIFF
--- a/torq.sh
+++ b/torq.sh
@@ -18,7 +18,7 @@ if [ -z $QCON ]; then
   QCON="qcon"                                                                                       # set default value if not already defined
 fi
 
-if [ -z "$QCMD" ]; then                                                                               # set default value if not already defined
+if [ -z "$QCMD" ]; then                                                                             # set default value if not already defined
   QCMD="q"
 fi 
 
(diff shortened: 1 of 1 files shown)


torq.sh
```diff
@@ -18,7 +18,7 @@ if [ -z $QCON ]; then
   QCON="qcon"                                                                                       # set default value if not already defined
 fi
 
-if [ -z $QCMD ]; then                                                                               # set default value if not already defined
+if [ -z "$QCMD" ]; then                                                                               # set default value if not already defined
```
jonathonmcmurray, line 21: re-align comment

SeanW17 (approved):

hobbes8548:
@ThomasGillespie Since you discovered this issue already, do you mind applying the same double-quotes fix to lines 9, 13, 17, and 71. The same problem may occur and have impacted me locally as well. Thanks!

PR #484, opened 2022-07-15
Tcs 116 config checks

SeanW17:
Most of the code changes are for the config check. Subscription change is for the wdbs and rdbs.

### DIFF
--- a/code/common/subscriptions.q
+++ b/code/common/subscriptions.q
@@ -107,9 +107,8 @@ subscribe:{[tabs;instrs;setschema;replaylog;proc]
   if[null tptype; .lg.e[`subscribe;e:"could not determine tickerplant type"]; 'e];
  
   // datastriping check on the tickerplant
-  $[.ds.datastripe:@[proc`w;({@[value;`.ds.datastripe;0b]};`);`];
-   .lg.o[`subscribe;"datastriping is turned on"];
-   .lg.o[`subscribe;"datastriping is turned off"]];
+  .ds.datastripe:@[proc`w;({@[value;`.ds.datastripe;0b]};`);`];
+  .lg.o[`subscribe;"datastriping is turned ",$[.ds.datastripe;"on";"off"]];
  
   // depending on the type of tickerplant being subscribed to, change the functions for requesting
   // the tables and subscriptions
--- a/code/common/subscriptions.q
+++ b/code/common/subscriptions.q
@@ -106,6 +106,11 @@ subscribe:{[tabs;instrs;setschema;replaylog;proc]
   tptype:@[proc`w;({@[value;`tptype;`standard]};`);`];
   if[null tptype; .lg.e[`subscribe;e:"could not determine tickerplant type"]; 'e];
  
+  // datastriping check on the tickerplant
+  $[.ds.datastripe:@[proc`w;({@[value;`.ds.datastripe;0b]};`);`];
+   .lg.o[`subscribe;"datastriping is turned on"];
+   .lg.o[`subscribe;"datastriping is turned off"]];
+ 
   // depending on the type of tickerplant being subscribed to, change the functions for requesting
   // the tables and subscriptions
   $[tptype=`standard;
--- a/code/processes/segmentedtickerplant.q
+++ b/code/processes/segmentedtickerplant.q
@@ -114,7 +114,10 @@ configcheck:{
      .lg.o[`configcheck;"initiate config check"];
      scpath:first .proc.getconfigfile[string .ds.stripeconfig];
      // Check striping.json file exists then runs configload
-     $[()~key hsym scpath; .lg.o[`configcheck;"Datastriping is not enabled as the following file can not be found: ",string .ds.stripeconfig]; configload[scpath]]
+     $[()~key hsym scpath;
+       .lg.o[`configcheck;"Datastriping is not enabled as the following file can not be found: ",string .ds.stripeconfig];
+       configload[scpath]
+     ];
      };
 
 // Have the init function and config check called from torq.q
--- a/code/processes/segmentedtickerplant.q
+++ b/code/processes/segmentedtickerplant.q
@@ -93,13 +93,13 @@ configload:{[scpath]
 // Checks if each subscriptiondefault is set for each segment and errors if not defined
 jsonchecks:{[]
      // check defaults are ignore or all
-     defaults:{first (flip .stpps.stripeconfig[x])[`subscriptiondefault]}each key .stpps.stripeconfig;
+     defaults:value first each .stpps.stripeconfig[;`subscriptiondefault];
      errors:1+ where {[x] not ("ignore"~x) or ("all"~x)}each defaults;
-     if[0<count errors;.lg.o[`jsonchecks;m:"subscriptiondefault not defined as \"ignore\" or \"all\" for segment(s) "," " sv string[errors]]];
+     if[0<count errors;.lg.o[`jsonchecks;"subscriptiondefault not defined as \"ignore\" or \"all\" for segment(s) "," " sv string[errors]]];
      // check for valid tables
-     keydict: key(flip .stpps.stripeconfig[`segid]);
-     stripedtables: .stpps.t inter keydict;
-     wrongtables:(keydict except `subscriptiondefault) except stripedtables;
+     configtables: key(flip .stpps.stripeconfig[`segid]);
+     stripedtables: .stpps.t inter configtables;
+     wrongtables:(configtables except `subscriptiondefault) except stripedtables;
      if[0<count wrongtables;.lg.o[`jsonchecks;"Table(s) ",(" " sv string[wrongtables])," not recognised"]];
      //Enable datastriping if all checks pass
      .ds.datastripe:min (0=count errors; 0=count wrongtables);
--- a/code/processes/segmentedtickerplant.q
+++ b/code/processes/segmentedtickerplant.q
@@ -100,7 +100,7 @@ jsonchecks:{[]
      keydict: key(flip .stpps.stripeconfig[`segid]);
      stripedtables: .stpps.t inter keydict;
      wrongtables:(keydict except `subscriptiondefault) except stripedtables;
-     {if[0<count x;.lg.o[`jsonchecks;m:"Table ",string[x]," is not recognised"]]}each wrongtables;
+     if[0<count wrongtables;.lg.o[`jsonchecks;"Table(s) ",(" " sv string[wrongtables])," not recognised"]];
      //Enable datastriping if all checks pass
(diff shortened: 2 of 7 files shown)


code/common/subscriptions.q
```diff
@@ -106,6 +106,11 @@ subscribe:{[tabs;instrs;setschema;replaylog;proc]
   tptype:@[proc`w;({@[value;`tptype;`standard]};`);`];
   if[null tptype; .lg.e[`subscribe;e:"could not determine tickerplant type"]; 'e];
  
+  // datastriping check on the tickerplant
+  $[.ds.datastripe:@[proc`w;({@[value;`.ds.datastripe;0b]};`);`];
+   .lg.o[`subscribe;"datastriping is turned on"];
+   .lg.o[`subscribe;"datastriping is turned off"]];
```
jonathonmcmurray, line 112: think it's simpler & clearer if you just make part of the logging conditional?

```suggestion
  .ds.datastripe:@[proc`w;({@[value;`.ds.datastripe;0b]};`);`];
  .lg.o[`subscribe;"datastriping is turned ",$[.ds.datastripe;"on";"off"]];
```

code/processes/segmentedtickerplant.q
```diff
@@ -83,5 +83,38 @@ init:{
...
-// Have the init function called from torq.q
+// Loads the striping.json config file and logs any problems
+configload:{[scpath]
+     @[jsonchecks; 
+       .stpps.stripeconfig:@[{.j.k read1 x};scpath;{.lg.o[`configload;"Failed to load in striping.json file: ",x]}];
+       {.lg.o[`configload;"Datastriping is not enabled due to problems with striping.json"]}]
+     };
+
+// Checks if each subscriptiondefault is set for each segment and errors if not defined
+jsonchecks:{[.stpps.stripeconfig]
```
jonathonmcmurray, line 94: this looks weird - is a variable being passed in here, or are we referrencing a global variable in the `.stpps` namespace?

if we're referrencing a global variable, just make this

```suggestion
jsonchecks:{[]
```

and pass a null or something when calling it above

code/processes/segmentedtickerplant.q
```diff
@@ -83,5 +83,38 @@ init:{
   .stplg.init[string .proc.procname];
  };
 
-// Have the init function called from torq.q
+// Loads the striping.json config file and logs any problems
+configload:{[scpath]
+     @[jsonchecks; 
+       .stpps.stripeconfig:@[{.j.k read1 x};scpath;{.lg.o[`configload;"Failed to load in striping.json file: ",x]}];
+       {.lg.o[`configload;"Datastriping is not enabled due to problems with striping.json"]}]
```
jonathonmcmurray, line 90: probably worth including the error?

```suggestion
       {.lg.o[`configload;"Datastriping is not enabled due to problems with striping.json: ",x]}]
```

code/processes/segmentedtickerplant.q
```diff
@@ -83,5 +83,38 @@ init:{
   .stplg.init[string .proc.procname];
  };
 
-// Have the init function called from torq.q
+// Loads the striping.json config file and logs any problems
+configload:{[scpath]
+     @[jsonchecks; 
+       .stpps.stripeconfig:@[{.j.k read1 x};scpath;{.lg.o[`configload;"Failed to load in striping.json file: ",x]}];
```
jonathonmcmurray, line 89: if we fail to read the json file, do we want to continue or should we just exit at that point? (assuming this happens during proc initialisation we can just log as an error I think & torq will exit the process automatically)

code/processes/segmentedtickerplant.q
```diff
@@ -83,5 +83,38 @@ init:{
...
+       {.lg.o[`configload;"Datastriping is not enabled due to problems with striping.json"]}]
+     };
+
+// Checks if each subscriptiondefault is set for each segment and errors if not defined
+jsonchecks:{[.stpps.stripeconfig]
+     // check defaults are ignore or all
+     defaults:{first (flip .stpps.stripeconfig[x])[`subscriptiondefault]}each key .stpps.stripeconfig;
```
jonathonmcmurray, line 96: I'm not 100% sure of the structure of `.stpps.stripeconfig` but can we do something like

```suggestion
     defaults:first each .stpps.stripeconfig[;`subscriptiondefault];
```

?

Maybe needs a flip in there somewhere or something, but I think it should be possible to at least avoid iterating over the keys

code/processes/segmentedtickerplant.q
```diff
@@ -83,5 +83,38 @@ init:{
...
+     errors:1+ where {[x] not ("ignore"~x) or ("all"~x)}each defaults;
```
jonathonmcmurray, line 97:
```suggestion
     errors:1+ where not defaults in\:("ignore";"all");
```
think this is the same

code/processes/segmentedtickerplant.q
```diff
@@ -83,5 +83,38 @@ init:{
...
+     {if[0<count x;.lg.o[`jsonchecks;m:"subscriptiondefault not defined as \"ignore\" or \"all\" for segment ",string[x]," "]]}each errors;
```
jonathonmcmurray, line 98: why iterate through them all? also I don't think `m` is used anywhere? and no need to have the space at the end of the log msg

```suggestion
     if[0<count errors;.lg.o[`jsonchecks;m:"subscriptiondefault not defined as \"ignore\" or \"all\" for segment(s) "," " sv string[errors]]];
```

code/processes/segmentedtickerplant.q
```diff
@@ -83,5 +83,38 @@ init:{
...
+     // check for valid tables
+     keydict: key(flip .stpps.stripeconfig[`segid]);
+     stripedtables: .stpps.t inter keydict;
+     wrongtables:(keydict except `subscriptiondefault) except stripedtables;
+     {if[0<count x;.lg.o[`jsonchecks;m:"Table ",string[x]," is not recognised"]]}each wrongtables;
```
jonathonmcmurray, line 103: again, don't think iteration is required here

```suggestion
     if[0<count wrongtables;.lg.o[`jsonchecks;"Table(s) ",(" " sv string[wrongtables])," not recognised"]];
```

code/processes/segmentedtickerplant.q
```diff
@@ -83,5 +83,38 @@ init:{
...
+     //Enable datastriping if all checks pass
+     $[min (0=count errors; 0=count wrongtables);
+      (.lg.o[`jsonchecks;"config checks complete and datastriping is on"];.ds.datastripe:1b);
+      .lg.o[`jsonchecks;"Datastriping is not enabled due to failed config checks"]]
```
jonathonmcmurray, line 107: I think this might be clearer if rearranged a little

```suggestion
     .ds.datastripe:min (0=count errors; 0=count wrongtables);
     $[.ds.datastripe;
      .lg.o[`jsonchecks;"config checks complete and datastriping is on"];
      .lg.o[`jsonchecks;"Datastriping is not enabled due to failed config checks"]
     ];
```

code/processes/segmentedtickerplant.q
```diff
@@ -83,5 +83,38 @@ init:{
...
+     };
+
+// Initialise config check
+configcheck:{
+     .lg.o[`configcheck;"initiate config check"];
+     scpath:first .proc.getconfigfile[string .ds.stripeconfig];
+     // Check striping.json file exists then runs configload
+     $[()~key hsym scpath; .lg.o[`configcheck;"Datastriping is not enabled as the following file can not be found: ",string .ds.stripeconfig]; configload[scpath]]
```
jonathonmcmurray, line 115: break this over multiple lines for clarity

```suggestion
     $[()~key hsym scpath;
       .lg.o[`configcheck;"Datastriping is not enabled as the following file can not be found: ",string .ds.stripeconfig];
       configload[scpath]
     ];
```

code/processes/segmentedtickerplant.q
```diff
@@ -83,5 +83,38 @@ init:{
...
+     };
+
+// Checks if each subscriptiondefault is set for each segment and errors if not defined
+jsonchecks:{[.stpps.stripeconfig]
+     // check defaults are ignore or all
+     defaults:{first (flip .stpps.stripeconfig[x])[`subscriptiondefault]}each key .stpps.stripeconfig;
+     errors:1+ where {[x] not ("ignore"~x) or ("all"~x)}each defaults;
+     {if[0<count x;.lg.o[`jsonchecks;m:"subscriptiondefault not defined as \"ignore\" or \"all\" for segment ",string[x]," "]]}each errors;
+     // check for valid tables
+     keydict: key(flip .stpps.stripeconfig[`segid]);
```
jonathonmcmurray, line 100: `keydict` isn't a very descriptive varibale name - maybe something like `configtables` (I'm assuming from the next few lines that this variable contains the tables that are listed in the config file?)

code/processes/segmentedtickerplant.q
SeanW17: I believe the error is logged in an earlier log message. This log message is just to log that datastriping will not be on as a consequence so the user knows normal TorQ is running instead of TorQv5 functionality. (I believe Jonny said the requirements were to have normal TorQ runing if any errors occur)

jonathonmcmurray: I think there's probably no harm including it here as well?

code/processes/segmentedtickerplant.q
jonathonmcmurray: OK fair enough - as long as everything else works OK even if this fails to load.

Might be worth doing the json load first, then returning early from the `configload` function if it fails rather than continuing & running the `jsonchecks` function which is surely going to fail as well if the json load failed?

code/processes/segmentedtickerplant.q
SeanW17: I will just put in one log message instead.

code/processes/segmentedtickerplant.q
SeanW17: This is the structure of the .stpps.stripeconfig
stripeconfigstructure
This line will do the same thing as original lie 96:
value first each .stpps.stripeconfig[;`subscriptiondefault]

code/processes/segmentedtickerplant.q
SeanW17: I believe in testing if the json file failed to load then the output message was logged and no other checks were done. Could be mistaken though.

jonathonmcmurray: as it's written here, I'm pretty sure `jsonchecks` will get called regardless of whether the json successfully loads

ThomasGillespie (approved):

jonathonmcmurray (approved):

dmitchard (approved):

PR #486, opened 2022-07-15
Changing casting of segid 

mccusker901:
Changing casting of segid from long to symbol, allowing for better range of naming constructs


jonathonmcmurray (commented):
minor formatting/comment stuff but looks fine

SeanW17 (changes requested):
I believe there is still whitespace on line 5 that needs removed.

jonathonmcmurray (approved):

PR #489, opened 2022-07-20
Access tables: Added `time field to tickerplant End-of-Period payload; access tables should now populate stptime field at EOP.

Lenzaum:
- `.stplg.checkends[]` updated to submit `.stplg.endofperioddata[]` to `.stplg.stpeoperiod[]`.
- `.wdb.datastripeendofperiod[]` updated to use offset UTC time `.z.p+.eodtime.dailyadj` when populating `stptime` field in `.wdb.access`.


code/wdb/datastripe.q
```diff
@@ -42,12 +40,15 @@ initdatastripe:{
...
+    .wdb.tablekeycols:.ds.loadtablekeycols[];
     t:tables[`.] except .wdb.ignorelist;
+
+    // create or load the access table
     .wdb.access: @[get;(` sv(.ds.td;.proc.procname;`access));([] table:t ; start:0Np ; end:0Np ; stptime:0Np ; keycol:`sym^.wdb.tablekeycols[t])];
     modaccess[.wdb.access];
     (` sv(.ds.td;.proc.procname;`access)) set .wdb.access;
-    .wdb.access:{[x] last .wdb.access where .wdb.access[`table]=x} each (key .wdb.tablekeycols);
+    .wdb.access:{[x] last .wdb.access where .wdb.access[`table]=x} each t;
     .wdb.access:`table xkey .wdb.access;
```
jonathonmcmurray, line 52: I might be missing something but I _think_ this is the same:

```suggestion
    .wdb.access:select by table from .wdb.access where table in t;
```

code/wdb/datastripe.q
```diff
@@ -57,25 +58,30 @@ if[.ds.datastripe;.proc.addinitlist[(`initdatastripe;`)]];
...
+    directory:` sv .Q.par[dir;partitionint;tablename],`;
+    .lg.o[`save;"Saving ",string[s]," data from ",string[tablename]," table to partition ",string[partitionint],". Table contains ",string[count enumdata]," rows."];
+    .lg.o[`save;"Saving data down to ",string[directory]];
+
+    /- upsert select data matched on partition to specific directory
+    .[upsert;
+        (directory;enumdata);
+        {[e] .lg.e[`upserttopartition;"Failed to save table to disk : ",e];'e}
+    ];
+    };
```
jonathonmcmurray, line 84: is there any actual functional change in this section or is it just layout changes? layout changes should typically be kept out of functional change PRs (can make a separate PR for layout changes if needed) as it makes it difficult to review what has actually changed (if anything) in this section.

code/wdb/datastripe.q
```diff
@@ -22,14 +22,12 @@ modaccess:{[accesstab]};
     // on first save down we need to replace the null valued start time in the access table
     // using the first value in the saved data
     starttimes:.ds.getstarttime each t;
-    .wdb.access:update start:starttimes^start, end:?[(nextp>starttimes)&(starttimes<>0Np);nextp;0Np], stptime:data[][`time] from .wdb.access;
+    .wdb.access:update start:starttimes^start, end:?[(nextp>starttimes)&(starttimes<>0Np);nextp;0Np], stptime:data[][`p] from .wdb.access;
```
jonathonmcmurray, line 25: so was referencing `time` before a bug? I can't see from context or PR description etc. why this is changed

code/wdb/datastripe.q
```diff
@@ -22,14 +22,12 @@ modaccess:{[accesstab]};
...
     modaccess[.wdb.access];
 
     // call the savedown function
     .ds.savealltablesoverperiod[.ds.td;nextp;lasttime];
-
-    tabs:.ds.deletetablebefore'[t;`time;lasttime];
-    .lg.o[`reload;"Kept ",string[.ds.periodstokeep]," period",$[.ds.periodstokeep>1;"s";""]," of data from : ",", " sv string[tabs]];
+    .lg.o[`reload;"Kept ",string[.ds.periodstokeep]," period",$[.ds.periodstokeep>1;"s";""]," of data from : ",", " sv string[t]];
```
jonathonmcmurray, line 30: why do we not need to use `.ds.deletetablebefore` anymore?

code/wdb/datastripe.q
```diff
@@ -42,12 +40,15 @@ initdatastripe:{
     // update endofperiod function
     endofperiod::.wdb.datastripeendofperiod;
     
-    .wdb.tablekeycols:.ds.loadtablekeycols[];  // replace with dictionaries from json file
+    // load in variables
+    .wdb.tablekeycols:.ds.loadtablekeycols[];
     t:tables[`.] except .wdb.ignorelist;
+
+    // create or load the access table
```
jonathonmcmurray, line 47:
```suggestion
    // load the access table, fall back to generating table if load fails
```
think something like this would be more descriptive of what's happening here

code/wdb/datastripe.q
```diff
@@ -42,12 +40,15 @@ initdatastripe:{
     // update endofperiod function
     endofperiod::.wdb.datastripeendofperiod;
     
-    .wdb.tablekeycols:.ds.loadtablekeycols[];  // replace with dictionaries from json file
+    // load in variables
```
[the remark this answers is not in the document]
Lenzaum: Hm, I don't remember making this change...given that the original comment provides a better description, I've no objections to reverting this.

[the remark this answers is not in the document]
jonathonmcmurray: tbh I don't think the original comment is much better - still too vague to be useful IMO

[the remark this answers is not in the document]
drgdavies: > Hm, I don't remember making this change...given that the original comment provides a better description, I've no objections to reverting this.

If you put up a PR please ensure you have reviewed it also to ensure it only contains the changes intended.

code/wdb/datastripe.q
Lenzaum: That does appear more concise & coherent - I can commit this change, if no-one else has any objections?

code/wdb/datastripe.q
Lenzaum: Just layout changes here - format was incongruent with rest of script, and difficult to read. I can revert this change & make a separate PR for it later, as you've suggested?

code/wdb/datastripe.q
Lenzaum: Updating the `checkends[]` function in `segmentedtickerplant/stplog.q` resolves the bug with referencing `time`; the change here to reference `p` is to populate `stptime` with the offset UTC time `.z.p+.eodtime.dailyadj`, as opposed to the local time `.z.P`.

jonathonmcmurray: OK - so why are we making that change to use offset UTC time instead? And if someone looks at this piece of code in 6 months, how are they going to know what `p` means here? Is that documented somewhere?

code/wdb/datastripe.q
Lenzaum: Hm, I don't recall making this change either - I feel it'd be safer to revert this as well.

code/wdb/datastripe.q
jonathonmcmurray: Understood - but what is the reason for that suggestion? (my point here is basically that you should generally have at least some understanding of the reason/justification for the change you are making)

Regarding `p` - the fact that it's used elsewhere may not be that helpful to someone looking at this line of code. Ideally I shouldn't have to go digging through a bunch of other code to understand what a variable/name here means. It's not apparent looking at this piece of code where the `data` object might be coming from, or what it contains etc. - even adding a comment here to explain the structure of `data` & what `p` means would be very helpful IMO

jonathonmcmurray (approved):

code/wdb/datastripe.q
Lenzaum: Had a chat with Callum Reid, and updated to use the `time` field of `.stplg.endofperioddata[]`, i.e. the local time `.z.P`.

dmitchard (approved):

PR #490, opened 2022-07-20
Tcs 118 extend rdb time window

esweeney97:
Increase the number of periods the rdb/datastore keeps in memory if the end of day tail sort process is still in progress when end of period occurs (tail sort process still to be created)

- [ ] update to work with tail sort process when created
- [ ] unit tests

### DIFF
--- a/code/processes/wdb.q
+++ b/code/processes/wdb.q
@@ -194,6 +194,7 @@ endofday:{[pt;processdata]
 		$[sortenabled;endofdaysort;informsortandreload] . (savedir;pt;tablist;writedownmode;mergelimits;hdbsettings)];
 	.lg.o[`eod;"deleting data from tabsizes"];
 	@[`.wdb;`tabsizes;0#];
+    // if[.ds.datastripe;.ds.resetrdbwindow[]];
     .lg.o[`eod;"end of day is now complete"];
     .wdb.currentpartition:pt+1;
 	};
@@ -284,6 +285,12 @@ reloadsymfile:{[symfilepath]
   .lg.o[`sort; "reloading the sym file from: ",string symfilepath];
   @[load; symfilepath; {.lg.e[`sort;"failed to reload sym file: ",x]}]
  }
+
+resetrdbwindow:{
+    .lg.o[`rdbwindow;"resetting rdb moving time window"];
+    rdbprocs:.servers.getservers[`proctype;.wdb.rdbtypes;()!();1b;0b];
+    {neg[x]".rdb.tailsortcomplete:1b";neg[x][]}each exec w from rdbprocs;
+    };
 
 endofdaysortdate:{[dir;pt;tablist;hdbsettings]
   /-sort permitted tables in database
@@ -376,6 +383,7 @@ endofdaymerge:{[dir;pt;tablist;mergelimits;hdbsettings]
   if[permitreload; 
     doreload[pt];
     ];
+  if[.ds.datastripe;resetrdbwindow[]];
   };
 	
 /- end of day sort [depends on writedown mode]
@@ -424,7 +432,7 @@ informgateway:{[message]
 /- function to call that will cause sort & reload process to sort data and reload rdb and hdbs
 informsortandreload:{[dir;pt;tablist;writedownmode;mergelimits;hdbsettings]
 	.lg.o[`informsortandreload;"attempting to contact sort process to initiate data sort"];
-	$[count sortprocs:.mformsoreservers.getservers[`proctype;sorttypes;()!();1b;0b];
+	$[count sortprocs:.servers.getservers[`proctype;sorttypes;()!();1b;0b];
 		{.[{neg[y]@x;neg[y][]};(x;y);{.lg.e[`informsortandreload;"unable to run command on sort and reload process"];'x}]}[(`.wdb.endofdaysort;dir;pt;tablist;writedownmode;mergelimits;hdbsettings);] each exec w from sortprocs;
 		[.lg.e[`informsortandreload;"can't connect to the sortandreload - no sortandreload process detected"];
 		 // try to run the sort locally
--- a/code/processes/wdb.q
+++ b/code/processes/wdb.q
@@ -194,7 +194,6 @@ endofday:{[pt;processdata]
 		$[sortenabled;endofdaysort;informsortandreload] . (savedir;pt;tablist;writedownmode;mergelimits;hdbsettings)];
 	.lg.o[`eod;"deleting data from tabsizes"];
 	@[`.wdb;`tabsizes;0#];
-    // if[.ds.datastripe;.ds.resetrdbwindow[]];
     .lg.o[`eod;"end of day is now complete"];
     .wdb.currentpartition:pt+1;
 	};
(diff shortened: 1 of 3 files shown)


code/processes/wdb.q
```diff
@@ -4,7 +4,7 @@
 /-the row check is set on a timer - the interval may be specified by the user
 /-at eod the on-disk data may be sorted and attributes applied as specified in the sort.csv file
 
-/- load parameters & functions from common script.
+/- load parameters & functions from common script.
```
SeanW17, line 7: Can this be removed from the commit?

code/rdb/datastripe.q
```diff
@@ -19,12 +24,26 @@
...
     };
 
+// end of day function - no savedown functionality, just wipes tables and updates date partition
+.rdb.datastripeendofday:{[date;processdata]
+    .rdb.tailsortcomplete:0b;
+    // add date+1 to rdbpartition global
+    .rdb.rdbpartition,:: date+1;
+    .lg.o[`rdbpartition;"rdbpartition contains - ","," sv string .rdb.rdbpartition];
+    t:tables[`.] except .rdb.ignorelist;
+    .lg.o[`clear;"Wiping following tables - ","," sv string t];
```
SeanW17, line 34: Why are the commas needed in these log messages?

code/rdb/datastripe.q
```diff
@@ -1,13 +1,18 @@
+.rdb.extendperiods:.ds.periodstokeep
 .rdb.datastripeendofperiod:{[currp;nextp;data]
 
     .lg.o[`reload;"reload command has been called remotely"];
 
     t:tables[`.] except .rdb.ignorelist;
 
+    // check if tail sort process is complete
+    if[.rdb.tailsortcomplete;.rdb.extendperiods:.ds.periodstokeep];
```
SeanW17, line 9: Is this not reassigned on line 1 regardless of tailsortcomplete?

code/processes/wdb.q
esweeney97: yeah, will remove this

code/rdb/datastripe.q
esweeney97: The first comma in the log messages is performing a join operation, the second is being used to create a comma-separated list using the `sv` keyword

SeanW17: Would this look more clear?
```suggestion
    .lg.o[`rdbpartition;"rdbpartition contains - ", "," sv string .rdb.rdbpartition];
    t:tables[`.] except .rdb.ignorelist;
    .lg.o[`clear;"Wiping following tables - ", "," sv string t];
```

code/rdb/datastripe.q
esweeney97: Line 1 is just initialising `.rdb.extendperiods` to whatever value `.ds.periodstokeep` is set to, it only runs when the code is first loaded

code/processes/tailsort.q
```diff
@@ -75,3 +77,12 @@ endofday:{[pt;procname]
   .lg.o[`endofday;"end of day message received from ",string[procname]," - ",string[pt]];
   loadandsave[pt;procname];
   };
+
+/- notify rdb when tail sort process complete
+resetrdbwindow:{
+  .lg.o[`rdbwindow;"resetting rdb moving time window"];
+  rdbprocs:.servers.getservers[`proctype;.ts.rdbtypes;()!();1b;0b];
+  {neg[x]".rdb.tailsortcomplete:1b";neg[x][]}each exec w from rdbprocs;
+  };
```
SeanW17, line 86: I believe this may need to be defined before it's called.

code/processes/tailsort.q
```diff
@@ -42,6 +43,13 @@ addpattr:{[hdbdir;pt;tabname]
   ];
   };
 
+/- notify rdb when tail sort process complete
+resetrdbwindow:{
+  .lg.o[`rdbwindow;"resetting rdb moving time window"];
+  rdbprocs:.servers.getservers[`proctype;.ts.rdbtypes;()!();1b;0b];
+  {neg[x]".rdb.tailsortcomplete:1b";neg[x][]}each exec w from rdbprocs;
```
jonathonmcmurray, line 50: why are we blocking on the handles? we don't require a response, do we have to wait for the first rdb before we send to second etc.?

tests/datastore/tests.csv
```diff
@@ -0,0 +1,15 @@
+action,ms,bytes,lang,code,repeat,minver,comment
+before,0,0,q,.servers.startup[],1,,"Start TorQ connection management"
+before,0,0,q,.proc.sys "sleep 2",1,,"Wait for proc to start"
```
jonathonmcmurray, line 3:
```suggestion
before,0,0,q,.os.sleep 2,1,,"Wait for proc to start"
```

tests/datastore/settings.q
```diff
@@ -0,0 +1,8 @@
+// IPC connection parameters
+.servers.CONNECTIONS:`segmentedtickerplant`rdb;
+
+// rmdtfromgetpar function in processes/rdb.q
+.rdb.rmdtfromgetpar:{[date] 
+    .rdb.rdbpartition:: .rdb.rdbpartition except date;
+    .lg.o[`rdbpartition;"rdbpartition contains - ","," sv string .rdb.rdbpartition];
+    }
```
jonathonmcmurray, line 8: we shouldn't replicate code for the tests - we should be testing the _actual_ code that gets used. Reason being, if the code in `processes/rdb.q` gets changed down the line, there is very little chance anyone is going to remember that this test also needs to be updated, it will still be using the old version and could be "passing" even though the change broke something.

that may mean refactoring a little to move the function from `code/processes/rdb.q` to a file in `code/rdb/` so that it gets loaded in the test process.

tests/datastore/run.sh
```diff
@@ -0,0 +1,28 @@
...
+  -procfile ${testpath}/process.csv \
+  -segid 1 \
+  $debug $stop $write $quiet
+
+#$QCMD ${TORQHOME}/torq.q \
+#      -proctype wdb \
+#      -procname wdb1 -test ${KDBTESTS}/tailer/savedown/ \
+#      -procfile ${KDBTESTS}/tailer/savedown/process.csv \
+#      -load ${KDBCODE}/processes/wdb.q -debug \
+
```
jonathonmcmurray, line 27:
```suggestion
```

remove commented lines

tests/datastore/tests.csv
```diff
@@ -0,0 +1,15 @@
...
+before,0,0,q,".rdb.ignorelist,:`KUT`KUTR",1,,"Add test result tables to ignore list"
+before,0,0,q,stph:gethandle[`stp1],1,,"Open STP handle"
```
jonathonmcmurray, line 5: I'm not sure this line is required - `.servers.startup[]` should have opened the handle and I don't think you use `stph` later in the tests

```suggestion
```

code/processes/tailsort.q
esweeney97: you're right, we don't need to wait for the first rdb, will change this

jonathonmcmurray (approved):

PR #493, opened 2022-07-25
Tcs 135 update access table in rdb to use ignorelist

dmitchard:
-RDB access now uses the ignore list in the same way the tailer does

### DIFF
--- a/code/rdb/datastripe.q
+++ b/code/rdb/datastripe.q
@@ -44,3 +44,4 @@ updategw:{[h]
     neg[h](`.ds.updateaccess;newtab);
 
     };
+    
--- a/code/rdb/datastripe.q
+++ b/code/rdb/datastripe.q
@@ -10,7 +10,10 @@
     .lg.o[`reload;"Kept ",string[.ds.periodstokeep]," period",$[.ds.periodstokeep>1;"s";""]," of data from : ",", " sv string[tabs]];
 
     // update the access table in the rdb
-    .rdb.access:update start:lasttime^(.ds.getstarttime each key .rdb.tablekeycols), stptime:data[][`time] from .rdb.access;
+    t:tables[`.] except .rdb.ignorelist;
+    starttimes:.ds.getstarttime each t;
+    .rdb.access:update start:starttimes, stptime:data[][`time] from .rdb.access;
+    /.rdb.access:update start:lasttime^(.ds.getstarttime each key .rdb.tablekeycols), stptime:data[][`time] from .rdb.access;
     modaccess[.rdb.access];
 
     };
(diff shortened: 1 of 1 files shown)


code/rdb/datastripe.q
```diff
@@ -10,7 +10,9 @@
     .lg.o[`reload;"Kept ",string[.ds.periodstokeep]," period",$[.ds.periodstokeep>1;"s";""]," of data from : ",", " sv string[tabs]];
 
     // update the access table in the rdb
-    .rdb.access:update start:lasttime^(.ds.getstarttime each key .rdb.tablekeycols), stptime:data[][`time] from .rdb.access;
+    t:tables[`.] except .rdb.ignorelist;
+    starttimes:.ds.getstarttime each t;
+    .rdb.access:update start:starttimes^start, stptime:data[][`time] from .rdb.access;
```
CReid98, line 15: `.rdb.access:update start:starttimes, stptime:data[][`time] from .rdb.access;`

otherwise this will not overwrite the old start column

dmitchard: so does the not need to happen in the wdb? in the same place

CReid98: It gets overwritten in the rdb because data is cleared from memory each period. What is on disk isn't getting cleared so the start time stays the same.

CReid98 (approved):

code/rdb/datastripe.q
```diff
@@ -10,7 +10,9 @@
     .lg.o[`reload;"Kept ",string[.ds.periodstokeep]," period",$[.ds.periodstokeep>1;"s";""]," of data from : ",", " sv string[tabs]];
 
     // update the access table in the rdb
-    .rdb.access:update start:lasttime^(.ds.getstarttime each key .rdb.tablekeycols), stptime:data[][`time] from .rdb.access;
+    t:tables[`.] except .rdb.ignorelist;
+    starttimes:.ds.getstarttime each t;
+    .rdb.access:update start:starttimes, stptime:data[][`time] from .rdb.access;
```
jonathonmcmurray, line 15: is it guaranteed that `t` will be in the right order? I assume `.rdb.access` includes the table name, can this not just be something like

```suggestion
    .rdb.access:update start:.ds.getstarttime each table, stptime:data[][`time] from .rdb.access;
```

(assuming table name column in `.rdb.access` is `table`)

code/rdb/datastripe.q
dmitchard: but this line in the wdb relates to the access table in memory which uses the in memory tables to calculate its times in the same way as in the rdb

code/rdb/datastripe.q
dmitchard: Yes, you are right!

jonathonmcmurray (approved):
Change looks fine - please make sure that PRs have descriptive titles & descriptions though; the title is quite vague here & I think the description is outdated now as well

PR #495, opened 2022-07-26
Tcs 141 fix initiate data stripe not running

dmitchard:
- TCS-116 Changed .ds.datastripe default value to 0b for both wdb and rdb as datastriping should only be turned on if tickerplant is configured for datastriping
- This introduced an issue that meant the initdatastripe functions in the datastripe.q scripts of the rdb/wdb were never be run as .ds.datastripe was set to 1b when subscribing which was after the check in each of these scripts.
- rdb/wdb now call initdatastripe after subscribing if .ds.datastripe is on

### DIFF
--- a/code/processes/rdb.q
+++ b/code/processes/rdb.q
@@ -236,5 +236,5 @@ $[.rdb.connectonstart;
 //Updating .rdb.subtables to only include tables that actually subscribed
 .rdb.subtables:$[.ds.datastripe;tables[] except `heartbeat;.rdb.subtables];
 
-//Initialise datastripe
+/- initialise datastripe
 if[.ds.datastripe;initdatastripe[];.lg.o[`dsinit;"datastripe on: initialising datastripe"]];
--- a/code/processes/rdb.q
+++ b/code/processes/rdb.q
@@ -235,3 +235,6 @@ $[.rdb.connectonstart;
 
 //Updating .rdb.subtables to only include tables that actually subscribed
 .rdb.subtables:$[.ds.datastripe;tables[] except `heartbeat;.rdb.subtables];
+
+//Initialise datastripe
+if[.ds.datastripe;initdatastripe[];.lg.o[`dsinit;"datastripe on: initialising datastripe"]];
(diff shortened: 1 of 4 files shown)


SeanW17 (approved):

code/processes/rdb.q
```diff
@@ -235,3 +235,6 @@ $[.rdb.connectonstart;
 
 //Updating .rdb.subtables to only include tables that actually subscribed
 .rdb.subtables:$[.ds.datastripe;tables[] except `heartbeat;.rdb.subtables];
+
+/- initialise datastripe
+if[.ds.datastripe;initdatastripe[];.lg.o[`dsinit;"datastripe on: initialising datastripe"]];
```
jonathonmcmurray, line 240: minor nitpick - you're logged _after_ calling initdatastripe, so should the log be `initialised datastripe` instead? or come before the `initdatastripe[]` call?

code/rdb/datastripe.q
```diff
@@ -28,5 +28,4 @@ initdatastripe:{
     .ds.checksegid[];    
     };
 
-if[.ds.datastripe;.proc.addinitlist[(`initdatastripe;`)]];
```
jonathonmcmurray, line 31: I'm a little confused why this doesn't work but the other way does? my understanding of the initlist in TorQ is that the functions in it will be run after all the scripts are loaded etc., so it should be running after the subscription anyway?

jonathonmcmurray (commented):
as per comment, I'm confused about the reason the old way doesn't work. change looks OK, but I can't see how it changes the overall behaviour

code/processes/rdb.q
dmitchard: Yes sorry, log should be before

code/rdb/datastripe.q
dmitchard: So .ds.datastripe used to be set as default for rdb/wdb as on so it would then add it to the list, now it is default to 0 and then is set to on inside the subscription function. This means that the initdatastripe function was not being added to the init list

jonathonmcmurray: Ahhh, of course, that makes sense. Cool, thanks for explaining :+1:

jonathonmcmurray (approved):

PR #497, opened 2022-07-28
stplog.q: Minor changes to currperiod/nextperiod updating by end-of-period functions.

Lenzaum:
These are primarily aesthetic changes for consistency, but may also future-proof against unforeseen issues.
- `endofperiod[]` now updates value of global variable `nextperiod` as `multilogperiod+currperiod`, to mirror `stpeoperiod[]`.
- `stpeoperiod[]` now updates value of global variable `currperiod` using the input `nextpd`, as opposed to the variable `nextperiod`, to mirror `endofperiod[]`.


CReid98 (approved):

code/segmentedtickerplant/stplog.q
```diff
@@ -209,7 +210,7 @@ endofperiod:{[currentpd;nextpd;data]
 stpeoperiod:{[currentpd;nextpd;data;rolllogs]
   .lg.o[`stpeoperiod;"passing on endofperiod message to subscribers"];
   .stpps.endp[currentpd;nextpd;data];                      // sends endofperiod message to subscribers
-  currperiod::nextperiod;                                  // increments current period
+  currperiod::nextpd;                                      // increments current period
```
jonathonmcmurray, line 213: is there a reason for this to be passed as an argument rather than use the global? I see on the next line that the global is updated, is it valid for anything _other_ than the global to be passed as `nextpd`?

Suppose what I'm really asking is, should the change instead be to remove the `nextpd` argument from this function and update every call to this function?

code/segmentedtickerplant/stplog.q
```diff
@@ -201,6 +201,7 @@ endofperiod:{[currentpd;nextpd;data]
   .lg.o[`endofperiod;"executing end of period for ",.Q.s1 `currentperiod`nextperiod!(currentpd;nextpd)];
   .stpps.endp[currentpd;nextpd;data];                   // sends endofperiod message to subscribers
   currperiod::nextpd;                                   // increments current period
+  nextperiod::multilogperiod+currperiod;                // increments next period
```
jonathonmcmurray, line 204: `nextperiod` is already updated inside `stpeoperiod` (line 214) - are we sure this won't cause a double increment?

Lenzaum: As I understand, `endofperiod[]` is intended to be called on chained tickerplants, immediately downstream from `stpeoperiod[]` on the "core" TP.
Consequently, `stpeoperiod[]` and `endofperiod[]` update the `nextperiod` values on their _specific_ processes; alternatively, `nextperiod` on a given process is only updated by one of `stpeoperiod[]` and `endofperiod[]`.

code/segmentedtickerplant/stplog.q
Lenzaum: The log message at the end of this function (line 218) specifically uses the `currentpd` & `nextpd` arguments, which at this stage would have different values to `currperiod` & `nextperiod`.
That said, it occurs to me that a log message for "new values for current and next period" ought to use the freshly updated `currperiod` & `nextperiod` instead - in which case, the `currentpd` & `nextpd` arguments could be dropped from `stpeoperiod[]` without consequence.

jonathonmcmurray: yep, I agree, that log message is currently wrong

code/segmentedtickerplant/stplog.q
Lenzaum: From what I can determine, `nextperiod` is not utilised by the chained TP. However, `currperiod` is, as a consequence of the call to `periodrollover[]` on line 205; I feel that keeping `nextperiod` updated to a "sensible" counterpart value is safer & more future-proof than leaving it unchanged from its initial definition as part of running `init[]`.

The alternative is adding logic to `init[]` to only define `nextperiod` if its called on a chained TP process, but that may be another ticket/PR entirely.

jonathonmcmurray: If `nextperiod` isn't used by chainedtp (which IMO, it shouldn't be - CTP does not control it's own rollover, so it never needs to know when the next period starts), I don't think we should be adding code to update it for no reason.

Lenzaum: Okay, I can revert this addition then - I feel a further update to not have the chainedtp define nextperiod at all may help conserve a sliver of memory, but that's for another time.

jonathonmcmurray (approved):
LGTM

PR #498, opened 2022-07-29
Tcs 147 start time of access on disk not updating

dmitchard:
- Start time of access table never updating as a ^ join was being used which would only update if it was null. This meant that the very first entry into the start time would be repeated forever
-  Once this was fixed, I realised that the timing for the start time did not quite match what was in memory in the tailer
- this was because the savedown of tables was being called after the access was updated, so the access was displaying times for data that no longer existed in memory, this has been amended

### DIFF
--- a/code/wdb/datastripe.q
+++ b/code/wdb/datastripe.q
@@ -18,7 +18,11 @@ modaccess:{[accesstab]};
     // remove periods of data from tables
     t:tables[`.] except .wdb.ignorelist;
     lasttime:nextp-.ds.periodstokeep*(nextp-currp);
-
+    
+    // call the savedown function
+    .ds.savealltablesoverperiod[.ds.td;nextp;lasttime];
+    .lg.o[`reload;"Kept ",string[.ds.periodstokeep]," period",$[.ds.periodstokeep>1;"s";""]," of data from : ",", " sv string[t]];
+    
     // update the access table in the wdb
     // on first save down we need to replace the null valued start time in the access table
     // using the first value in the saved data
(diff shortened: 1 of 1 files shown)


CReid98 (commented):
The originally intended functionality was that the start time was only meant to update once at the start of day and contained the time of the first value saved to disk. If this is changing I think you need to write a bit more to explain why.

code/wdb/datastripe.q
```diff
@@ -18,18 +18,18 @@ modaccess:{[accesstab]};
...
-
+    
+    // call the savedown function
+    .ds.savealltablesoverperiod[.ds.td;nextp;lasttime];
+    .lg.o[`reload;"Kept ",string[.ds.periodstokeep]," period",$[.ds.periodstokeep>1;"s";""]," of data from : ",", " sv string[t]];
+    
     // update the access table in the wdb
     // on first save down we need to replace the null valued start time in the access table
     // using the first value in the saved data
     starttimes:.ds.getstarttime each t;
```
CReid98, line 29: is this variable still being used? looks like it can be removed

dmitchard: > The originally intended functionality was that the start time was only meant to update once at the start of day and contained the time of the first value saved to disk. If this is changing I think you need to write a bit more to explain why.

So with the way this is joined and saved to disk, you have that same start value repeated forever, it then in memory does not relate to what data is currently in memory for the wdb

CReid98: It's not recording what is in memory in the wdb, it's just recording what is saved to disk. It should be cleared as part of the endofday procedure (or moved to hdb) but it's expected that the start value will be repeated during the day.

dmitchard: if the access on disk is meant to be a record of all the periods, then you need the start time of those periods to update otherwise you will have a million

CReid98: In that case you would need to update the start time for the access table on disk, but the start time for the access table in memory should always be the first value saved that day.

PR #502, opened 2022-08-03
TCS-144 - Change enumeration in the Tailer and create symlink to HDB sym file

eimear-dalzell:
Change to enumeration in tailer to enumerate data against HDB symfile, rather than creating a sym file specific to each tailer.
Checks tailer directories for symlink to HDB sym file and creates this symlink if necessary.

### DIFF
--- a/code/wdb/datastripe.q
+++ b/code/wdb/datastripe.q
@@ -59,6 +59,19 @@ initdatastripe:{
 
 \d .ds
 
+symlink:{
+    /- function to create HDB sym file and symlink to this sym file at start up
+    /- create HDB sym file
+    symdir:`$raze string .wdb.hdbdir,"/sym";
+    if[not `sym in key .wdb.hdbdir;symdir set `symbol$()];
+    .lg.o[`hdbsym;"creating HDB symfile"];
+
+    /- create symlink
+    basedir:` sv .ds.td,.proc.procname;
+    createsymlink[basedir;.wdb.hdbdir;`sym];
+    .lg.o[`symlink;"creating symlink"];
+    };
+
 createsymlink:{[tdpath;hdbpath;symfile]
     /- function to create symlink to HDB sym file in taildir
     tdsympath:1_raze string tdpath,"/",symfile;
--- a/code/wdb/datastripe.q
+++ b/code/wdb/datastripe.q
@@ -61,7 +61,7 @@ initdatastripe:{
 
 createsymlink:{[tdpath;hdbpath;symfile]
     /- function to create symlink to HDB sym file in taildir
-    tdsympath:1_raze string tdpath,"/",.proc.procname,"/",symfile;
+    tdsympath:1_raze string tdpath,"/",symfile;
     hdbsympath:1_raze string hdbpath,"/",symfile;
 
     /- linux command to create symlink in specified dirs
@@ -95,7 +95,7 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
     ];
 
     /- check if sym file exists in taildir, create sym link to HDB sym file if not
-    if[not `sym in key hsym basedir;createsymlink[.ds.td;.wdb.hdbdir;`sym]];
+    if[not `sym in key hsym basedir;createsymlink[basedir;.wdb.hdbdir;`sym]];
     };
 
 savetablesoverperiod:{[dir;tablename;nextp;lasttime]
--- a/code/wdb/datastripe.q
+++ b/code/wdb/datastripe.q
@@ -62,14 +62,14 @@ initdatastripe:{
 symlink:{
     /- function to create HDB sym file and symlink to this sym file at start up
     /- create HDB sym file
-    symdir:`$raze string .wdb.hdbdir,"/sym";
-    if[not `sym in key .wdb.hdbdir;symdir set `symbol$()];
-    .lg.o[`hdbsym;"creating HDB symfile"];
+    sympath:` sv (.wdb.hdbdir;`sym);
+    .lg.o[`hdbsym;"creating HDB sym file"];
+    if[not `sym in key .wdb.hdbdir;sympath set `symbol$()];
 
     /- create symlink
     basedir:` sv .ds.td,.proc.procname;
-    if[not `sym in key basedir;createsymlink[basedir;.wdb.hdbdir;`sym]];
     .lg.o[`symlink;"creating symlink"];
+    if[not `sym in key basedir;createsymlink[basedir;.wdb.hdbdir;`sym]];
     };
 
 createsymlink:{[tdpath;hdbpath;symfile]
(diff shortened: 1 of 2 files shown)


SeanW17 (approved):

code/wdb/datastripe.q
```diff
@@ -59,6 +59,15 @@ initdatastripe:{
 
 \d .ds
 
+createsymlink:{[tdpath;hdbpath;symfile]
+    /- function to create symlink to HDB sym file in taildir
+    tdsympath:1_raze string tdpath,"/",.proc.procname,"/",symfile;
+    hdbsympath:1_raze string hdbpath,"/",symfile;
+
+    /- linux command to create symlink in specified dirs
+    system"ln -s ",hdbsympath," ",tdsympath;
```
jonathonmcmurray, line 68: I think we should error trap this and think about what we do in case of failure - how critical is the symlink, can we log an error and continue or do we need to do something more?

code/wdb/datastripe.q
```diff
@@ -68,6 +77,7 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
     s:first raze value'[?[enumdata;();1b;enlist[keycol]!enlist keycol]];
 
     /- get process specific taildir location
+    basedir:` sv dir,.proc.procname;
     dir:` sv dir,.proc.procname,`$ string .wdb.currentpartition;
```
jonathonmcmurray, line 81:
```suggestion
    basedir:` sv dir,.proc.procname;
    dir:` sv basedir,`$ string .wdb.currentpartition;
```

code/wdb/datastripe.q
```diff
@@ -83,6 +93,9 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
         (directory;enumdata);
         {[e] .lg.e[`upserttopartition;"Failed to save table to disk : ",e];'e}
     ];
+
+    /- check if sym file exists in taildir, create sym link to HDB sym file if not
+    if[not `sym in key hsym basedir;createsymlink[.ds.td;.wdb.hdbdir;`sym]];
```
jonathonmcmurray, line 98: this is a little confusing/non-obvious to me - why are we checking for sym in `basedir` but then creating the symlink in `.ds.td`? surely we should be creating the symlink in the same location we checked for it, are `basedir` and `.ds.td` the same? I think we should use the same variable in both places in this line. If they are the same, do we need the local `basedir` variable if this is already stored globally as `.ds.td`?

code/wdb/datastripe.q
```diff
@@ -68,6 +77,7 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
     s:first raze value'[?[enumdata;();1b;enlist[keycol]!enlist keycol]];
 
     /- get process specific taildir location
+    basedir:` sv dir,.proc.procname;
```
jonathonmcmurray, line 80: is `dir` here the same in `savetablesoverperiod` below? because there it looks like a similar line was replaced by `.wdb.hdbdir`, would the same be valid here?

code/wdb/datastripe.q
```diff
@@ -93,7 +106,7 @@ savetablesoverperiod:{[dir;tablename;nextp;lasttime]
     partitionlist: ?[tablename;();();(distinct;keycol)];
 
     /- enumerate and then split by keycol
-    symdir:` sv dir,.proc.procname;
+    symdir:.wdb.hdbdir;
     enumkeycol: .Q.en[symdir;?[tablename;enlist (<;`time;nextp);0b;()]];
```
jonathonmcmurray, line 110:
```suggestion
    enumkeycol: .Q.en[.wdb.hdbdir;?[tablename;enlist (<;`time;nextp);0b;()]];
```

can possibly just use `.wdb.hdbdir` directly instead of creating a local copy of that variable?

code/wdb/datastripe.q
eimear-dalzell: Yep, I think logging an error and continuing would be fine - I'll add in some error trapping here 
I believe the symlink isn't critical to the functionality of the tailer - it's more to have a complete IDB/for the tail reader to have when it is loading data

jonathonmcmurray: if the tail reader won't be able to load the intraday data without the symlink, it sounds pretty critical IMO - might be worth having a discussion on what should happen in the scenario that symlink fails

code/wdb/datastripe.q
```diff
@@ -59,6 +59,32 @@ initdatastripe:{
 
 \d .ds
 
+symlink:{
+    /- function to create HDB sym file and symlink to this sym file at start up
+    /- create HDB sym file
+    symdir:`$raze string .wdb.hdbdir,"/sym";
+    if[not `sym in key .wdb.hdbdir;symdir set `symbol$()];
+    .lg.o[`hdbsym;"creating HDB symfile"];
```
jonathonmcmurray, line 67: I think the log should probably be before the action - at the moment by the time it logs, the file is already created

code/wdb/datastripe.q
```diff
@@ -59,6 +59,32 @@ initdatastripe:{
 
 \d .ds
 
+symlink:{
+    /- function to create HDB sym file and symlink to this sym file at start up
+    /- create HDB sym file
+    symdir:`$raze string .wdb.hdbdir,"/sym";
```
jonathonmcmurray, line 65: I think the variable name here is misleading, as we're not creating a directory

you also shouldn't need the raze

```suggestion
    symdir:`$string[.wdb.hdbdir],"/sym";
```

or possibly can just use `sv` https://code.kx.com/q/ref/sv/#filepath-components

```suggestion
    symdir:` sv (wdb.hdbdir;`sym);
```

code/wdb/datastripe.q
```diff
@@ -59,6 +59,32 @@ initdatastripe:{
...
+    if[not `sym in key .wdb.hdbdir;symdir set `symbol$()];
+    .lg.o[`hdbsym;"creating HDB symfile"];
+
+    /- create symlink
+    basedir:` sv .ds.td,.proc.procname;
+    if[not `sym in key basedir;createsymlink[basedir;.wdb.hdbdir;`sym]];
+    .lg.o[`symlink;"creating symlink"];
```
jonathonmcmurray, line 72: same as above with the logging

code/wdb/datastripe.q
```diff
@@ -59,6 +59,32 @@ initdatastripe:{
...
+    };
+
+createsymlink:{[tdpath;hdbpath;symfile]
+    /- function to create symlink to HDB sym file in taildir
+    tdsympath:1_raze string tdpath,"/",symfile;
+    hdbsympath:1_raze string hdbpath,"/",symfile;
```
jonathonmcmurray, line 78:
```suggestion
    tdsympath:1_string ` sv (tdpath;symfile);
    hdbsympath:1_string ` sv (hdbpath;symfile);
```

I think you can probably use `sv` here to build the paths

code/wdb/datastripe.q
eimear-dalzell: good point, I've renamed the variable to `sympath` and went for the `sv` method to keep it consistent with other path creations in this file

code/wdb/datastripe.q
eimear-dalzell: fixed the logging in both places now too

jonathonmcmurray (approved):

PR #505, opened 2022-08-04
Move access table to within date partition in taildir

dmitchard:
- Access table used to be in taildir/wdb
- Now I have moved it so that it is in tailer/wdb1/2022.08.04
- This means when the tailreader then loads the access table from the current partition, it will just get info on that partition rather than one access table with a start time that is the time of first launch

### DIFF
--- a/code/wdb/datastripe.q
+++ b/code/wdb/datastripe.q
@@ -31,9 +31,9 @@ modaccess:{[accesstab]};
     .lg.o[`reload;"Kept ",string[.ds.periodstokeep]," period",$[.ds.periodstokeep>1;"s";""]," of data from : ",", " sv string[t]];
     
     // update the access table on disk
-    atab:get ` sv(.ds.td;.proc.procname;`access);
+    atab:get ` sv(.ds.td;.proc.procname;`$ string .wdb.currentpartition;`access);
     atab,:() xkey .ds.access;
-    (` sv(.ds.td;.proc.procname;`access)) set atab;
+    (` sv(.ds.td;.proc.procname;`$ string .wdb.currentpartition;`access)) set atab;
 
     // update the access table in the gateway
     handles:(.servers.getservers[`proctype;`gateway;()!();1b;1b])[`w];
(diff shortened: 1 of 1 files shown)


code/wdb/datastripe.q
```diff
@@ -31,9 +31,9 @@ modaccess:{[accesstab]};
     .lg.o[`reload;"Kept ",string[.ds.periodstokeep]," period",$[.ds.periodstokeep>1;"s";""]," of data from : ",", " sv string[t]];
     
     // update the access table on disk
-    atab:get ` sv(.ds.td;.proc.procname;`access);
+    atab:get ` sv(.ds.td;.proc.procname;`$ string .wdb.currentpartition;`access);
```
0Cal-MacK0, line 34:
```suggestion
    accesspath: ` sv(.ds.td;.proc.procname;`$ string .wdb.currentpartition;`access)
    atab:get accesspath;
```

code/wdb/datastripe.q
```diff
@@ -31,9 +31,9 @@ modaccess:{[accesstab]};
...
     atab,:() xkey .ds.access;
-    (` sv(.ds.td;.proc.procname;`access)) set atab;
+    (` sv(.ds.td;.proc.procname;`$ string .wdb.currentpartition;`access)) set atab;
```
0Cal-MacK0, line 36:
```suggestion
    accesspath set atab;
```
From above comment

code/wdb/datastripe.q
```diff
@@ -50,10 +50,10 @@ initdatastripe:{
     t:tables[`.] except .wdb.ignorelist;
 
     // load the access table; fall back to generating table if load fails
-    .ds.access: @[get;(` sv(.ds.td;.proc.procname;`access));([] table:t ; start:0Np ; end:0Np ; stptime:0Np ; keycol:`sym^.wdb.tablekeycols[t])];
+    .ds.access: @[get;(` sv(.ds.td;.proc.procname;`$ string .wdb.currentpartition;`access));([] table:t ; start:0Np ; end:0Np ; stptime:0Np ; keycol:`sym^.wdb.tablekeycols[t])];
```
0Cal-MacK0, line 53:
```suggestion
    default:([]table:t; start:0Np; end:0Np; stptime:0Np; keycol:`sym^.wdb.tablekeycols t)
    .ds.access: @[get;accesspath;default];
```

code/wdb/datastripe.q
```diff
@@ -50,10 +50,10 @@ initdatastripe:{
...
     modaccess[.ds.access];
     .ds.checksegid[];
-    (` sv(.ds.td;.proc.procname;`access)) set .ds.access;       
+    (` sv(.ds.td;.proc.procname;`$ string .wdb.currentpartition;`access)) set .ds.access;
```
0Cal-MacK0, line 56:
```suggestion
    accesspath set .ds.access;       
```

code/wdb/datastripe.q
dmitchard: At that point would it not make more sense to make the accesspath a global

dmitchard: Actually what I said wouldn't work as it needs to be within one of the functions called by the wdb in order to get currentpartition

PR #506, opened 2022-08-05
Tcs 116 tests and minor changes

SeanW17:
Added json files to the test folder which can be used to show the error handling and logging easier.
Also moved code.
Also changed .ds.stripeconfig definition.


tests/stp/configchecks/additionaltests
```diff
@@ -0,0 +1,19 @@
+action,ms,bytes,lang,code,repeate,minver,comment
+before,0,0,q,.ds.stripeconfig:`not.json,1,,"Sets json config to one which should fail as no json file by this name"
+before,0,0,q,configcheck[],1,,"Runs check functions on the json file"
+true,0,0,q,.ds.datastripe=0b,1,,"Checks datastriping has not been activated by checks"
+before,0,0,q,.ds.stripeconfig:`empty.json,1,,"Sets json config to one which should fail as file contains nothing"
+before,0,0,q,configcheck[],1,,"Runs load function on the json file"
```
jonathonmcmurray, line 6: I may be mistaken but I think with a `before` action in k4unit, these will all get run before the `true` actions, so results might not be correct. I think these should possibly be `run` actions to preserve the order with the `true` actions etc.

jonathonmcmurray (approved):
LGTM

emansanti (approved):

PR #508, opened 2022-08-11
Stop wdb saving to wdbhdb when using TorQ Cloud

esweeney97:
Data should only be saved to tail directory when using TorQ Cloud.
This change checks if datastriping is enabled. If so, the timer that runs standard TorQ wdb savedown will not be started, data will only be saved to tail directory at end of period.


eimear-dalzell (approved):
looks good to me

jonathonmcmurray (approved):

PR #509, opened 2022-08-12
Tcs 117 separate wdb and tailer

ThomasGillespie:
- Created tailer process, with functionality of wdb and extra tailer functionality
- reverted wdb to torq classic


code/processes/wdb.q
```diff
@@ -391,8 +371,8 @@ endofdaysort:{[dir;pt;tablist;writedownmode;mergelimits;hdbsettings]
 	};
 
 /-function to send reload message to rdbs/hdbs
-reloadproc:{[h;d;ptype;reloadlist]
-	.wdb.countreload:count[raze .servers.getservers[`proctype;;()!();1b;0b] each reloadlist];
+reloadproc:{[h;d;ptype]
+	.wdb.countreload:count[raze .servers.getservers[`proctype;;()!();1b;0b]each reloadorder];
```
emansanti, line 375: is this just reverting the changes originally made for the tailer?

ThomasGillespie: Yes, any changes to the wdb.q reverts it to the torq classic version

code/processes/tailer.q
```diff
@@ -0,0 +1,55 @@
...
+  };
+reloadproc:{[h;d;ptype;reloadlist]
+        .wdb.countreload:count[raze .servers.getservers[`proctype;;()!();1b;0b] each reloadlist];
+        $[eodwaittime>0;
+                {[x;y;ptype].[{neg[y]@x};(x;y);{[ptype;x].lg.e[`reloadproc;"failed to reload the ",string[ptype]];'x}[ptype]]}[({@[`. `reload;x;()]; (neg .z.w)(`.wdb.handler;1b); (neg .z.w)[]};d);h;ptype];
+                @[h;(`reload;d);{[ptype;e] .lg.e[`reloadproc;"failed to reload the ",string[ptype],".  The error was : ",e]}[ptype]]
+        ];
+        .lg.o[`reload;"the ",string[ptype]," has been successfully reloaded"];
+        }
+
```
dmitchard, line 32:
```suggestion
```
Make whitespace in-between functions consistent

code/processes/tailer.q
```diff
@@ -0,0 +1,55 @@
+\d .wdb
```
dmitchard, line 1: Do we not want these functions under a tailer namespace?
```suggestion
\d .tailer
```

dmitchard (changes requested):

code/processes/tailer.q
ThomasGillespie: as some some of the functions are updating functions present in the wdb namespace, I placed the script in that namespace

dmitchard: So would you not then want the tailer specific ones in a tailer namespace rather than wdb?

eimear-dalzell (approved):

code/processes/tailer.q
```diff
@@ -0,0 +1,57 @@
...
+.proc.loadf [getenv[`KDBCODE],"/processes/wdb.q"]
+hdbsettings:(`compression`hdbdir`taildir)!(compression;hdbdir;getenv`KDBTAIL)
```
jonathonmcmurray, line 3:
```suggestion
hdbsettings[`taildir]:getenv`KDBTAIL
```

If I'm following correctly, you're really only changing this key in the dict, so no need to overwrite the whole thing

code/processes/tailer.q
```diff
@@ -0,0 +1,57 @@
...
+getprocs:{[x;y]
+        a:exec (w!x) from .servers.getservers[`proctype;x;()!();1b;0b];
+        /-exit if no valid handle
+        if[0=count a; .lg.e[`connection;"no connection to the ",(string x)," could be established... failed to reload ",string x];:()];
+        .lg.o[`connection;"connection to the ", (string x)," has been located"];
+        /-send message along each handle a
+        reloadproc[;y;value a;x] each key a;
+        }
+
+.servers.CONNECTIONS:(distinct .servers.CONNECTIONS,.wdb.hdbtypes,.wdb.rdbtypes,.wdb.gatewaytypes,.wdb.tickerplanttypes,.wdb.sorttypes,.wdb.sortworkertypes,.tailer.tailreadertypes) except `
```
jonathonmcmurray, line 46: this is updating another variable from wdb.q, so I'd put this after line modifying `hdbsettings`. Similar to that one, I don't think it's necessary to overwrite the whole variable

```suggestion
.servers.CONNECTIONS,:.tailer.tailreadertypes;
```

code/processes/tailer.q
```diff
@@ -0,0 +1,57 @@
+\d .wdb
+.proc.loadf [getenv[`KDBCODE],"/processes/wdb.q"]
+hdbsettings:(`compression`hdbdir`taildir)!(compression;hdbdir;getenv`KDBTAIL)
+
+\d .tailer
+trtype:`$"tr_",last "_" vs string .proc.proctype                           /-extract wdb proc segname and append to "tr_"
+tailreadertypes:trtype
```
jonathonmcmurray, line 7: why do we need both these variables if they have the same value?

code/processes/tailer.q
```diff
@@ -0,0 +1,57 @@
...
+\d .
+
+/- initialise datastripe
+if[.ds.datastripe;
+  .lg.o[`dsinit;"datastripe on: initialising datastripe"];
+  initdatastripe[]];
+
+/- create HDB sym file and taildir symlink
+if[.ds.datastripe;
+  .ds.symlink[]];
```
jonathonmcmurray, line 57: conditions are the same, why not just combine these?

personal preference but I also prefer to close `if` statements with the closing `]` on it's own line, just to make it super clear at a glance where it ends

```suggestion
/- initialise datastripe
if[.ds.datastripe;
  .lg.o[`dsinit;"datastripe on: initialising datastripe"];
  initdatastripe[];
  /- create HDB sym file and taildir symlink
  .ds.symlink[];
 ];
```

final point about this - is it valid to be running a tailer process with `.ds.datastripe` false? My understanding is that the tailer only makes sense with datastriping enabled, so should it instead be a case of throwing an error here if tailer is run and datastriping is *not* enabled?

config/settings/tailer.q
```diff
@@ -0,0 +1,63 @@
+// Bespoke WDB config
```
jonathonmcmurray, line 1:
```suggestion
// Bespoke tailer config
```

config/settings/tailer.q
```diff
@@ -0,0 +1,63 @@
...
+permitreload:1b                                                                             // enable reload of hdbs/rdbs
+compression:()                                                                              // specify the compress level, empty list if no required
+gc:1b                                                                                       // garbage collect at appropriate points (after each table save and after sorting data)
+eodwaittime:0D00:00:10.000                                                                  // time to wait for async calls to complete at eod
+tpcheckcycles:0W                                                                            // number of attempts to connect to tp before process is killed
+
+\d .ds
+datastripe:0b
+period:0D01
+lasttimestamp:.z.p-4*period
```
jonathonmcmurray, line 55: I don't think this should be calced here - I think there should maybe be a config value for number of periods, but calc the lasttimestamp in tailer.q, using the relevant config variables

On a side note, we need to make sure it's well documented all the places that things like period & number of periods etc. is configured.

config/settings/tailer.q
```diff
@@ -0,0 +1,63 @@
...
+hdbdir:`:hdb                                                                                // move wdb database to different location
+sortcsv:hsym first .proc.getconfigfile"sort.csv"                                            // location of csv file
+permitreload:1b                                                                             // enable reload of hdbs/rdbs
+compression:()                                                                              // specify the compress level, empty list if no required
+gc:1b                                                                                       // garbage collect at appropriate points (after each table save and after sorting data)
+eodwaittime:0D00:00:10.000                                                                  // time to wait for async calls to complete at eod
+tpcheckcycles:0W                                                                            // number of attempts to connect to tp before process is killed
+
+\d .ds
+datastripe:0b
```
jonathonmcmurray, line 53: Should datastriping be disabled by default? Mentioned above, but my understanding of tailer is that it only makes sense when we're using datastriping (otherwise we'd just use the regular wdb), so not sure it makes sense that datastriping is turned off by default

jonathonmcmurray (commented):
Few thoughts here but overall I think this looks good

code/processes/tailer.q
ThomasGillespie: the `distinct` at the beginning ensures that no values are repeated, so could it be the case that .tailer.tailreadertypes already exists inside .server.CONNECTIONS?

if this is the case .tailer.tailreadertypes might end up repeated.

would `.servers.CONNECTIONS: distinct .servers.CONNECTIONS,.tailer.tailreadertypes` be okay?

config/settings/tailer.q
ThomasGillespie: At the moment I have not included wdbs in the start all for torq5, as such they will not be running by default, this is why I thought it was okay for tailers to accept unstriped data.

would it be better to have the wdbs starting as well, then stop either the wdbs, if datastriping is off, or the tailers, if it is on?

this would also remove the need for the previous ifs.

jonathonmcmurray: Perhaps I'm confused but I think tailers should always have datastriping on, and then we control whether we're using tailers (& therefore datastriping) or wdb(s) (& therefore no datastriping) by which we processes we start (i.e. by changing the start all stuff in process.csv)

To me, it seems confusing if we have the wdb as a separate proctype, but the tailer can also be used without datastriping - in which case it does the same job as a wdb(?), so why would we need to have both existing?

ThomasGillespie: Okay, that sounds good

dmitchard (approved):

PR #511, opened 2022-08-22
TCS-137 - Filtered Tailer Savedown on Timer

CONeillAQ:
Tailer functionality for only saving tables from memory that pass a configurable threshold. 
Check this condition on a configurable timer.
Improve savedown logging.
Reload TailReader tables from TDB.


code/processes/tailer.q
```diff
@@ -50,9 +51,10 @@ getprocs:{[x;y]
...
-initdatastripe[];
-
+if[.ds.datastripe;
+  .lg.o[`dsinit;"datastripe on: initialising datastripe"];
+  initdatastripe[]];
 
 /- create HDB sym file and taildir symlink
-.ds.symlink[];
+if[.ds.datastripe;
+  .ds.symlink[]];
```
SeanW17, line 60: Shouldn't be needed as tailer will only be on if .ds.datastripe is enabled in the STP.

code/processes/tailer.q
```diff
@@ -50,9 +51,10 @@ getprocs:{[x;y]
 \d .
 
 /- initialise datastripe
-.lg.o[`dsinit;"initialising datastripe"];
-initdatastripe[];
-
+if[.ds.datastripe;
+  .lg.o[`dsinit;"datastripe on: initialising datastripe"];
+  initdatastripe[]];
```
SeanW17, line 56: Shouldn't be needed as tailer will only be on if .ds.datastripe is enabled in the STP.

code/tailer/datastripe.q
```diff
@@ -118,33 +112,40 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
...
     };
 
-savealltablesoverperiod:{[dir;nextp;lasttime]
+savealltablesoverperiod:{[]
     /- function takes the tailer hdb directory handle and a timestamp
     /- saves each table up to given period to their respective partitions
-    savetablesoverperiod[dir;;nextp;lasttime]each .wdb.tablelist[];
+    dir:.ds.td; 
+    nextp:.z.p;
+    lasttime:.z.p-`second$.wdb.period;
```
SeanW17, line 134: Do these need defined on separate lines or can they be done within the functions they are used in?
Do you need to substitute the names or can you just use .ds.td and .z.p in most lines?

code/tailer/datastripe.q
```diff
@@ -118,33 +112,40 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
...
+    totals:{count get x}each .wdb.tablelist[];
+    .wdb.timepartition:.z.p;
```
SeanW17, line 136: Is this variable used?

code/tailer/datastripe.q
```diff
@@ -118,33 +112,40 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
...
+    if[max totals>.wdb.rowthresh;
+        .lg.o[`save;"Saving ",(", " sv string .wdb.tablelist[] where totals>.wdb.rowthresh)," table(s)"];
+        savetablesoverperiod[dir;;nextp;lasttime]each (.wdb.tablelist[] where totals>.wdb.rowthresh);
     /- trigger reload of access tables and intradayDBs in all tail reader processes
-    .tailer.dotailreload[`]};
+    .tailer.dotailreload[`]
+    ];  
+    if[min totals<.wdb.rowthresh;
+        .lg.o[`save;"No tables above threshold, no tables saved"]
```
SeanW17, line 144: Where is rowthresh defined and should it be quoted in this comment?

code/tailer/datastripe.q
```diff
@@ -118,33 +112,40 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
...
-
     /- enumerate and then split by keycol
     symdir:` sv dir,.proc.procname;
     enumkeycol: .Q.en[symdir;?[tablename;enlist (<;`time;nextp);0b;()]];
     splitkeycol: {[enumkeycol;keycol;s] ?[enumkeycol;enlist (=;keycol;enlist s);0b;()]}[enumkeycol;keycol;] each partitionlist;
-
     /-upsert table to partition
-    upserttopartition[dir;tablename;keycol;;nextp] each splitkeycol where 0<count each splitkeycol; 
-
+    upserttopartition[dir;tablename;keycol;;nextp] each splitkeycol where 0<count each splitkeycol;
```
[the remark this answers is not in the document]
CONeillAQ: Unsure as to why this is listed as a change, although it could have been a result of how a merge conflict was resolved. Regardless, this line works unchanged.

code/processes/tailer.q
CONeillAQ: No worries, will streamline this by the end of the day

code/processes/tailer.q
CONeillAQ: No worries, will streamline this by the end of the day

code/tailer/datastripe.q
CONeillAQ: The names were substituted to improve readability but it should work without the replacement.

code/tailer/datastripe.q
aidanloughranAQ: That seems to be a variable we don't use anymore so we can take this out

code/tailer/datastripe.q
CONeillAQ: rowthresh is defined in the tailer settings file, will add a comment now

code/processes/tailer.q
```diff
@@ -51,10 +51,8 @@ getprocs:{[x;y]
...
-  .lg.o[`dsinit;"datastripe on: initialising datastripe"];
-  initdatastripe[]];
+.lg.o[`dsinit;"initialising datastripe"];
+initdatastripe[];
 
 /- create HDB sym file and taildir symlink
-if[.ds.datastripe;
-  .ds.symlink[]];
\ No newline at end of file
+.ds.symlink[]];
```
SeanW17: Think you still have an extra right square bracket.

SeanW17: Nevermind seen your next commit.

SeanW17 (approved):

code/tailer/datastripe.q
```diff
@@ -118,33 +112,40 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
...
+    ];
+    /- if totals are below rowthresh log message indicates no tables saved
+    if[min totals<.wdb.rowthresh;
+        .lg.o[`save;"No tables above threshold, no tables saved"]
+    ];
+    };
 
-.timer.repeat[00:00+.z.d;0W;0D00:10:00;(`.ds.savealltablesoverperiod;.ds.td;.z.p);"Saving tables"]
+/- Timer to repeat savealltablesoverperiod with period defined in tailer.q settings
+.timer.repeat[00:00+.z.d;0W;.wdb.period;(`.ds.savealltablesoverperiod;`);"Saving tables"]
```
jonathonmcmurray, line 148: should we be using `.wdb.period` here? I think the "typical" period that will be used is 1 hour, so we'll only check once per hour if we've met the threshold?

how often does the regular wdb check this?

config/settings/tailer.q
```diff
@@ -50,7 +50,9 @@ eodwaittime:0D00:00:10.000
 tpcheckcycles:0W                                                                            // number of attempts to connect to tp before process is killed
 
 \d .ds
+datastripe:0b
 period:0D01
+lasttimestamp:.z.p-4*period
```
jonathonmcmurray, line 56: are these changes relevant to this PR? I don't understand why they're here

tests/tailer/savedown_withcount/savedown.csv
```diff
@@ -0,0 +1,20 @@
+action,ms,bytes,lang,code,repeate,minver,comment
+before,0,0,q,tr:.servers.gethandlebytype[`tr_seg1;`any],1,,"Connect to tailreader handle to give access the tailreader1 for testing tailer1"
+before,0,0,q,tl:.servers.gethandlebytype[`tailer_seg1;`any],1,,"Connect to tailer1 handle,so tailer can be tested through tl handle"
+run,0,0,q,(tr"count tables[]")=2,1,,"Check the number of tables in the tailreader is currently 2"
+run,0,0,q,tl".ds.savealltablesoverperiod[]",1,,"Running savedown function sould save tables but, there should be no data to savedown so the table count should stay the same"
+run,0,0,q,(tr"count tables[]")=2,1,,"Check the number of tables in the tailreader has remained the same indicating the savedown didn't add any tables as there was none to add"
+run,0,0,q,system "sleep 10",1,,"Wait 10 seconds to populate tables with rows of data for trade and quote"
```
jonathonmcmurray, line 7: torq has .os.sleep

```suggestion
run,0,0,q,.os.sleep 10,1,,"Wait 10 seconds to populate tables with rows of data for trade and quote"
```

code/tailer/datastripe.q
```diff
@@ -118,33 +112,40 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
...
 
-savealltablesoverperiod:{[dir;nextp;lasttime]
+savealltablesoverperiod:{[]
     /- function takes the tailer hdb directory handle and a timestamp
     /- saves each table up to given period to their respective partitions
-    savetablesoverperiod[dir;;nextp;lasttime]each .wdb.tablelist[];
+    /- totals calculates the row count of tables in .wdb.tablelist[] 
+    totals:{count get x}each .wdb.tablelist[];
+    /- If a totals value is greater than rowthresh which is the rowcount for saving down tables
+    if[max totals>.wdb.rowthresh;
```
jonathonmcmurray, line 135: IMO this should be either

```suggestion
    if[max[totals]>.wdb.rowthresh;
```

or

```suggestion
    if[any totals>.wdb.rowthresh;
```

Technically these will all serve the same purpose, but I think it is semantically clearer to use `any` here

code/tailer/datastripe.q
```diff
@@ -118,33 +112,40 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
...
     .gc.run[];
     };
 
-savealltablesoverperiod:{[dir;nextp;lasttime]
+savealltablesoverperiod:{[]
     /- function takes the tailer hdb directory handle and a timestamp
     /- saves each table up to given period to their respective partitions
-    savetablesoverperiod[dir;;nextp;lasttime]each .wdb.tablelist[];
+    /- totals calculates the row count of tables in .wdb.tablelist[] 
+    totals:{count get x}each .wdb.tablelist[];
```
jonathonmcmurray, line 133: `get` is for loading tables from disk, I think `value` is more correct here (functionally identical but conventionally `value` is used to evaluate a symbol/string, so it's clearer what's happening in this line if we use `value`)

```suggestion
    totals:{count value x}each .wdb.tablelist[];
```

code/tailer/datastripe.q
```diff
@@ -118,33 +112,40 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
...
+    /- If a totals value is greater than rowthresh which is the rowcount for saving down tables
+    if[max totals>.wdb.rowthresh;
+        .lg.o[`save;"Saving ",(", " sv string .wdb.tablelist[] where totals>.wdb.rowthresh)," table(s)"];
+        savetablesoverperiod[.ds.td;;.z.p;(.z.p-`second$.wdb.period)]each (.wdb.tablelist[] where totals>.wdb.rowthresh);
     /- trigger reload of access tables and intradayDBs in all tail reader processes
-    .tailer.dotailreload[`]};
+    .tailer.dotailreload[`]
+    ];
+    /- if totals are below rowthresh log message indicates no tables saved
+    if[min totals<.wdb.rowthresh;
```
jonathonmcmurray, line 142: as above but

```suggestion
    if[not any totals<.wdb.rowthresh;
```

or

```suggestion
    if[all totals>.wdb.rowthresh;
```

I would also move this conditional to the start of the function & return early from it (i.e. `:();`), then have the code that does the save *outside* any conditional (because obviously if the first conditional is false, the other is true, so don't necessarily need to specify it)

jonathonmcmurray (commented):
Please make sure to have a descriptive PR title - it's not clear to me what the current title means

There seem to be a few unrelated/non-functional changes in the PR - always best to keep the PR as clean as possible with only the changes that are actually relevant, to make it as easy as possible to review

On a larger point - the wdb already has the functionality to check row thresholds on a timer - is there any way we can use the existing functionality or do we _need_ to reimplement it here?

code/tailer/datastripe.q
CONeillAQ: This has been updated to use `value`

code/tailer/datastripe.q
CONeillAQ: This has been restructured to only use one conditional, the semantics of which has been improved as per this suggestion to use `all`

code/tailer/datastripe.q
CONeillAQ: This conditional has been removed, and the function has been restructured to only use one conditional

tests/tailer/savedown_withcount/savedown.csv
CONeillAQ: This has been fixed

config/settings/tailer.q
CONeillAQ: These lines have now been removed

eimear-dalzell (approved):

code/tailer/datastripe.q
```diff
@@ -89,22 +89,16 @@ upserttopartition:{[dir;tablename;keycol;enumdata;nextp]
     /- function takes a (dir)ectory handle, tablename as a symbol
     /- column to key table on, an enumerated table and a timestamp.
     /- partitions the data on the keycol and upserts it to the given dir
-
```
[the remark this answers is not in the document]
jonathonmcmurray: this was marked as resolved, but it's still true that there are non-functional whitespace changes

code/tailer/datastripe.q
```diff
@@ -151,17 +151,26 @@ savetables:{[dir;tablename]
...
+    totals:{count value x}each .wdb.tablelist[];
+    /- log and return from function early if no table has crossed threshold
+    if[all totals<.wdb.numrows;
+        .lg.o[`save;"No tables above threshold, no tables saved"];
+        :();
+        ];
+
+    /- log and savedown any tables above threshold
+    .lg.o[`save;"Saving ",(", " sv string .wdb.tablelist[] where totals>.wdb.numrows)," table(s)"];
+    savetables[dir;]each (.wdb.tablelist[] where totals>.wdb.numrows);
```
jonathonmcmurray, line 167:
```suggestion
    .lg.o[`save;"Saving ",(", " sv string tabstosave:.wdb.tablelist[] where totals>.wdb.numrows)," table(s)"];
    savetables[dir;]each tabstosave;
```

Don't need to derive same list of tables twice

code/tailer/datastripe.q
```diff
@@ -151,17 +151,26 @@ savetables:{[dir;tablename]
...
-    /- delete data that has been saved
-    @[`.;;0#] each .wdb.tablelist[];
-	
+    /- totals calculates the row count of tables in .wdb.tablelist[]
+    totals:{count value x}each .wdb.tablelist[];
+    /- log and return from function early if no table has crossed threshold
+    if[all totals<.wdb.numrows;
+        .lg.o[`save;"No tables above threshold, no tables saved"];
+        :();
+        ];
```
jonathonmcmurray, line 163:
```suggestion
    ];
```

IMO the closing `]` of a multi-line `if` should be aligned with the `i` in opening line, just makes it clearer when scanning the file/function quickly

code/tailer/datastripe.q
```diff
@@ -151,17 +151,26 @@ savetables:{[dir;tablename]
     .gc.run[];
     };
 
-savealltables:{[dir]
+savealltablesoverperiod:{[dir;lasttime]
```
jonathonmcmurray, line 154: I can't see where `lasttime` is used, am I just being blind?

tests/tailer/savedown_withcount/settings.q
```diff
@@ -0,0 +1,9 @@
+// IPC connection parameters
+.servers.CONNECTIONS:`segmentedtickerplant`sort`gateway`rdb`hdb`tailer_seg1`tr_seg1
+.servers.USERPASS:`admin:admin;
```
jonathonmcmurray, line 3: the test proc has the `tailer` proctype, so surely it should be automatically using the tailer username & pass? i.e. this line shouldn't be needed I don't think

tests/tailer/savedown_withcount/tests.csv
```diff
@@ -0,0 +1,16 @@
+action,ms,bytes,lang,code,repeate,minver,comment
+before,0,0,q,tr:.servers.gethandlebytype[`tr_seg1;`any],1,,"Connect to tailreader handle to give access the tailreader1 for testing tailer1"
+before,0,0,q,tl:.servers.gethandlebytype[`tailer_seg1;`any],1,,"Connect to tailer1 handle,so tailer can be tested through tl handle"
+before,0,0,q,tl".wdb.numrows:100",1,,"Set row count as the standard row count is 10000 would be to large for a test"
+run,0,0,q,(tr"count tables[]")=2,1,,"Check the number of tables in the tailreader is currently 2"
+run,0,0,q,tl".ds.savealltablesoverperiod[]",1,,"Running savedown function sould save tables but, there should be no data to savedown so the table count should stay the same"
```
jonathonmcmurray, line 6: if I'm not mistaken, this function takes two parameters? how does this test work if that's the case?

tests/tailer/savedown_withcount/tests.csv
```diff
@@ -0,0 +1,16 @@
...
+run,0,0,q,(tr"count tables[]")=2,1,,"Check the number of tables in the tailreader has remained the same indicating the savedown didn't add any tables as there was none to add"
+run,0,0,q,.os.sleep 10,1,,"Wait 10 seconds to populate tables with rows of data for trade and quote"
+run,0,0,q,tl"![;();0b;`symbol$()]`trade";tl".ds.savealltablesoverperiod[]",1,,"Running savedown function on tables,should save quote"
```
jonathonmcmurray, line 9: Why the functional form here? Surely we can just do this?

```suggestion
run,0,0,q,tl"delete from `trade";tl".ds.savealltablesoverperiod[]",1,,"Running savedown function on tables,should save quote"
```

it's also not explained in the comment why we're deleting from trade table

tests/tailer/savedown_withcount/tests.csv
```diff
@@ -0,0 +1,16 @@
+action,ms,bytes,lang,code,repeate,minver,comment
+before,0,0,q,tr:.servers.gethandlebytype[`tr_seg1;`any],1,,"Connect to tailreader handle to give access the tailreader1 for testing tailer1"
+before,0,0,q,tl:.servers.gethandlebytype[`tailer_seg1;`any],1,,"Connect to tailer1 handle,so tailer can be tested through tl handle"
+before,0,0,q,tl".wdb.numrows:100",1,,"Set row count as the standard row count is 10000 would be to large for a test"
+run,0,0,q,(tr"count tables[]")=2,1,,"Check the number of tables in the tailreader is currently 2"
```
jonathonmcmurray, line 5: this seems very config dependent etc., I'm not sure we should be failing a test if e.g. there are more than 2 tables?

jonathonmcmurray (commented):
few questions here, particularly around the tests

code/tailer/datastripe.q
CONeillAQ: I believe this is just a vestige of how the function used to operate, this can be safely removed

code/tailer/datastripe.q
```diff
@@ -151,17 +151,26 @@ savetables:{[dir;tablename]
     .gc.run[];
     };
 
-savealltables:{[dir]
+savealltablesoverperiod:{[dir]
```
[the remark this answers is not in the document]
ThomasGillespie: other than that, looks good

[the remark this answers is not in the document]
CONeillAQ: I have now updated this branch to be more compliant with the changes made in #514

code/tailer/datastripe.q
```diff
@@ -154,14 +154,26 @@ savetables:{[dir;tablename]
 savealltables:{[dir]
     /- function takes the tailer hdb directory handle and a timestamp
     /- saves each table up to given period to their respective partitions
-    savetables[dir;]each .wdb.tablelist[];
-	
+    /- totals calculates the row count of tables in .wdb.tablelist[]
+    totals:{count value x}each .wdb.tablelist[];
+    /- log and return from function early if no table has crossed threshold
+    if[all totals<.wdb.numrows;
```
dmitchard, line 160: As the savealltables function is used by both eod and eop, this condition will cause issues at eop and eod where the save actually might not occur because of it. I think this needs to be seperated into its own function outside of savealltables so that the functionality is just for saving and nothing else.

dmitchard (changes requested):
As the savealltables function is used by both eod and eop, this condition will cause issues at eop and eod where the save actually might not occur because of it. I think this needs to be seperated into its own function outside of savealltables so that the functionality is just for saving and nothing else.

ThomasGillespie (approved):

dmitchard (approved):

jonathonmcmurray (approved):

PR #513, opened 2022-08-23
TCS-100 Create EOD process to move data from tailDBs to HDB

eimear-dalzell:
Creates new EOD process which loads data from the tailDBs and saves this data to the HDB. Process is triggered when it receives an endofday message from the tailers (TCS-75).
This process compiles the data from all tailDBs into one HDB date partition. Once data from all tailDBs are saved to HDB, the process applies the p attribute to the relevant columns in the on-disk HDB tables and deletes the data from the tailDBs.

Also resolves jira ticket TCS-75 - create EOD message to move the intraday data to the HDB.
End of day functions added to the tailer to flush its remaining in-memory data to disk and then send an endofday message to the EOD process.


code/processes/tailer.q
```diff
@@ -46,9 +47,40 @@ getprocs:{[x;y]
...
+
+    /- find handle to send message to eod process
+    eodp:exec w from .servers.getservers[`proctype;.tailer.eodtypes;()!();1b;0b];
+
+    /- exit early if no eod process connected
+    if[0=count eodp;.lg.e[`connection;"no connection to the ",(string .tailer.eodtypes)," could be established, failed to send end of day message"];:()];
+
+    /- send procname to eod process so it loads correct TDB
+    procname:.proc.procname;
+    neg[first eodp](`endofday;pt;procname);
```
emansanti, line 69: Is this first in here just because eodp is a list? Am guessing only one eodprocess is connected at any point.

code/processes/tailer.q
```diff
@@ -46,9 +47,40 @@ getprocs:{[x;y]
...
+endofdaysave:{[processdata]
+  /- function to flush remaining in-memory data to disk when end of day message is received
+  .lg.o[`save;"saving the ",(", " sv string tl:.wdb.tablelist[],())," table(s) to disk"];
+  currp:first exec distinct end from .ds.access where end<>0N;
+  .wdb.datastripeendofperiod[currp;.z.p;processdata[],(enlist `p)!enlist .z.p];
+  .lg.o[`savefinish;"finished saving remaining data to disk"];
+  };
+
+/- add endofday to tailer namespace and overwrite .wdb.endofday function
+.tailer.endofday:.wdb.endofday:endofday;
```
emansanti, line 82: Why does wdb.endofday need to be updated here? Is this a hangover from having wdb.q loaded into tailer.q?

code/processes/tailer.q
emansanti, line 69: Is this first in here just because eodp is a list? Am guessing only one eodprocess is connected at any point.

code/processes/tailer.q
emansanti, line 82: Why does wdb.endofday need to be updated here? Is this a hangover from having wdb.q loaded into tailer.q?

code/processes/tailer.q
eimear-dalzell: Yeah you're right that's just because eodp is a list, when secondary eod sort processes do eventually get added they should only connect to the main eodprocess so there should only ever be one connected to the tailer

code/processes/tailer.q
eimear-dalzell: Yep, because the endofday functions from wdb.q are loaded in I wanted to make sure that the only endofday function present in the tailer is the one defined here

code/processes/eodprocess.q
```diff
@@ -0,0 +1,90 @@
...
+  /-reset savescompleted counter and .eod.taildirs
+  savescompleted::0;
+  .eod.taildirs:();
+  };
+
+upserttopartition:{[dir;pt;tabname]
+  /-create directory path data will be saved to
+  /-need to delete int col from tables before they can be saved to HDB
+  directory:(` sv .Q.par[dir;pt;tabname],`);
+  data:delete int from select from tabname;
```
emansanti, line 53: I think assigning data in this way increases the memory usage of the process significantly. This may be the best approach though as deleting the column on disk or using os deldir or something similar might be tricky as then you have to update .d and other issues. There may be a better solutions somewhere though.

code/processes/eodprocess.q
```diff
@@ -0,0 +1,90 @@
...
+  /-upsert table to directory partition
+  upserttopartition[dir;pt;] each tablist;
+  };
+
+addpattr:{[hdbdir;pt;tabname]
+  /-load column to add p attribute on
+  pcol:.ds.loadtablekeycols[][tabname];
+
+  /-add p attr to on-disk table
+  addattr:{[hdbdir;pt;tabname;pcol] @[.Q.par[hdbdir;pt;tabname];pcol;`p#]};
```
emansanti, line 77: juts to be consistent with the rest of the script might be better to put
`@[.Q.par[hdbdir;pt;tabname];pcol;`p#]};`
on a new line

code/processes/eodprocess.q
```diff
@@ -0,0 +1,90 @@
...
+  .eod.taildirs,:taildir;
+  .lg.o[`load;"loading TDB data from ",string[procname]];
+  @[.Q.l; taildir; {[e] .lg.e[`load;"Failed to load TDB : ",e]}];
+
+  /-define list of tables to be saved to HDB
+  tablelist:tables[] except .eod.ignorelist;
+
+  /-save tables to HDB
+  savetables[.eod.hdbdir;pt;tablelist];
+  savescompleted+::1;
```
emansanti, line 31: So just to check, this saves completed gets updated each time one of the tailers triggers the eod sort and once all tailers have triggered the eod sort the savecompleted runs to delete directories and reset the variables?

emansanti (changes requested):
Looks good to me, just a few minor comments with the main one being the int delete line.

code/processes/eodprocess.q
eimear-dalzell: I believe the int column comes the subpartitions for each sym in the tailDB so that column gets added when the tailDB is loaded into memory. Because of this I couldn't find an easy way around it as it seems to come from the structure of the tailDB and just using `delete int from tabname` was giving a 'par error

code/processes/eodprocess.q
eimear-dalzell: Sounds good, I'll make that change

code/processes/eodprocess.q
eimear-dalzell: Yes, so `savescompleted` acts as a counter as we expect the same number of saves as we have tailDBs, so it gets updated every save and once all tailDBs have been saved it'll trigger the `savecomplete` function to add any attributes and delete the tailDB directories. And yeah the variables are reset within `savecomplete` too so the the next day's save can run without issues.
We discussed in the demo this counter approach probably isn't the best method as there could be some issues if something needs saved twice or something, but I think Gary said he'd be happy enough with it as a starting point

code/processes/eodprocess.q
emansanti: Oh forgot it is partitioned by int on disk so probably can't delete the int column. I will have a look at the savedown and see if there is I can get it to partition by sym without needing a duplicate int column.

code/processes/eodprocess.q
emansanti: Okay great. Yeah that makes sense, worth getting it working for now at least.

code/processes/eodprocess.q
jonathonmcmurray: Haven't looked in depth here, but I believe it should be similar to how the classic TorQ sort process handles things if wdb is in partbyattr mode (where intraday data is partitioned).

The code involved in merging at EOD is in https://github.com/AquaQAnalytics/TorQ/blob/master/code/common/merge.q - might be worth having a look at that and seeing if we can reuse the code from there rather than re-implementing from scratch

eimear-dalzell: okay great I'll have a look at that, thanks Jonathon

esweeney97 (approved):
This all looks good to me

code/processes/tailer.q
```diff
@@ -46,9 +47,28 @@ getprocs:{[x;y]
...
 
+/- eod - send end of day message to main tailsort process
+endofday:{[pt;processdata]
+  .lg.o[`eod;"end of day message received - ",spt:string pt];
+  /- call datastripeendofday
+  .wdb.datastripeendofday[pt;processdata];
+  /- find handle to send message to tailsort process
+  ts:exec w from .servers.getservers[`proctype;.tailer.tailsorttypes;()!();1b;0b];
+  /- exit early if no tailsort process connected
+  if[0=count ts;.lg.e[`connection;"no connection to the ",(string .tailer.tailsorttypes)," could be established, failed to send end of day message"];:()];
```
jonathonmcmurray, line 62: what happens if we don't have a tailsort process? I know this line will throw an error and return early, but what state will the data be left in? will we still be able to query the data for this day etc.?

code/processes/tailer.q
```diff
@@ -46,9 +47,28 @@ getprocs:{[x;y]
...
+  /- send procname to tailsort process so it loads correct tailDB
+  procname:.proc.procname;
+  neg[first ts](`endofday;pt;procname);
```
jonathonmcmurray, line 65:
```suggestion
  neg[first ts](`endofday;pt;.proc.procname);
```

code/processes/tailsort.q
```diff
@@ -0,0 +1,77 @@
...
+mergebypart:{[dir;pt;tabname;dest]
+  /-function to merge table partitions from tailDB and save to HDB
+  /-get list of partitions to be merged
+  parts:(key dir) except `access;
+  partdirs:{` sv (x;y;z)}[dir;;tabname] each parts;
+  /-load data from each partition
+  .lg.o[`merge;"reading partition(s) ", (", " sv string[partdirs])];
+  data:get each partdirs;
+  /-if multiple partitions have been read in data will be a list of tabs, if this is the case - join into single tab
+  if[98<>type data;data:(,/)data];
```
jonathonmcmurray, line 41:
```suggestion
  if[98<>type data;data:raze data];
```

`raze` is the same as `(,/)` but more recognisable

code/processes/tailsort.q
```diff
@@ -0,0 +1,77 @@
...
+  /-upsert data to partition in destination directory
+  dest:` sv .Q.par[dest;pt;tabname],`;
+  .[upsert;
+    (dest;data);
+    {[e] .lg.e[`upserttopartition;"failed to save table to disk : ",e];'e}
```
jonathonmcmurray, line 46: is it intended that the error is signalled here? Or do we want to log the error and continue?

code/processes/tailsort.q
```diff
@@ -0,0 +1,77 @@
...
+  /-function to merge tables from subpartitions in tailDB and save to HDB
+  taildir:` sv (.ts.taildir;procname;`$string pt);
+  .ts.taildirs,:taildir;
+  /-merge tables from tailDBs and save to HDB
+  mergebypart[taildir;pt;;.ts.hdbdir] each .ts.savelist;
+  /-increase savescompleted counter
+  savescompleted+::1;
+  .lg.o[`sortcomplete;"end of day sort complete for ",string[procname]];
+  /-check if all eod saves have been completed, if so trigger savecomplete
+  if[savescompleted = count .ts.taildbs;savecomplete[pt;.ts.savelist]];
```
jonathonmcmurray, line 29: if I'm reading this right, this means we can only have one tailsort process which will do EOD sort for all taildbs? assume that is intended and having multiple tailsort procs to parallelise EOD sort will be a future enhancement?

code/processes/tailsort.q
```diff
@@ -0,0 +1,77 @@
...
+  };
+
+addpattr:{[hdbdir;pt;tabname]
+  /-load column to add p attribute on
+  pcol:.ds.loadtablekeycols[][tabname];
+  /-add p attr to on-disk table
+  .lg.o[`attr;"adding p attribute to the ",string[pcol]," col in ",string[tabname]];
+  addattr:{[hdbdir;pt;tabname;pcol]
+    @[.Q.par[hdbdir;pt;tabname];pcol;`p#]
+    };
```
jonathonmcmurray, line 66:
```suggestion
  };
```

align `}` with the start of the function declaration on line 64

code/processes/tailsort.q
```diff
@@ -0,0 +1,77 @@
...
+
+savescompleted:0;                                                          /-variable to count how many tailDBs have been saved to HDB
+
+endofday:{[pt;procname]
+  /- function to trigger data load & save to HDB once endofday message is received from tailer(s)
+  .lg.o[`endofday;"end of day message received from ",string[procname]," - ",string[pt]];
+  loadandsave[pt;procname];
+  };
+
+loadandsave:{[pt;procname]
```
jonathonmcmurray, line 19: not just this function, but throughout this file convention would be to re-arrange so that functions are defined before they are used e.g. `loadandsave` is used inside `endofday`, so `loadandsave` should be defined above it in the file etc.

tests/tailsort/loadandsave/tests.csv
```diff
@@ -0,0 +1,11 @@
+action,ms,bytes,lang,code,repeat,minver,comment
+before,0,0,q,.ts.hdbdir:`:/home/edalzell/cloudTorQ/TorQ/tests/tailsort/loadandsave/hdb,1,,"set test HDB path"
+before,0,0,q,.ts.taildir:`:/home/edalzell/cloudTorQ/TorQ/tests/tailsort/loadandsave,1,,"set test tailDB path"
```
jonathonmcmurray, line 3: shouldn't have absolute file paths in tests, use env vars

tests/tailsort/loadandsave/tests.csv
```diff
@@ -0,0 +1,11 @@
...
+before,0,0,q,pt:2022.08.19,1,,"set partition for loadandsave function"
+before,0,0,q,procname:`tailer1`tailer2,1,,"set procnames for loadandsave function to locate tailDBs"
+run,0,0,q,loadandsave[pt;] each procname,1,,"run loadandsave function for both tailDBs"
```
jonathonmcmurray, line 6:
```suggestion
run,0,0,q,loadandsave[pt;] each `tailer1`tailer2,1,,"run loadandsave function for both tailDBs"
```

tests/tailsort/loadandsave/tests.csv
```diff
@@ -0,0 +1,11 @@
...
+true,0,0,q,(`$string pt) in key .ts.hdbdir,1,,"check date partition was created in HDB"
+true,0,0,q,`quote in key ` sv (.ts.hdbdir;`$string pt),1,,"check quote table is present in HDB partition"
+true,0,0,q,`trade in key ` sv (.ts.hdbdir;`$string pt),1,,"check trade table is present in HDB partition"
+true,0,0,q,not (`$string pt) in key ` sv (.ts.taildir;`wdb1),1,,"check date partition was deleted in tailDB"
+true,0,0,q,not (`$string pt) in key ` sv (.ts.taildir;`wdb2),1,,"check date partition was deleted in tailDB"
```
jonathonmcmurray, line 11: I think these should be `tailer1` and `tailer2` instead of `wdb1` and `wdb2`?

```suggestion
true,0,0,q,not (`$string pt) in key ` sv (.ts.taildir;`tailer1),1,,"check date partition was deleted in tailDB"
true,0,0,q,not (`$string pt) in key ` sv (.ts.taildir;`tailer2),1,,"check date partition was deleted in tailDB"
```

code/processes/tailsort.q
eimear-dalzell: okay yeah I'll change the order of functions in this file, thanks

code/processes/tailsort.q
eimear-dalzell: yeah with this setup we will only be able to have one tailsort process to do all the work and adding secondary processes to parallelise the sort is a new jira ticket that will need done - will probably need to rethink the counter method for secondary processes but I'll detail this in the ticket

jonathonmcmurray: Sure, makes sense. Thanks

code/processes/tailer.q
jonathonmcmurray: Might be worth having a discussion on this point, and also looking at what standard TorQ WDB does in this scenario

tests/tailsort/loadandsave/tests.csv
eimear-dalzell: yep you're right - must've missed that when we split the wdb and tailers to different procs

tests/tailsort/loadandsave/tests.csv
eimear-dalzell: took these out and defined kdbtail and kdbhdb in run.sh instead

code/processes/tailsort.q
eimear-dalzell: I put the error here as I thought this was convention based on other merge functions/upserttopartition functions

code/processes/tailer.q
eimear-dalzell: I've created an endofdaysort function in the tailer to be called if the tailsort process isn't connected - it uses functions from the tailsort.q script and I've tried using a sleep command to avoid multiple tailers trying to write to the HDB at the same time

code/processes/tailer.q
```diff
@@ -46,9 +47,48 @@ getprocs:{[x;y]
...
 
 \d .
 
+endofdaysort:{[pt]
+  /- function to be called when no tailsort process can be connected, to carry out eod sort from tailer
+  .lg.o[`eodsort;"attempting to complete eod sort from tailer"];
+  /- get tailsort code
+  .proc.loadf [getenv[`KDBCODE],"/processes/tailsort.q"];
+  /- sleep allows for multiple tailers to save to HDB at different times uninterrupted
+  .os.sleep[`$first .proc.params[`segid]];
```
jonathonmcmurray, line 60: I don't think this will be very robust. Couple of issues here - I think we've decided that segment IDs may not be numerical, so this probably wouldn't work at all in that case.

Additionally, I think this is only going to stagger tailers by 1 second? what if it takes e.g. 10 seconds for a tailer to write to the hdb? or 5 mins etc.

I think for this to be robust there would need to be some mechanism to ensure that other tailers have finished their writes, rather than just waiting a bit and hoping

code/processes/tailsort.q
jonathonmcmurray: I think we need to decide on the intended behaviour here - signalling an error will mean execution is interrupted and nothing after this will execute - so in `loadandsave` we'll never update the `savescompleted` counter etc. - is that what we want to happen?

code/processes/tailsort.q
```diff
@@ -0,0 +1,77 @@
...
+  /-load column to add p attribute on
+  pcol:.ds.loadtablekeycols[][tabname];
+  /-add p attr to on-disk table
+  .lg.o[`attr;"adding p attribute to the ",string[pcol]," col in ",string[tabname]];
+  addattr:{[hdbdir;pt;tabname;pcol]
+    @[.Q.par[hdbdir;pt;tabname];pcol;`p#]
+  };
+  .[addattr;
+    (hdbdir;pt;tabname;pcol);
+    {[e] .lg.e[`attr;"Failed to add attr : ",e];e}
```
jonathonmcmurray, line 41: in this error trap, we don't signal the error, we just return it. the semi-colon on line 42 means that the return here is not used for anything, so as it is currently, there is no point in returning the error string

```suggestion
    {[e] .lg.e[`attr;"Failed to add attr : ",e]}
```

alternatively I suppose we may want to signal it like in `mergebypart` (if we do actually want to signal it there)

```suggestion
    {[e] .lg.e[`attr;"Failed to add attr : ",e];'e}
```

jonathonmcmurray (commented):
overall looks good to me, just a few questions around error handling

code/processes/tailsort.q
eimear-dalzell: yeah, I don't think this error would need to be signalled so I'll remove the returned error

code/processes/tailsort.q
jonathonmcmurray: > but I guess the issue with signalling the error would be if it happened eg during the save for the first tailer and so would prevent any saves from tailers waiting to be sorted?

I don't think it should prevent that, as those will each be triggered by separate IPC messages I think? the signal will only interrupt the excution of the current IPC message

eimear-dalzell: okay yeah, then I think we should signal the error to ensure that we don't lose the data that failed to save

jonathonmcmurray (approved):

PR #514, opened 2022-08-24
Tcs 157 eod specific to tailer

dmitchard:
- Changes access table so that they are now saved under each partition in taildir, instead of the top level of taildir
- Update save functions and remove unecessary nextp, allowing an eod function to use the table saving functions that the eop is using
- Created eod to save all remaining data to disk and clear mem tables along with move access into new partition and clear it in mem
- Minor refactoring of the datastripe script

### DIFF
--- a/code/tailer/datastripe.q
+++ b/code/tailer/datastripe.q
@@ -16,19 +16,18 @@ modaccess:{[accesstab]};
     .lg.o[`reload;"reload command has been called remotely"];
 
     // remove periods of data from tables
-    t:tables[`.] except .wdb.ignorelist;
     lasttime:nextp-.ds.periodstokeep*(nextp-currp);
 
     // update the access table in the wdb
     // on first save down we need to replace the null valued start time in the access table
     // using the first value in the saved data
-    starttimes:.ds.getstarttime each t;
+    starttimes:.ds.getstarttime each .wdb.tablelist[];
     .ds.access:update start:starttimes^start, end:?[(nextp>starttimes)&(starttimes<>0Np);nextp;0Np], stptime:data[][`p] from .ds.access;
     modaccess[.ds.access];
 
     // call the savedown function
     .ds.savealltablesoverperiod[.ds.td;lasttime];
-    .lg.o[`reload;"Kept ",string[.ds.periodstokeep]," period",$[.ds.periodstokeep>1;"s";""]," of data from : ",", " sv string[t]];
+    .lg.o[`reload;"Kept ",string[.ds.periodstokeep]," period",$[.ds.periodstokeep>1;"s";""]," of data from : ",", " sv string[.wdb.tablelist[]]];
     
     // update the access table on disk
     accesspath: ` sv(.ds.td;.proc.procname;`$ string .wdb.currentpartition;`access);
@@ -52,8 +51,7 @@ modaccess:{[accesstab]};
     //create accesspath
     accesspath: ` sv(.ds.td;.proc.procname;`$ string .wdb.currentpartition;`access);
     //make access for next partition
-    t:tables[`.] except .wdb.ignorelist;
-    .ds.access:([]table:t; start:0Np; end:0Np; stptime:0Np; keycol:`sym^.wdb.tablekeycols t);
+    .ds.access:([]table:.wdb.tablelist[]; start:0Np; end:0Np; stptime:0Np; keycol:`sym^.wdb.tablekeycols .wdb.tablelist[]);
     modaccess[.ds.access];
     accesspath set .ds.access;
     };
@@ -67,16 +65,15 @@ initdatastripe:{
 	
     // load in variables
     .wdb.tablekeycols:.ds.loadtablekeycols[];
-    t:tables[`.] except .wdb.ignorelist;
     accesspath: ` sv(.ds.td;.proc.procname;`$ string .wdb.currentpartition;`access);
 
     // load the access table; fall back to generating table if load fails
-    default:([]table:t; start:0Np; end:0Np; stptime:0Np; keycol:`sym^.wdb.tablekeycols t);
+    default:([]table:.wdb.tablelist[]; start:0Np; end:0Np; stptime:0Np; keycol:`sym^.wdb.tablekeycols .wdb.tablelist[]);
     .ds.access: @[get;accesspath;default];
     modaccess[.ds.access];
     .ds.checksegid[];
     accesspath set .ds.access;      
-    .ds.access:select by table from .ds.access where table in t;
+    .ds.access:select by table from .ds.access where table in .wdb.tablelist[];
     };
 
 \d .ds
--- a/code/tailer/datastripe.q
+++ b/code/tailer/datastripe.q
@@ -60,7 +60,7 @@ initdatastripe:{
     // update endofperiod function
     endofperiod::.wdb.datastripeendofperiod;
     
-	//update endofday function
+    //update endofday function
     endofday::.wdb.datastripeendofday;
 	
     // load in variables
--- a/code/tailer/datastripe.q
+++ b/code/tailer/datastripe.q
@@ -143,7 +143,7 @@ savetables:{[dir;tablename]
 
     /- enumerate and then split by keycol
     symdir:` sv dir,.proc.procname;
-    enumkeycol: .Q.en[symdir;eval tablename];
+    enumkeycol: .Q.en[symdir;value tablename];
     splitkeycol: {[enumkeycol;keycol;s] ?[enumkeycol;enlist (=;keycol;enlist s);0b;()]}[enumkeycol;keycol;] each partitionlist;
 
     /-upsert table to partition
--- a/code/tailer/datastripe.q
+++ b/code/tailer/datastripe.q
@@ -161,7 +161,7 @@ savealltablesoverperiod:{[dir;lasttime]
     /- saves each table up to given period to their respective partitions
     savetables[dir;]each .wdb.tablelist[];
 	
-	/- delete data from last period
+    /- delete data from last period
     .ds.deletetablebefore[;`time;lasttime]each .wdb.tablelist[];
 	
     /- trigger reload of access tables and intradayDBs in all tail reader processes
--- a/code/tailer/datastripe.q
+++ b/code/tailer/datastripe.q
@@ -27,7 +27,7 @@ modaccess:{[accesstab]};
     modaccess[.ds.access];
 
     // call the savedown function
-    .ds.savealltablesoverperiod[.ds.td;nextp;lasttime];
+    .ds.savealltablesoverperiod[.ds.td;lasttime];
     .lg.o[`reload;"Kept ",string[.ds.periodstokeep]," period",$[.ds.periodstokeep>1;"s";""]," of data from : ",", " sv string[t]];
     
     // update the access table on disk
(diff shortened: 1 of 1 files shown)


code/tailer/datastripe.q
```diff
@@ -27,33 +27,55 @@ modaccess:{[accesstab]};
...
     // update endofperiod function
     endofperiod::.wdb.datastripeendofperiod;
     
+	//update endofday function
+    endofday::.wdb.datastripeendofday;
+	
     // load in variables
     .wdb.tablekeycols:.ds.loadtablekeycols[];
     t:tables[`.] except .wdb.ignorelist;
+	accesspath: ` sv(.ds.td;.proc.procname;`$ string .wdb.currentpartition;`access);
```
SeanW17, line 71: Is there a reason for this being indented?

SeanW17, line 71:
```suggestion
accesspath: ` sv(.ds.td;.proc.procname;`$ string .wdb.currentpartition;`access);
```

SeanW17 (approved):

code/tailer/datastripe.q
```diff
@@ -27,33 +27,55 @@ modaccess:{[accesstab]};
     modaccess[.ds.access];
 
     // call the savedown function
-    .ds.savealltablesoverperiod[.ds.td;nextp;lasttime];
+    .ds.savealltablesoverperiod[.ds.td;lasttime];
     .lg.o[`reload;"Kept ",string[.ds.periodstokeep]," period",$[.ds.periodstokeep>1;"s";""]," of data from : ",", " sv string[t]];
     
     // update the access table on disk
-    atab:get ` sv(.ds.td;.proc.procname;`access);
+    accesspath: ` sv(.ds.td;.proc.procname;`$ string .wdb.currentpartition;`access);
```
emansanti, line 34: Should we be trying to move away from .wdb namesapce for currentpartition and other variables? It is fine for wdb functions I think (especially until Conor finishes the wdb-tailer split, but for variables it is a bit confusing I think.

jonathonmcmurray: IMO it depends if this variable is unique to the tailer or if it's present & referenced in wdb code - if the latter, my opinion would be it should stay in `.wdb` namespace rather than duplicating

dmitchard: The .wdb.currentpartition is used in the same way in the wdb endofday function as I have used it in the tailer's endofday

code/tailer/datastripe.q
```diff
@@ -27,33 +27,55 @@ modaccess:{[accesstab]};
...
     // update endofperiod function
     endofperiod::.wdb.datastripeendofperiod;
     
+	//update endofday function
+    endofday::.wdb.datastripeendofday;
+	
     // load in variables
     .wdb.tablekeycols:.ds.loadtablekeycols[];
     t:tables[`.] except .wdb.ignorelist;
+accesspath: ` sv(.ds.td;.proc.procname;`$ string .wdb.currentpartition;`access);
```
emansanti, line 71: Inconsistent indentation

emansanti, line 71:
```suggestion
        accesspath: ` sv(.ds.td;.proc.procname;`$ string .wdb.currentpartition;`access);
```

code/tailer/datastripe.q
```diff
@@ -27,33 +27,55 @@ modaccess:{[accesstab]};
...
     // update endofperiod function
     endofperiod::.wdb.datastripeendofperiod;
     
+	//update endofday function
+    endofday::.wdb.datastripeendofday;
+	
     // load in variables
     .wdb.tablekeycols:.ds.loadtablekeycols[];
     t:tables[`.] except .wdb.ignorelist;
+        accesspath: ` sv(.ds.td;.proc.procname;`$ string .wdb.currentpartition;`access);
```
emansanti, line 71:
```suggestion
    accesspath: ` sv(.ds.td;.proc.procname;`$ string .wdb.currentpartition;`access);
```

code/tailer/datastripe.q
```diff
@@ -27,33 +27,55 @@ modaccess:{[accesstab]};
...
+    //save all tables
+    .ds.savetables[.ds.td;] each .wdb.tablelist[];
+    //clear tables
+    @[`.;;0#] each .wdb.tablelist[];
+    //move to next partition
+    .wdb.currentpartition:pt+1;
+    //create accesspath
+    accesspath: ` sv(.ds.td;.proc.procname;`$ string .wdb.currentpartition;`access);
+    //make access for next partition
+    t:tables[`.] except .wdb.ignorelist;
```
emansanti, line 55: Is there a reason you don't use .wdb.tablelist[] here? Think it is the same, unless something happens at EOD that I have forgotten

code/tailer/datastripe.q
```diff
@@ -27,33 +27,55 @@ modaccess:{[accesstab]};
...
+    .ds.access:([]table:t; start:0Np; end:0Np; stptime:0Np; keycol:`sym^.wdb.tablekeycols t);
+    modaccess[.ds.access];
+    accesspath set .ds.access;
```
emansanti, line 58: At EOP you have `accesspath set atab` but here you have it set `.ds.access`. Is this intentional?

dmitchard: Yes atab at eop is the full on disk access table whereas at eod you want to start a new on disc access table with the blank table in .ds.access that has just been made above.

emansanti: Yeah sorry deleted my comment as I realised what was going on, think you must have commented before I deleted it.

emansanti (commented):
Very nice PR overall, just a few minor points. Functionality seems like it should work from following the code through.

code/tailer/datastripe.q
```diff
@@ -16,45 +16,64 @@ modaccess:{[accesstab]};
...
+    .ds.access:([]table:.wdb.tablelist[]; start:0Np; end:0Np; stptime:0Np; keycol:`sym^.wdb.tablekeycols .wdb.tablelist[]);
+    modaccess[.ds.access];
+    accesspath set .ds.access;
+    };
+
 initdatastripe:{
     // update endofperiod function
     endofperiod::.wdb.datastripeendofperiod;
     
+	//update endofday function
```
jonathonmcmurray, line 63: inconsistent indentation (this line uses a tab, others are spaces)

code/tailer/datastripe.q
```diff
@@ -124,28 +143,27 @@ savetablesoverperiod:{[dir;tablename;nextp;lasttime]
 
     /- enumerate and then split by keycol
     symdir:` sv dir,.proc.procname;
-    enumkeycol: .Q.en[symdir;?[tablename;enlist (<;`time;nextp);0b;()]];
+    enumkeycol: .Q.en[symdir;eval tablename];
```
jonathonmcmurray, line 146: is there a reason to use `eval` here? I think `value` would be more conventional

code/tailer/datastripe.q
```diff
@@ -124,28 +143,27 @@ savetablesoverperiod:{[dir;tablename;nextp;lasttime]
...
     /- run a garbage collection (if enabled)
     .gc.run[];
     };
 
-savealltablesoverperiod:{[dir;nextp;lasttime]
+savealltablesoverperiod:{[dir;lasttime]
     /- function takes the tailer hdb directory handle and a timestamp
     /- saves each table up to given period to their respective partitions
-    savetablesoverperiod[dir;;nextp;lasttime]each .wdb.tablelist[];
+    savetables[dir;]each .wdb.tablelist[];
```
jonathonmcmurray, line 159: I could definitely be missing something here but...

If we're not passing `lasttime` to the save function, how do we guarantee that the data that's deleted in line 162 is the same as what's been saved? it seems to me like we'll save everything in the table, but then only delete the data before `lasttime`, so the next time we save we could be duplicating some data that was saved last time we saved? (e.g. if there was already some data where `time>lasttime` when we did the first save, it wouldn't get deleted and then would be saved again in second save)

code/tailer/datastripe.q
dmitchard: Yeh, I am not sure how I used eval here..

code/tailer/datastripe.q
dmitchard: The delete on line 162 is passed lasttime, but that function is savealltablesoverperiod and and is only used by eop. Savetables is the function above and only saves the tables. EOD func and savealltablesoverperiod use the savetables function to save tables and then tables are deleted in different ways afterwards by eod and savealltablesoverperiod.

jonathonmcmurray: OK sure - but my point is, if `savetables` doesn't know about any limitations on what to save, surely it just saves everything in the table? So should line 162 just delete everything in the table? What I'm struggling to understand is if we save everything, but then only delete data before a certain time within this same function (`savealltablesoverperiod`)? And if so, why?

emansanti: There is another PR which has it changed so that after the savedown the contents of the tablelist tables are deleted, so that should resolve this issue.

jonathonmcmurray: So we have two open PRs atm that are modifying this function? Should we remove the changes to this function in one or other PR? I assume we're going to get a conflict if we try to merge both

emansanti: Yeah I am talking to the guys about it now. There was a bit of scope creep on one ticket that ended-up affecting the other, so will make sure the second PR is adapted to work with David's. The other PR needs reworking in various ways anyway before it is ready for review again anyway.

code/tailer/datastripe.q
```diff
@@ -16,45 +16,64 @@ modaccess:{[accesstab]};
...
+    modaccess[.ds.access];
+    accesspath set .ds.access;
+    };
+
 initdatastripe:{
     // update endofperiod function
     endofperiod::.wdb.datastripeendofperiod;
     
+    //update endofday function
+    endofday::.wdb.datastripeendofday;
```
eimear-dalzell, line 64: could we remove this update endofday? I've created an endofday function in tailer.q that will call .wdb.datastripeendofday so this would stop that from running

dmitchard: That would work with your code, but why do you have an end of day function in the tailer that calls an end of day function?

dmitchard: would you not just reassign it within datastripe.q the same way eop is

eimear-dalzell: at end of day the tailer needs to send a message to the sort process to move the data to hdb and going off the wdb code there was an endofday and endofdaysave - your function eliminates the need for an endofdaysave so I was calling that in endofday. but whatever setup would work best, we just need to integrate both

dmitchard: Alright no worries, that makes sense, yeh calling it from your tailer eod makes sense.

code/tailer/datastripe.q
jonathonmcmurray: I see this conversation is resolved - but what's the resolution?

dmitchard: For eod the data is saved and then all in mem tables are deleted by line 48 it does not use deletealltablesbefore. The deletealltablesbefore that uses lasttime is used by eop and is already merged into torq5. So would this not need to be a separate ticket to investigate?

dmitchard: I have updated savealltables so that it now just removes all data that has been saved both at eop and eod

jonathonmcmurray: Looks good now

jonathonmcmurray (approved):

PR #515, opened 2022-08-25
TCS-154: Moved common parameter/function definitions from processes/wdb.q to n…

Lenzaum:
- Created new script `wdb/common.q`.
- Majority of parameter/function definitions moved from `processes/wdb.q` to `wdb/common.q`.
- `processes/tailer.q` now loads `wdb/common.q` instead of `processes/wdb.q`.
    - Some end-of-script function calls duplicated from `processes/wdb.q` into `processes/tailer.q`.


emansanti (approved):
All seems correct.

ThomasGillespie (approved):

code/processes/tailer.q
```diff
@@ -1,9 +1,15 @@
...
+upd:.wdb.upd;
+
 \d .wdb
-.proc.loadf [getenv[`KDBCODE],"/processes/wdb.q"]
+
 hdbsettings[`taildir]:getenv`KDBTAIL
 .ds.lasttimestamp:.z.p-.ds.periodstokeep*.ds.period
 
-if[not .ds.datastripe;.lg.e[`load;"datastiping not enabled"]]                       /-errors out of tailer if datastriping is not turned on
+if[not .ds.datastripe;.lg.e[`load;"datastriping not enabled"]]                      /-errors out of tailer if datastriping is not turned on
```
jonathonmcmurray, line 12: think this should be moved down a bit, past the `.tailer` namespace stuff - atm the script switches to .wdb, then to .tailer, then back to .wdb, think it'd be a bit neater if each namespace was only switched to once

code/wdb/common.q
```diff
@@ -0,0 +1,528 @@
...
+.wdb.currentpartition:.wdb.getpartition[];
+
+/-  adds endofday and endofperiod functions to top level namespace
+endofday: .wdb.endofday;
+endofperiod:{[currp;nextp;data] .lg.o[`endofperiod;"Received endofperiod. currentperiod, nextperiod and data are ",(string currp),", ", (string nextp),", ", .Q.s1 data]};
+
+/- setting the upd and .u.end functions as the .wdb versions
+.u.end:{[pt]
+        .wdb.endofday[.wdb.getpartition[];()!()];
+    }
```
jonathonmcmurray, line 528: to me this stuff all feels like starting to actually initialise the process (there is actually code being run in some of these lines, not just functions/variables being defined etc.), so I think this should be in the `processes/wdb.q` script. Some of it may be duplicated in to the `processes/tailer.q` script as well, but it may or may not need all of it (e.g. not sure if `.wdb.getsortparams` is applicable to tailer) - I think some duplication is OK though. IMO this sort of "library" code shouldn't set/modify any root namespace variables (i.e. there probably shouldn't be a `\d .` in it anywhere)

code/wdb/common.q
```diff
@@ -0,0 +1,528 @@
+/-Definitions for commonly/used parameters & functions in the .wdb namespace.
```
jonathonmcmurray, line 1: I'm assuming nothing in this file has actually changed from it's original definition in `processes/wdb.q`?

jonathonmcmurray (commented):
couple of comments, but overall I think this looks good. How has this been tested to ensure that wdb & tailer both still work fully as expected?

code/wdb/common.q
Lenzaum: - It was at Gary's request that I moved in lines 515-528 from `processes/wdb.q` & `processes/tailer.q` to minimise duplication: I can revert this so that function calls are ran in these two scripts again, but I'd feel more comfortable if I had consensus from the senior dev team.
- `.wdb.startup[]` throws an error if `.wdb.getsortparams[]` has not been defined: I've traced the chain of dependencies, if an outline of them would be of help?

jonathonmcmurray: If Gary requested it, that's fine then

code/processes/tailer.q
Lenzaum: I agree that the redundant namespace switching is ugly - on reflection, I should be able to move `hdbsettings[taildir]` & `.ds.lasttimestamp` definitions down to the second instance of `\d .wdb`; and the checks on `.ds.datastripe` should be able to run in the default namespace. This would allow me to remove this first instance of `\d .wdb` without issue.

Lenzaum: Yes, proposed changes appear to work without issue - I've pushed a new commit for them.

Lenzaum:
> couple of comments, but overall I think this looks good. How has this been tested to ensure that wdb & tailer both still work fully as expected?

In addition to `wdb` & `tailer` processes starting back up, Eman & I did some manual testing to confirm the following:
- Tailer receives data;
- Latest tailer data is of the correct time;
- `.ds.savealltablesoverperiod[]` saves down to the `data/taildir` directory;
- On reloading, the tailreader picks up freshly saved data;
- In-memory `.ds.access` table gets updated after end-of-period procedures;
- Log file replays run successfully.

jonathonmcmurray (approved):

PR #518, opened 2022-09-01
Tcs 159 add segmentid to access tables

esweeney97:
Adds segment information to a proc's access table which then gets published to the gateway access table.

### DIFF
--- a/code/rdb/datastripe.q
+++ b/code/rdb/datastripe.q
@@ -27,7 +27,7 @@ initdatastripe:{
     endofperiod::.rdb.datastripeendofperiod;
     .rdb.tablekeycols:.ds.loadtablekeycols[];
     t:tables[`.] except .rdb.ignorelist;
-    .ds.access:([table:t] start:.ds.getstarttime each t; end:0Np ; stptime:0Np ; keycol:`sym^.rdb.tablekeycols[t]);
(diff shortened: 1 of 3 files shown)


esweeney97:
The tailreader just loads the access table that gets saved down by the tailer so the tailreader will already have the new column

ThomasGillespie:
oh, yeah that makes sense

ThomasGillespie (approved):

jonathonmcmurray (approved):

code/rdb/datastripe.q
```diff
@@ -27,7 +27,7 @@ initdatastripe:{
     endofperiod::.rdb.datastripeendofperiod;
     .rdb.tablekeycols:.ds.loadtablekeycols[];
     t:tables[`.] except .rdb.ignorelist;
-    .ds.access:([table:t] start:.ds.getstarttime each t; end:0Np ; stptime:0Np ; keycol:`sym^.rdb.tablekeycols[t]);
+    .ds.access:([table:t] start:.ds.getstarttime each t; end:0Np ; stptime:0Np ; keycol:`sym^.rdb.tablekeycols[t]; segment:first .ds.segmentid);
```
drgdavies, line 30: Any reason why this is first?

dmitchard: .ds.segmentid is a one item list, not really sure why, but I imagine this is why Ethan is using first to get the value

esweeney97: .ds.segmentid is a list of one symbol, was throwing a `length` error when loading the process code

PR #523, opened 2022-09-05
Add pubclear call for stpeoperiod and stpeod

dmitchard:
- Noticed that data from the very end of previous day was getting saved in the next days partition
- Added call of pubclear within stpeoperiod and stpeod to flush remaining data to subscribers before the EOP/EOD message is sent

### DIFF
--- a/code/segmentedtickerplant/stplog.q
+++ b/code/segmentedtickerplant/stplog.q
@@ -212,23 +212,14 @@ periodrollover:{[data]
   rolllog[multilog;dldir;rolltabs;data`p];
   }
 
-// endofday function defined in SCTP
-// passes on eod messages to subscribers and rolls logs
+// common eod for STP and SCTP to send out eod messages and roll logs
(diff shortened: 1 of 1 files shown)


ThomasGillespie (approved):

code/segmentedtickerplant/stplog.q
```diff
@@ -221,6 +223,8 @@ endofday:{[date;data]
 
 // STP runs function to send out eod messages and roll logs
 stpeod:{[date;data]
+  .lg.o[`endofday;"flushing remaining data to subscribers and clearing tables"];
+  .stpps.pubclear[.stplg.t];
```
jonathonmcmurray, line 227: I think this effectively makes `stpeod` functionally identical to `endofday` - should we just eliminate `stpeod` and replace any references to it with `endofday` instead?

jonathonmcmurray (approved):
LGTM

PR #527, opened 2022-09-07
TorQ5 release of torqv5_dev into torqv5_master

emansanti:
-Replaced datastriping csv files with striping.json
-Added tailer processes to perform intraday savedowns to tailDBs when in striping mode
-Added tailreader proccesses to load tailDBs and allow for on-disk data to be queried
-Updated tplog replay behaviour to replay only required tplogs and filter replayed tables accurately
-Introduces accesstables for rdbs, tailers and tailreaders to enable the gateway to accurately route queries
-Added datastriping EOD sort process to copy tailDBs to the HDB


code/tailer/datastripe.q
```diff
@@ -0,0 +1,196 @@
...
+    savetables[dir;]each tabstosave;
+
+    /- delete data that has been saved
+    @[`.;;0#] each tabstosave;
+
+    /- trigger reload of access tables and intradayDBs in all tail reader processes
+    .tailer.dotailreload[`]
+    };
+
+/- Timer to repeat savealltables with period defined in tailer.q settings
```
dmitchard, line 186:
```suggestion
/- Timer to call savedownfilter with period defined in tailer.q settings
```

dmitchard (changes requested):

code/tailer/datastripe.q
emansanti: @jonathonmcmurray This change looks good to me but I don't have permissions to commit it

jonathonmcmurray: applied

dmitchard (approved):

jonathonmcmurray (approved):

PR #529, opened 2022-09-09
Added tailreader log message to alert user if datastriping is disabled.

Lenzaum:
- Log message is written if `not .ds.datastripe` is true.

### DIFF
--- a/code/processes/tailreader.q
+++ b/code/processes/tailreader.q
@@ -7,6 +7,9 @@ currentpartition:@[value;`currentpartition;getpartition[]]
 basedir:raze (getenv`KDBTAIL),"/tailer",(string .ds.segmentid),"/"         /-define associated tailer base directory
 taildir:`$ basedir,string currentpartition;                                /-define tailDB direction
 
+/- log message if tailreader is started without datastriping activated
+if[not .ds.datastripe;.lg.o[`load;"Datastriping is disabled: please verify whether ",(string .proc.procname)," process should be running."]];
+
 \d .
 endofday:{[pt]
   /- end of day function that will be triggered by EOD Sorter once TailDB is copied to HDB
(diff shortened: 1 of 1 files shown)


code/processes/tailreader.q
```diff
@@ -7,6 +7,9 @@ currentpartition:@[value;`currentpartition;getpartition[]]
 basedir:raze (getenv`KDBTAIL),"/tailer",(string .ds.segmentid),"/"         /-define associated tailer base directory
 taildir:`$ basedir,string currentpartition;                                /-define tailDB direction
 
+/- log message if tailreader is started without datastriping activated
+if[not .ds.datastripe;.lg.o[`load;"Datastriping is disabled: please verify whether ",(string .proc.procname)," process should be running."]];
```
jonathonmcmurray, line 11: An out log might get lost/ignored in the vast amount of other out logs TorQ produces, unless someone is looking specifically for it (and if someone knows to look for this particular log, they probably also know not there's no point having a tailreader without datastriping).

Should we just log an error (which will cause the process to exit) to make it a bit more obvious for people who aren't looking?

Lenzaum: Good point: also maintains parity with a similar `.lg.e` message in the tailer script.

jonathonmcmurray (approved):
LGTM

code/processes/tailreader.q
```diff
@@ -7,6 +7,9 @@ currentpartition:@[value;`currentpartition;getpartition[]]
 basedir:raze (getenv`KDBTAIL),"/tailer",(string .ds.segmentid),"/"         /-define associated tailer base directory
 taildir:`$ basedir,string currentpartition;                                /-define tailDB direction
 
+/- log message if tailreader is started without datastriping activated
+if[not .ds.datastripe;.lg.e[`load;"Datastriping is disabled: terminating ",(string .proc.procname)," process."]];
+
```
SeanW17, line 12: Does this not need to be at the top of the file so nothing is defined prior in case of an uncaught error such as .ds.sementid not being specified?

PR #534, opened 2022-10-12
Change rdb proctype for different segments

dmitchard:
- Changed RDB proctypes within fsp
- Function in gateway to add processes specified in JSON to list of connections
- Add getaccess function to tailreader


AdownieAQ (approved):
Not sure why the change from flip to first was made, other than that looks good!

mattdeshAQ:
No longer needed - updated branch already merged

PR #536, opened 2022-10-14
TCS-165 Adding check to ensure datastriping is only enabled on STP

David-Mc-Nally:
Specificing the proctype as the Segmented Tickerplant only

### DIFF
--- a/code/processes/segmentedtickerplant.q
+++ b/code/processes/segmentedtickerplant.q
@@ -85,4 +85,4 @@ init:{
 
 // Have the init function and config check called from torq.q
 .proc.addinitlist(`init;`);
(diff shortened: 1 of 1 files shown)


THarveyWork (commented):
Looks Good

AdownieAQ (approved):
LGTM

ccrease01 (approved):
Looks good

jonathonmcmurray (commented):
This looks fine, but I think the PR title is wrong/misleading (this matters because it becomes the commit message when this gets merged, and it needs to be clear).

I think it's more accurate to say that we're adding a check to ensure that datastriping is only enabled on segmentedtickerplant

jonathonmcmurray (approved):

PR #538, opened 2022-11-03
TR to get access table from tailer

ShawnTay96:
[TCS-169](https://torqv5cloudscaling.atlassian.net/jira/software/projects/TCS/boards/1/backlog?selectedIssue=TCS-169)
Tailreader to obtain access table from the respective tailer's in-memory table
Currently, access table is being read from disk. Might be an issue if tailer is writing to disk and tailreader attempts to read from disk at the same time.


THarveyWork (commented):
Looks Good!

AdownieAQ (approved):
LGTM

code/processes/tailreader.q
```diff
@@ -6,6 +6,9 @@ getpartition:@[value;`getpartition;                                        /-fun
 currentpartition:@[value;`currentpartition;getpartition[]]
 basedir:raze (getenv`KDBTAIL),"/tailer",(string .ds.segmentid),"/"         /-define associated tailer base directory
 taildir:`$ basedir,string currentpartition;                                /-define tailDB direction
+tailertypes:`$"tailer_",last "_" vs string .proc.proctype                  /-define tailer to make connection to
```
jonathonmcmurray, line 9: this seems a bit fragile to me - IMO it might be better to have this config driven? as-is I think it ties users into a fairly specific naming convention for proctypes

also minor nitpick but if there's only going to be one type, variable should be called `tailertype` IMO

code/processes/tailreader.q
```diff
@@ -6,6 +6,9 @@ getpartition:@[value;`getpartition;                                        /-fun
 currentpartition:@[value;`currentpartition;getpartition[]]
 basedir:raze (getenv`KDBTAIL),"/tailer",(string .ds.segmentid),"/"         /-define associated tailer base directory
 taildir:`$ basedir,string currentpartition;                                /-define tailDB direction
+tailertypes:`$"tailer_",last "_" vs string .proc.proctype                  /-define tailer to make connection to 
+.servers.CONNECTIONS:(distinct .servers.CONNECTIONS,.tr.tailertypes) except `
```
jonathonmcmurray, line 10:
```suggestion
.servers.CONNECTIONS:(.servers.CONNECTIONS union .tr.tailertypes) except ` 
```

code/processes/tailreader.q
```diff
@@ -25,13 +28,9 @@ reload:{
   @[.Q.l ;.tr.taildir;{.lg.e[`load;"Failed to load intradayDB with error: ",x]}];
   .lg.o[`load;"intradayDB loaded"];
   .lg.o[`load;"loading accesstable"];
-  .ds.access:@[.Q.l;accesstabdir;{.lg.e[`load;"Failed to load tailer accesstable with error: ",x]}];
-  /- select last set of entries from accesstable
-  .ds.access:select by table from .ds.access;
+  /- make a connection to the tailer to get the in-memory access table
+  tailerhandle: first exec w from .servers.getservers[`proctype;.tr.tailertypes;()!();1b;0b];
+  .ds.access:tailerhandle".ds.access";
```
jonathonmcmurray, line 33: should probably error trap this & think about how we handle scenario where we can't retrieve access table for whatever reason

I think we should also do a check before this line to make sure we got a valid handle, and log an error if not

PR #540, opened 2022-12-05
Phase 1 of Torqestration

jamielamont:
Created new orchestrator process
Setup configurable lower and upper limits
Implemented scale up and scale down functions
Function to auto scale up upon startup to meet lower limit set in config 
Ability to call orchestrator to scale up and down from gateway

### DIFF
--- a/code/processes/gateway.q
+++ b/code/processes/gateway.q
@@ -535,6 +535,18 @@ addserversfromconnectiontable[.servers.CONNECTIONS]
 // Join active .gw.servers to .servers.SERVERS table
 activeservers:{lj[select from .gw.servers where active;`handle xcol `w xkey .servers.SERVERS]}
 
+/function to tell orchestrator to scale up
+scaleup:{[procname]
+handle:first exec w from .servers.SERVERS where proctype=`orchestrator;
+neg[handle](`.orch.scaleup;procname);
+}
+
+/function to tell orchestrator to scale down
+scaledown:{[procname]
+handle:first exec w from .servers.SERVERS where proctype=`orchestrator;
+neg[handle](`.orch.scaledown;procname);
+}
+
 \d .
 
 // functions called by end-of-day processes
--- a/code/processes/orchestrator.q
+++ b/code/processes/orchestrator.q
@@ -4,7 +4,7 @@
 
 /default parameters
 
-scalingdetails:([] time:`timestamp$(); procname:`$(); dir:`$(); totalnumofinstances:`int$(); lowerlimit:`int$(); upperlimit:`int$());	/table for tracking scaling
+scalingdetails:([] time:`timestamp$(); procname:`$(); dir:`$(); instancecreated:`$(); instanceremoved:`$(); totalnumofinstances:`int$(); lowerlimit:`int$(); upperlimit:`int$());	/table for tracking scaling
 
 processlimitscsv:hsym first .proc.getconfigfile"processlimits.csv";	/location of csv file
 limits:1!("SII";enlist ",")0:processlimitscsv;	/table of scalable processes and the max number of instances allowed for each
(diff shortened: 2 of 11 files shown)


code/processes/gateway.q
```diff
@@ -535,6 +535,18 @@ addserversfromconnectiontable[.servers.CONNECTIONS]
 // Join active .gw.servers to .servers.SERVERS table
 activeservers:{lj[select from .gw.servers where active;`handle xcol `w xkey .servers.SERVERS]}
 
+/function to tell orchestrator to scale up
```
drgdavies, line 538: The code here is duplicated across both.  Could you have a generic function to cover both scenarios and either use params or projections?

code/processes/orchestrator.q
```diff
@@ -0,0 +1,56 @@
+/TorQ Orchestrator Process
+
+\d .orch
+
+/default parameters
+
+scalingdetails:([] time:`timestamp$(); procname:`$(); dir:`$(); instancecreated:`$(); instanceremoved:`$(); totalnumofinstances:`int$(); lowerlimit:`int$(); upperlimit:`int$());	/table for tracking scaling
+
+processlimitscsv:hsym first .proc.getconfigfile"processlimits.csv";	/location of csv file
+limits:1!("SII";enlist ",")0:processlimitscsv;	/table of scalable processes and the max number of instances allowed for each
```
drgdavies, line 10: What is the key of the file?  1! doesn't give an indication, I think I'd prefer use of xkey here given we don't know the file.

2 other points - is there already csv loading functionality in TorQ? Could it be leveraged? For example what happens if file doesn't exist here? OR if the file load fails?  We need to handle those scenarios.

code/processes/orchestrator.q
```diff
@@ -0,0 +1,56 @@
...
+scaleprocslist:exec procname from limits;	/list of scalable processes
+
+/initialises connection to discovery process and creates keyed table containing the number of instances of each scalable process
+getscaleprocsinstances:{[] 
+	.servers.startup[];
+	`.orch.procs set string@/:exec procname from .servers.procstab;
```
drgdavies, line 16: Why are these stringed?

code/processes/orchestrator.q
```diff
@@ -0,0 +1,56 @@
...
+	`.orch.scaleprocsinstances set ([procname:scaleprocslist] instances:{sum procs like x,"*"}@/:string@/:scaleprocslist);
+	}
+
+getscaleprocsinstances[];
+
+/function to scale up a process
+scaleup:{[procname]
```
drgdavies, line 23: Again scaleup and scaledown here there looks to be a lot of duplication in the code.  Potentially parameterize or project.

code/processes/gateway.q
simondalzell-aquaq: We were still working on adjusting the gateway code with differences between scaling up and down and this made it simpler. It can be a generic function with these changes added, I'll give it a look.

code/processes/orchestrator.q
jamielamont: In the next line (Line 17) there is a lambda which gets the count of instances for each scalable process. This is done by checking which of the strings in the .orch.procs list are like the scalable process. E.g. {sum procs like "rdb1*"} will recognize all instances of rdb1 i.e. rdb1.2 rdb1.3 and so on

code/processes/orchestrator.q
jamielamont: Ok cool I will have a look for this functionality

code/processes/orchestrator.q
```diff
@@ -0,0 +1,56 @@
...
+
+/function to ensure all processes have been scaled up to meet lower limit
+initialscaling:{[procname]
+        if[scaleprocsinstances[procname;`instances]<limits[procname;`lower];
+                reqinstances:limits[procname;`lower]-scaleprocsinstances[procname;`instances]; 
+		do[reqinstances;scaleup[procname]];
+        ];
+        }
+
+initialscaling@/:scaleprocslist;
```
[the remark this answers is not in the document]
jamielamont: Calls the intialscaling function for each of the scalable processes to ensure they have been scaled up to their lower limits on startup

code/processes/orchestrator.q
jonathonmcmurray: what if there's lots of rdbs and we have an `rdb10`, would that be detected as a replicated instance of `rdb1`?

I also don't think you need the `@/:` here?

code/processes/orchestrator.q
```diff
@@ -0,0 +1,56 @@
...
+/function to scale up a process
+scaleup:{[procname]
+	$[scaleprocsinstances[procname;`instances]<limits[procname;`upper];
+		[system "bash ${TORQHOME}/addproc.sh ",string procname;
+		/update number of process instances
+		getscaleprocsinstances[];
+		/update table with record for scaling up
+		`.orch.scalingdetails upsert (.z.p;procname;`up;`$(last procs where procs like string[procname],"*");`;scaleprocsinstances[procname;`instances];limits[procname;`lower];limits[procname;`upper])];
+		.lg.o[`scale;"upper limit hit for ",string procname]
+	];
```
jonathonmcmurray, line 31: Multi-line `$` with blocks in them are confusing IMO. I'd reverse the condition & early exit in an `if` e.g.

```suggestion
	if[scaleprocsinstances[procname;`instances]>=limits[procname;`upper];
		.lg.o[`scale;"upper limit hit for ",string procname];
		:();
	];
	system "bash ${TORQHOME}/addproc.sh ",string procname;
	/update number of process instances
	getscaleprocsinstances[];
	/update table with record for scaling up
	`.orch.scalingdetails upsert (.z.p;procname;`up;`$(last procs where procs like string[procname],"*");`;scaleprocsinstances[procname;`instances];limits[procname;`lower];limits[procname;`upper]);
```

code/processes/gateway.q
```diff
@@ -535,6 +535,18 @@ addserversfromconnectiontable[.servers.CONNECTIONS]
...
+scaleup:{[procname]
+handle:first exec w from .servers.SERVERS where proctype=`orchestrator;
```
jonathonmcmurray, line 540: I think you could/should probably use `.servers.gethandlebytype` here

```suggestion
handle:.servers.gethandlebytype[`orchestrator;`any];
```

(this will mean that if for whatever reason connection to orchestrator has been lost, it will attempt to re-open it first)

code/processes/orchestrator.q
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+
+/function to ensure all processes have been scaled up to meet lower limit
+initialscaling:{[procname]
+        if[scaleprocsinstances[procname;`instances]<limits[procname;`lower];
+                reqinstances:limits[procname;`lower]-scaleprocsinstances[procname;`instances]; 
+		do[reqinstances;scaleup[procname]];
+        ];
+        }
```
jonathonmcmurray, line 54: there's some indenting with spaces here, rest of file appears to be using tabs - make it consistent please

code/processes/orchestrator.q
jamielamont: Yes currently it would be detected so will need fixed and yes it's not needed, I see now 'like' accepts string or symbol in right operand. Thanks

PR #541, opened 2023-01-10
editing checks on rack function of analyticslib.q

mattdeshAQ:
Bug fix for new TorQ Version (4.2.1).

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,8 @@ Then:
 Head to the address it gives you to check if your changes have worked. More information about using mkdocs can be found [here](http://www.mkdocs.org/)
 
 ## Release Notes
+- **4.2.1, Jan 2023**
+  * Bug fixes
 - **4.2.0, Dec 2021**
   * Added functionality to Data Access API
   * Added Data Striping
--- a/README.md
+++ b/README.md
@@ -39,8 +39,17 @@ Then:
 Head to the address it gives you to check if your changes have worked. More information about using mkdocs can be found [here](http://www.mkdocs.org/)
 
 ## Release Notes
-- **4.2.1, Jan 2023**
+- **4.3.0, Jan 2023**
   * Bug fixes
+  * Updated documentation
+  * Added functionality for programmatically starting additonal TorQ processes within q processes in the cloud
+  * Improved distribution of syms when using stripe function
+  * Improved WDB reload functionality
+  * Improved functionality of Data Access API & applied bug fixes
+  * Added configurable WDB merge functionality
+  * Fixes to subscriber functionality via .u.sub when utilising batch mode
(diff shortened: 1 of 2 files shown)


hamza060 (approved):

code/common/analyticslib.q
```diff
@@ -96,8 +96,6 @@ intervals:{[args]
 
 /rack function
 rack:{[args]
-    / Call general check args function
-    checkargs[args;`table`keycols`base`fullexpansion`timeseries`round;" sSB"];
```
jonathonmcmurray, line 100: PR description doesn't give much context - why is this being removed? are there any potential issues that could arise from not checking args?

README.md
```diff
@@ -39,6 +39,8 @@ Then:
 Head to the address it gives you to check if your changes have worked. More information about using mkdocs can be found [here](http://www.mkdocs.org/)
 
 ## Release Notes
+- **4.2.1, Jan 2023**
+  * Bug fixes
```
jonathonmcmurray, line 43: Is there nothing other than bug fixes added since last release?

jonnypress: @mattdeshAQ can you make this 4.3.0

mattdeshAQ: My bad, I didn't realise there were so many other commits which will be part of the new version.

code/common/analyticslib.q
hamza060: It seems the second arg to `checkarg[]` is setting the dict keys to all being required rather than the actual required keys ```` `table`keycols ```` from [here](https://aquaqanalytics.github.io/TorQ/analyticslib/#usage_3)
Looking at the other functions, they have specified required keys as well.

README.md
mattdeshAQ: This has been updated now

code/common/analyticslib.q
mattdeshAQ: Yes as Hamza has said above, the checkargs function is reporting the keys: 
\`table\`keycols\`base\`fullexpansion\`timeseries\`round 
as required, whereas this doesn't seem to be the intention from the [documentation](https://aquaqanalytics.github.io/TorQ/analyticslib/#usage_3) (should only be \`table and \`keycols required).  Removing this for now will allow use of the .al.rack function as intended.

jonathonmcmurray: so why not keep table & keycols in here instead of completely removing? seems like if we remove this check, it will try to execute with those required args missing and then return a less clear error message when it later fails

mattdeshAQ: That's been re-added with those changes now

jonathonmcmurray (approved):
LGTM

PR #420, opened 2022-02-24
Tcs 66 gateway routing queries depending on the symbols/time periods being requested

mauricelimaquaq:

PR #546, opened 2023-01-23
Gateway routing changes for striped setup using Data Access API

mattdeshAQ:
This PR includes code from PR #500 and #420, as well as further additions/fixes.
Changes to route queries by time/instruments for efficient querying using striped processes.  Time routing ensures no overlapping duplicate data is returned, particularly for intraday databases (Tailreader and RDB).  Instrument routing will ensure only relevant striped processes are queried. Routing can be viewed using the `getquery` option in the .dataaccess.getdata parameters e.g.: 
.dataaccess.getdata \`tablename\`starttime\`endtime\`instruments\`getquery!(\`trade;2023.01.22D20:00;2023.01.23D17:30;\`AAPL\`AIG\`GOOG;1b)

Link to FSP Branch PR: https://github.com/AquaQAnalytics/TorQ-Finance-Starter-Pack/pull/98

Note1: Changes were made to the dataaccess.md regarding new changes.  Some of this is now outdated and we will work on removing the outdated documentation.
Note2: Changes to accommodate for segmented rdb proctypes will be presented in a separate PR.


code/processes/tailreader.q
```diff
@@ -1,40 +1,71 @@
...
+  .tr.basedir:(raze/)1_string[.tr.taildir],"/tailer",string .tr.segmentid,"/";
+  .tr.wdbdir:`$ .tr.basedir,string .tr.currentpartition;
+  accesstabdir:`$ .tr.basedir,"access";
+  .lg.o[`load;"Loading intradayDB"];
+  @[.Q.l ;.tr.wdbdir;{.lg.e[`load;"Failed to load intradayDB with error: ",x]}];
+  .lg.o[`load;"intradayDB loaded"];
+  .lg.o[`load;"loading accesstable"];
+   /- make a connection to the tailer to get the in-memory access table
+  tailerhandle: first exec w from .servers.getservers[`proctype;.tr.tailertypes;()!();1b;0b];
+  .ds.access:tailerhandle".ds.access";
```
mattdeshAQ, line 51: N.B. Error handling for this has been implemented on a separate branch/PR (#538).  We will need to ensure the change is there on the final merge.

code/gateway/gatewaylib.q
```diff
@@ -92,3 +92,198 @@ getserverscross:{[req;att;besteffort]
...
+    // Only return appropriate dates
+    types:(key procdict)where value procdict;
+    // If the dates are out of scope of processes then error
+    if[0=count types;
+        '`$"gateway error - no info found for that table name and time range. Either table does not exist; attributes are incorect in .gw.servers on gateway, or the date range is outside the ones present"
+       ];
+    .lg.o[`.gw.attributesrouting;"Finished finding relevant processes"];
+    :types;
+    };
+
```
AdownieAQ, line 142: Tabs & spaces not consistent throughout file

code/gateway/gatewaylib.q
```diff
@@ -92,3 +92,198 @@ getserverscross:{[req;att;besteffort]
...
+missinginstruments:{[instruments]
+    $[1 < count instruments;
+        missing:(instruments)[where not any each {(key .ds.subreq) in x}each instruments];
+        missing:(enlist instruments)[where not any each (key .ds.subreq) in/: enlist instruments]
+    ];
+    :missing;
+    };
+
+adjustqueriesoverlap:{[options;part]
```
AdownieAQ, line 175: Tabs & spaces not consistent throughout file

code/processes/tailreader.q
```diff
@@ -1,40 +1,71 @@
...
+
+getaccess:{[] `location`table xkey update location:.proc.procname,proctype:.proc.proctype from .ds.access};
+
+// function to update the access table in the gateway. Takes the gateway handle as argument
+updategw:{[h]
+
+    newtab:getaccess[];
+    neg[h](`.ds.updateaccess;newtab);
+
+    };
```
AdownieAQ, line 29: Spaces & tabs not consistent

code/hdb/hdbstandard.q
```diff
@@ -1,7 +1,29 @@
...
+    / select table using max date if date col exists
+	/ get all cols that contains date (of type "pdz")
+	timecolumns:{tcols:exec c from meta t:$[`date in cols x;
+		select from x where date=max date;
+		value x]where t in"pdz";
+		/ date cols
+		dcols:exec c from meta t where t="d";
+		/ functional select to get the max value (defaults to -1+`timestamp$.z.d for endtimestamp)
+		dict:?[t;();();
+			tcols!enlist,/:-0Wp,/:
```
ccrease01, line 21: Tabbing/indentation here too

code/rdb/rdbstandard.q
```diff
@@ -1,5 +1,27 @@
...
+    .ds.tblstripe:@[stphandle;"select tbl,filts from .stpps.subrequestfiltered where handle = .z.w";
+        {.lg.e[`.proc.getattributes;"Failed to retrieve stripe map from STP"]}];
+    if[`tblstripe in tables[`.ds];
+        .ds.tblstripemapping:update stripenum:{last .ds.tblstripe[`filts][x;0;0]}each til count .ds.tblstripe from .ds.tblstripe;
+        ];
+    instrumentsfilter:1!select tablename:tbl,instrumentsfilter:stripenum from .ds.tblstripemapping;
+    inftc:instrumentsfilter uj timecolumns;
+    dataaccess:enlist[`dataaccess]!enlist`segid`tablename!(.ds.segmentid 0;(exec tablename from inftc)!value inftc);
+    default,:dataaccess;
+    default}
```
ccrease01, line 24: inconsistent indentation here

code/tailreader/trstandard.q
```diff
@@ -0,0 +1,18 @@
+// Get the relevant tailreader attributes
+.proc.getattributes:{default:`date`tables`procname!(.z.d;tables[] except .wdb.ignorelist;.proc.procname);
+    / get all cols that contains date (of type "pdz")
+    timecolumns:1!{tcols:exec c from meta value x where t in"pdz";
+        (enlist[`tablename]!enlist x),
+            /functional select to get the min value (defaults to `timestamp$.z.d for starttimestamp)
+            enlist[`timecolumns]!enlist?[x;();0b;                  //For 3rd arg: () produces dictionary with value of type list, first value of type 12h (desired) but does NOT work for tailreader. 0b runs without error but no dict created and type list on first value when flipped
+                tcols!(enlist,/:enlist each($;enlist`timestamp),/:enlist each((?),/:enlist each(=;0W),/:mtcols),'`.z.d,'mtcols:enlist each min,/:tcols),\:0Wp]}each tables[`.] except .wdb.ignorelist;
```
ccrease01, line 8: Indentation inconsistency

ccrease01 (commented):
Other than indentation pointed out it's good

AdownieAQ (approved):
LGTM

code/dataaccess/extractqueryparams.q
```diff
@@ -27,8 +27,8 @@ extractqueryparams:{[inputparams;queryparams]
 extracttablename:{[inputparams;queryparams]@[queryparams;`tablename;:;inputparams`tablename]};
 
 extractpartitionfilter:{[inputparams;queryparams]
-  //If an RDB return the partitionfilters as empty
-  if[`rdb~inputparams[`metainfo;`proctype];:@[queryparams;`partitionfilter;:;()]];
+  //If Intraday return the partitionfilters as empty
+  if[inputparams[`metainfo;`proctype] in `rdb`tr_seg1`tr_seg2;:@[queryparams;`partitionfilter;:;()]];
```
jonathonmcmurray, line 31: I don't think proctypes should be hardcoded here (I know `rdb` was already, but I think that was probably the wrong thing to do)

code/dataaccess/getdata.q
```diff
@@ -1,10 +1,16 @@
...
-
-getdata:{[inputparams]
-  if[.proc.proctype in key inputparams;inputparams:inputparams .proc.proctype];
+\d .dataaccess
+getprocdata:{[inputparams]
+  // if process is striped procname will be in the key
+  if[a:.proc.procname in key inputparams;inputparams:inputparams .proc.procname];
+  if[not[a]&.proc.proctype in key inputparams;inputparams:inputparams .proc.proctype];
+  // optimize hdb query from the gateway
+  if[`optimhdb in key inputparams;
```
jonathonmcmurray, line 8: is `optimhdb` generated elsewhere in code or are users exepcted to input this somewhere when using data access API? if the latter, it should have a more user-friendly name IMO

code/dataaccess/queryorder.q
```diff
@@ -52,7 +52,8 @@ extractkeys:{[queryparams;k]
 //- Put the partition filter top of the query
 orderpartedtable:{[queryparams;whereclause]
    // Errors out if there is no partition filter
-   if[queryparams[`partitionfilter]~();'"Include a partition filter"];
+   if[queryparams[`partitionfilter]~();
+        if[.proc.proctype=`hdb;'"Include a partition filter"]];
```
jonathonmcmurray, line 56:
```suggestion
   if[(queryparams[`partitionfilter]~())&.proc.proctype=`hdb;'"Include a partition filter"];
```

code/gateway/dataaccess.q
```diff
@@ -61,38 +61,53 @@ getdata:{[o]
...
-    default:`timeout`postback`sublist`getquery`queryoptimisation`postprocessing!(0Wn;();0W;0b;1b;{:x;});
+    default:`timeout`postback`sublist`getquery`optimisation`postprocessing!(0Wn;();0W;0b;1b;{:x;});
     // Use upserting logic to determine behaviour
     options:default,o;
-    if[`ordering in key o;options[`ordering]: go each options`ordering];
-    o:adjustqueries[o;partdict o];
+    k:key o;
+    if[`ordering in k;options[`ordering]: go each options`ordering];
+    // Instruments wildcard (`)
+    if[(`~o`instruments)&`instruments in key o;o _: `instruments];
```
jonathonmcmurray, line 74:
```suggestion
    if[(`~o`instruments)&`instruments in k;o _: `instruments];
```

`key o` is already in a separate variable

code/gateway/dataaccess.q
```diff
@@ -146,60 +146,26 @@ returntab:{[input;tab;reqno]
...
+	dict:.gw.adjustqueriesoverlap[options;part];
+	// adjust map reducable aggregations to get correct components
+	if[(1<count dict`dates)&`aggregations in key options;
+		if[all key[o:options`aggregations]in key aggadjust;
+			aggs:mapreduce[o;$[`grouping in key options;options`grouping;`]];
+			options:@[options;`aggregations;:;aggs]]];
+    // Modify queries based on striped processes
+    // Return query as a dict of table
+	:.gw.adjustqueriesstripe[options;dict];
+	};
```
jonathonmcmurray, line 166: inconsistent indenting (some spaces, some tabs)

code/gateway/gatewaylib.q
```diff
@@ -92,3 +92,199 @@ getserverscross:{[req;att;besteffort]
...
+    // Checktype
+    dict:.checkinputs.checkinputs dict;
+
+    // Get the Procs in a (nested) list of serverid(s)
+    dict[`procs]:attributesrouting[dict;part:partdict dict];
+
+    // Adjust queries based on relevant data and stripe
+    d:adjustqueriesoverlap[dict;part];
+    servers:adjustqueriesstripe[dict;d];
+	:key[servers]!(`serverid`attributes`checksperformed`optimhdb)_value servers;
```
jonathonmcmurray, line 123: tab

code/gateway/gatewaylib.q
```diff
@@ -92,3 +92,199 @@ getserverscross:{[req;att;besteffort]
...
+    };
+
+// Dynamic routing finds all processes with relevant data 
+attributesrouting:{[options;procdict]
```
jonathonmcmurray, line 127: this only does routing based on times, not the instrument striping stuff, is that right?

code/gateway/gatewaylib.q
```diff
@@ -92,3 +92,199 @@ getserverscross:{[req;att;besteffort]
...
+    procdict:serverids!(servers'[first each serverids]`attributes)@\:`date;
+    // Dictionary as min date/ max date
+    @[procdict;key procdict;{:(min x; max x)}]
+    };
+
+missinginstruments:{[instruments]
+    $[1 < count instruments;
+        missing:(instruments)[where not any each {(key .ds.subreq) in x}each instruments];
+        missing:(enlist instruments)[where not any each (key .ds.subreq) in/: enlist instruments]
+    ];
```
jonathonmcmurray, line 171:
```suggestion
    missing:((),instruments)[where not any each {(key .ds.subreq) in x}each (),instruments;
```

jonathonmcmurray (commented):
still working through this as it's chunky, a few initial comments - on the hardcoding proctypes stuff. it looks like there's a lot of this in existing data access API code; I think that's not great, but probably out of the scope of this change to fix that (this change is already massive)

code/gateway/gatewaylib.q
mattdeshAQ: that's right yes, the instruments routing comes later in adjustqueriesstripe (also in gatewaylib.q)

code/dataaccess/extractqueryparams.q
mattdeshAQ: I agree, I think for the most part the hardcoding of proctypes will be addressed with Shawn's RDB proctype change PR #547 as we had realised it was very impractical to change so many scripts for scalability.  Some other changes along the same vein with relation to the tailsort are being planned also.

code/hdb/hdbstandard.q
```diff
@@ -1,7 +1,29 @@
...
+				enlist each(?),/:(enlist each(=;-0W),/:mtcols),'(enlist(+;-1;($;ets;`.z.d))),/:enlist each($;ets:enlist`timestamp),/:mtcols:enlist each max,/:tcols];
+		dict[dcols]:`date$dict dcols;
+		enlist[`timecolumns]!enlist dict
+		}each t:tables[`.];
+	/ update date attribute for .gw.partdict and .gw.attributesrouting
+	default[`date]:asc default[`date]union first[d]+til 1+last deltas d:exec(min;max)@\:distinct`date$raze[value each timecolumns][;1]from timecolumns;
+	default,:enlist[`dataaccess]!enlist`segid`tablename!(.ds.segmentid 0;t!select instrumentsfilter:{""}'[i],timecolumns from timecolumns);
```
jonathonmcmurray, line 28:
```suggestion
	default[`dataaccess]:`segid`tablename!(.ds.segmentid 0;t!select instrumentsfilter:{""}'[i],timecolumns from timecolumns);
```

code/processes/gateway.q
```diff
@@ -355,6 +355,7 @@ getserversindependent:{[req;att;besteffort]
 
 // execute an asynchronous query
 asyncexecjpts:{[query;servertype;joinfunction;postback;timeout;sync]
+ o:query 1;
```
jonathonmcmurray, line 358: what is this? the variable name isn't descriptive and there's no explanation of why this is happening (it's not used until about 30 lines later)

queries can be sent in many formats to TorQ gw, is this always going to work?

code/processes/gateway.q
```diff
@@ -374,11 +375,20 @@ asyncexecjpts:{[query;servertype;joinfunction;postback;timeout;sync]
...
+       errcheck:@[getserverids;servertype;{.gw.errorprefix,x}];
+       if[10h=type errcheck; errstr:errcheck];
+       queryattributes:()!();
+      ];
+     if[all -6h = type each raze servertype;
+       // its a (nested) list of serverid(s) e.g. 1 2 3 4i/(1i;3i;5 6i)
+       // servertype requires all to be of same type
+       servertype:(),/:key o;
+       queryattributes:enlist[`servertype]!enlist distinct value[o]$[`servertype in cols value o;`servertype;`procs];
+       query[2;1]:@[value o;`procs]!value o;
```
jonathonmcmurray, line 390: mentioned above, but it seems like this is expecting a very specific query format - is there any guarantee that we will only reach this piece of code if the query is in that format?

also probably need some comments here, it's not very clear what this is doing or why

code/processes/tailer.q
```diff
@@ -56,3 +56,16 @@ initdatastripe[];
...
+
+/- reload the access table and tailDB in the tailreader process
+.tailer.dotailreload[`];
+
+/- define handle to stp and retrieve table stripe mapping
+stphandle:$[count i:.servers.getservers[`proctype;`segmentedtickerplant;()!();1b;0b];
+  first exec w from i;
+    .lg.e[`stphandle;"Failed to get a valid handle to the segmented tickerplant process"]];
+.ds.tblstripe:@[stphandle;"select tbl,filts from .stpps.subrequestfiltered where handle = .z.w";
+    {.lg.e[`.ds.tblstripe;"Failed to retrieve stripe map from STP"]}];
```
jonathonmcmurray, line 68: can this not be stored during subscription, rather than having to send another request to TP?

code/processes/tailreader.q
```diff
@@ -1,40 +1,71 @@
...
```
jonathonmcmurray, line 28:
```suggestion
// function to update the access table in the gateway. Takes the gateway handle as argument
updategw:{[h]neg[h](`.ds.updateaccess;getaccess[])};
```

code/processes/tailreader.q
```diff
@@ -1,40 +1,71 @@
...
   /- function to define the access table and IDB dir and then reload both tables
   /- reload is triggered by tailer after savedown occurs
-  basedir::(raze/)1_string[.tr.taildir],"/tailer",string .tr.segmentid,"/";
-  wdbdir::`$ basedir,string currentpartition;
-  accesstabdir::`$ basedir,"access";
-  
- @[.Q.l ;wdbdir;{.lg.e[`load;"Failed to load intradayDB with error: ",x]}];
- @[.Q.l ;accesstabdir;{.lg.e[`load;"Failed to load tailer accesstable with error: ",x]}];
-  load hsym `$basedir,"sym"
+  .tr.basedir:(raze/)1_string[.tr.taildir],"/tailer",string .tr.segmentid,"/";
```
jonathonmcmurray, line 42: This looks a bit messy. Often nice to build filepaths using `sv` https://code.kx.com/q/ref/sv/#filepath-components

(similar in next few lines)

code/processes/tailreader.q
jonathonmcmurray, line 51: this is overwritten in next line, should this be removed?

code/rdb/rdbstandard.q
```diff
@@ -1,5 +1,27 @@
 // Get the relevant RDB attributes
-.proc.getattributes:{`date`tables!(.rdb.rdbpartition[],();tables[])}
+.proc.getattributes:{default:`date`tables`procname!(.rdb.rdbpartition[];tables[] except .rdb.ignorelist;.proc.procname);
+    / get all cols that contains date (of type "pdz")
+    timecolumns:1!{tcols:exec c from meta value x where t in"pdz";
```
jonathonmcmurray, line 4:
```suggestion
    timecolumns:1!{tcols:exec c from meta x where t in"pdz";
```

don't need to value tablename

is the `timecolumns` structure actually needed here? seems like we reference `tcols` later, but I think for `timecolumns` we only really use `raze[value each timecolumns][;0]` - it might be simpler if we didn't build this structure here & just had separate variables for `tcols` and whatever the other bit of `timecolumns` is that we use later on? (just noticed it is joined to something later, so maybe it is needed - it still may be simpler to use some intermediate variables here, get the date range from those etc., then construct the final structure from those later in function)

code/rdb/rdbstandard.q
```diff
@@ -1,5 +1,27 @@
...
+        (enlist[`tablename]!enlist x),
+            /functional select to get the min value (defaults to `timestamp$.z.d for starttimestamp)
+            enlist[`timecolumns]!enlist?[x;();();
+                tcols!(enlist,/:enlist each($;enlist`timestamp),/:enlist each((?),/:enlist each(=;0W),/:mtcols),'`.z.d,'mtcols:enlist each min,/:tcols),\:0Wp]}each tables[`.] except .rdb.ignorelist;
+    / update date attribute for .gw.partdict and .gw.attributesrouting
+    default[`date]:asc default[`date]union first[d]+til 1+last deltas d:exec(min;max)@\:distinct`date$raze[value each timecolumns][;0]from timecolumns;
```
jonathonmcmurray, line 10:
```suggestion
    default[`date]:asc default[`date]union first[d]+til 1+(-). d:exec(max;min)@\:distinct`date$raze[value each timecolumns][;0]from timecolumns;
```

code/rdb/rdbstandard.q
```diff
@@ -1,5 +1,27 @@
...
+        / for striped rdbs, retrieve stripe mapping from stp and deduce instruments in each table to report as attributes to the gateway 
+    stphandle:$[count i:.servers.getservers[`proctype;`segmentedtickerplant;()!();1b;0b];
+        first exec w from i;
+            .lg.e[`.proc.getattributes[];"Failed to get a valid handle to the segmented tickerplant process"]];
+    .ds.tblstripe:@[stphandle;"select tbl,filts from .stpps.subrequestfiltered where handle = .z.w";
+        {.lg.e[`.proc.getattributes;"Failed to retrieve stripe map from STP"]}];
```
jonathonmcmurray, line 16: do we need to do this every time we update attributes? Can we not store this when we subscribe & just reference it here? (I don't think this info will change at any time other than subscription)

code/wdb/wdbstandard.q
```diff
@@ -0,0 +1,22 @@
...
+.proc.getattributes:{default:`date`tables`procname!(.wdb.currentpartition;tables[];.proc.procname);
+    / get all cols that contains date (of type "pdz")
+    timecolumns:1!{tcols:exec c from meta value x where t in"pdz";
+        (enlist[`tablename]!enlist x),
+            /functional select to get the min value (defaults to `timestamp$.z.d for starttimestamp)
+            enlist[`timecolumns]!enlist?[x;();();
+                tcols!(enlist,/:enlist each($;enlist`timestamp),/:enlist each((?),/:enlist each(=;0W),/:mtcols),'`.z.d,'mtcols:enlist each min,/:tcols),\:0Wp]}each tables[`.];
+    / update date attribute for .gw.partdict and .gw.attributesrouting
+    default[`date]:asc default[`date]union first[d]+til 1+last deltas d:exec(min;max)@\:distinct`date$raze[value each timecolumns][;0]from timecolumns;
+    if[.ds.datastripe;
```
jonathonmcmurray, line 11: I thought wdb wouldn't be running if data striping was turned on? Has that changed?

jonathonmcmurray (commented):
Has been mentioned in various places, but a lot of inconsistency in indentation. Most editors should have a setting to show whitespace chars (allowing you to distinguish between tabs & spaces) - to display them on GitHub PRs etc. which can be handy, you can use a browser extension like https://github.com/refined-github/refined-github

code/wdb/wdbstandard.q
mattdeshAQ: I think this is redundant code from before when the tailer and wdb were one in the same so it can be safely removed.

code/gateway/gatewaylib.q
```diff
@@ -92,3 +92,199 @@ getserverscross:{[req;att;besteffort]
...
+    segids:exec distinct[segid]except ` from querytable;
+    if[all[i,count segids];
+        if[0 < count missinginstruments[options`instruments];
+            // if a queried instrument is not in stripe mapping, reload stripe mapping to bring up-to-date mapping
+            .ds.getstripemapping[];
+            // create messaging if instrument is not in mapping
+            if[[missing:missinginstruments[options`instruments]; 0 < count missing];
+                .lg.o[`.gw.adjustqueriesstripe;raze("No intra-day data for the following instruments: ",(string missing))];
+                    // exit script for intra-day only queries
+                if[[(count options`instruments) = count missing;not `hdb in exec servertype from querytable];
```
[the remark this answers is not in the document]
mattdeshAQ: Yes you're right, it will execute dependant on the second cond which was not the intention.  This must've been one I had missed fixing

code/dataaccess/getdata.q
mattdeshAQ: It's a user input for the Data Access API - I believe its used for optimising hdb queries in which the data to query is across an entire day or more as the code will covert the starttime/endtime arguments to query into dates. We will look at this and add it into the documentation (and also rename to something like `optimisehdb`)

jonathonmcmurray: if it's existing in data access API prior to TorQ cloud, don't worry about changing it, but yeah if it's added, please give it a nicer name

code/gateway/gatewaylib.q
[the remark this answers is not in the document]
mattdeshAQ: Yes that's right - it would only execute dependant on the second condition which was not the intention.  I will implement an ampersand to ensure both return true to pass this conditional.

code/gateway/gatewaylib.q
```diff
@@ -92,3 +92,199 @@ getserverscross:{[req;att;besteffort]
...
+        ];
+    ];
+    // filter overlap timings between rdb and tailreader process
+    if[count segids;
+        querytable:{y;if[all exec any each procs like/:("*tr*";"rdb*") from x where segid=y;
+            rdbtimes:exec(starttime,endtime)from x where(segid=y)&servertype like"rdb*";
+            trtimes:exec(starttime,endtime)from x where(segid=y)&servertype like"tr*";
+            if[rdbtimes[0]<trtimes 1;trtimes[1]:rdbtimes[0]-1;
+                :update starttime:trtimes 0,endtime:trtimes 1 from x where(segid=y)&servertype like"tr*"]];x}/[querytable;segids];
+        querytable:delete from querytable where starttime>endtime;
```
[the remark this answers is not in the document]
mattdeshAQ: This isn't actually deleting bad user queries but rather handling overlaps in the processes to be queried. I believe the idea of it was to remove tailreader queries where they overlap rdb data and the query is able to only query in memory (rdb) data to return a the user's full query

code/gateway/gatewaylib.q
```diff
@@ -92,3 +92,199 @@ getserverscross:{[req;att;besteffort]
...
+
+        // routing instruments for striped databases
+        if[all[i,any exec any each procs like/:("tr*";"rdb*") from querytable];
+            hdbquery:select from querytable where servertype = `hdb;
+            stripedquery:select from querytable except hdbquery;
+            stripedquery:update 
+                instruments:?[max max each ((group .ds.subreq)first each inftc) in' instruments;
+                    ((group .ds.subreq)first each inftc) inter' instruments;0N] 
+                            from stripedquery;
+            querytable:stripedquery uj hdbquery];
```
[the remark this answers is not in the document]
mattdeshAQ: I've been attempting to implement logic similar to suggested here but I can't find a way around a 'type error pointing to the update statement which is only triggering when querying one instrument.  Querying multiple instruments appears to work. Anyone know why this may be the case?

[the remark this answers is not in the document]
jonathonmcmurray: So it's

```suggestion
            querytable :update 
                instruments:?[max max each ((group .ds.subreq)first each inftc) in' instruments;
                    ((group .ds.subreq)first each inftc) inter' instruments;0N] 
                            from querytable where servertype<>`hdb];
```

right?

and the existing code works for single instrument?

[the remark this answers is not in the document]
mattdeshAQ: Yep that's the solution I imagined would work too but it is unfortunately not.  Existing code does indeed work for single instrument queries, yes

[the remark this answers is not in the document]
mattdeshAQ: Behaviour is the same whether the query is intra-day only or historical and intra-day too I should note (so the same behaviour for both when `hdb is and isn't in the querytable)

code/gateway/gatewaylib.q
```diff
@@ -92,3 +92,199 @@ getserverscross:{[req;att;besteffort]
...
+        ];
+
+        // fixing length error for single sym queries
+        if[i;
+            querytable:update instruments:{$[1=count x;first x;x]}'[instruments] from querytable
+        ];
+    ];
+    // filter overlap timings between rdb and tailreader process
+    if[count segids;
+        querytable:{y;if[all exec any each procs like/:("*tr*";"rdb*") from x where segid=y;
```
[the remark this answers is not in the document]
mattdeshAQ: Removing this `y` causes a 'type error here which points to the `/` a few lines below (which is used for interacting over the lambdas in which the subject `y` is a part of)

code/gateway/gatewaylib.q
```diff
@@ -92,3 +92,199 @@ getserverscross:{[req;att;besteffort]
...
+		dates:key[dates]!?[value[dates][;0]<start;start;value[dates][;0]],'?[value[dates][;1]>end;end:options`endtime;value[dates][;1]];
+		];
+    .lg.o[`.gw.adjustqueriesoverlap;"adjustqueriesoverlap complete"];
+    :`part`isdate`dates!(part;a;dates);
+    }
+
+// Modify queries based on striped processes
+adjustqueriesstripe:{[options;dict]
+    // create a dictionary of procs and different queries
+	query:{@[@[x;`starttime;:;y 0];`endtime;:;y 1]}[options]'[dict`dates];
```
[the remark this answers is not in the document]
mattdeshAQ: I'd like to find a way to make this more readable, however the suggested change creates a 'type error pointing to the update statement when defining `querytable` a few lines down.

code/gateway/gatewaylib.q
```diff
@@ -92,3 +92,199 @@ getserverscross:{[req;att;besteffort]
...
+            stripedquery:update 
+                instruments:?[max max each ((group .ds.subreq)first each inftc) in' instruments;
+                    ((group .ds.subreq)first each inftc) inter' instruments;0N] 
+                            from stripedquery;
+            querytable:stripedquery uj hdbquery];
+
+        // remove queries for striped procs with no instuments to query
+        if[i;
+            querytable:delete from querytable where 0 = count each instruments;
+            if[[i;max any each null exec instruments from querytable];
```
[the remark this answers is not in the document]
mattdeshAQ: I think we may be able to remove null consideration altogether here as recent changes have prevented nulls from appearing in the querytable, and now we only have empty lists.  To address your question, I found that attempting to delete from a table which does not contain any nulls will throw a 'type error which was the reason for the conditional.  The suggestion using the logic or `|` will throw the same error.

code/gateway/gatewaylib.q
```diff
@@ -92,3 +92,199 @@ getserverscross:{[req;att;besteffort]
...
+    querytable:update serverid:{x where{any x in y}[;y]each x}[options`procs;serverid],
+        // get servertype
+        servertype:`${string .gw.servers'[x]`servertype}serverid,
+        // convert procs into procname if striped
+        procs:.gw.servers[;`attributes;`procname]@/:serverid from querytable;
+    querytable:update procs:servertype from querytable where not(last each serverid)in modquery`serverid;
+
+    // check if instruments are contained in stripe mapping
+    segids:exec distinct[segid]except ` from querytable;
+    if[all[i,count segids];
```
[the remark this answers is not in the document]
mattdeshAQ: Going to use an ampersand here as we want the conditional to run when both conditions are true.
```suggestion
    if[instrqueried&count segids;
```

code/gateway/gatewaylib.q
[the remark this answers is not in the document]
mattdeshAQ: We'll be looking at this section to change from using tr/rdb naming conventions also

code/processes/gateway.q
mattdeshAQ: o will extract a dictionary of query parameters for the DA API queries.  We'll change the name to be more descriptive and add some comments.

code/processes/gateway.q
mattdeshAQ: This code is specifically used for the DA API only. We'll add comments for this and change the conditional to bring it more in line with the function prior to the DA API changes. This will also make it more robust for the code to only be run when the DA API is used and resort to TorQ v4 logic otherwise.

code/processes/gateway.q
mattdeshAQ: Queries will work for all formats as desired - DA API will use newer logic but everything else should use previous TorQ version logic.

code/processes/tailreader.q
mattdeshAQ: This looks to be old code which hasn't been updated on this branch.  I believe the plan will be to use the format shown in the PR #538 for the final merge.

jonathonmcmurray: hmm, it shows in the diff, are you sure it hasn't been updated on this branch?

code/gateway/gatewaylib.q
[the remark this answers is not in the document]
mattdeshAQ: Solved by creating list for single instrument queries prior to running `adjustqueriesstripe`

code/dataaccess/getdata.q
mattdeshAQ: Found it's actually generated in the code, not an input as previously thought.  Renaming it to make the purpose of it more clear regardless.

code/gateway/gatewaylib.q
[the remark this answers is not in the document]
mattdeshAQ: This section of code is going to be replaced with new logic which will utilise functionality from the new rdb proctype changes (PR #547 ).  The plan is to update this after merging as implementing the code on this branch at this time would not work as intended.

code/processes/tailreader.q
mattdeshAQ: Sorry, I meant it hasn't been updated on this branch since an early stage - we had been working on this code elsewhere. I'm going to update this script to a final version here so that we can be sure to have the latest work all in one place.

jonathonmcmurray (commented):
looks good, couple of little things still

code/processes/wdb.q
```diff
@@ -430,7 +430,14 @@ subscribe:{[]
 		/- return the tables subscribed to and the tickerplant log date
 		subto:.sub.subscribe[subtabs;subsyms;schema;replay;subproc];
 		/- check the tp logdate against the current date and correct if necessary 
-		fixpartition[subto];];}
+		fixpartition[subto];];
+		/-retrieving table stripe mapping
+		stphandle:first exec w from s;
+		.ds.tblstripe:@[stphandle;"select tbl,filts from .stpps.subrequestfiltered where handle = .z.w";
```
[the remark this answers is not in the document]
mattdeshAQ: oh i missed that, will have a look into it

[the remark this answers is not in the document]
mattdeshAQ: From what I can tell when it comes to the tailer/rdb, `.stpps.subrequestfiltered` gets defined by the pubsub.q script rather than being passed with `subto`.  It appears empty in the tailer/rdb after being defined too, since the table tracks the mapping of the subscribers' tables, so it doesn't contain the info that would be needed.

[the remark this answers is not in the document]
jonathonmcmurray: Sure - my point is does `subto` contain the information we need here? From what I remember, we made it so that when a process does a striped subscription, STP will return the filter that is being used as part of the returned object (i.e. `subto` here), so that the subscriber knows what filter to apply when replaying TP logs etc. So it seems like we should already have that information without having to re-request it manually etc.

It's possible we changed that & `subto` doesn't contain the filter, in which can ignore this.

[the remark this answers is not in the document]
mattdeshAQ: It doesn't appear to contain a filter anymore.  `subtabs` and `subsyms` contained in `subto` are set to all, so the datastriping filter must be applied elsewhere. The other three args are set to a true boolean - `schema`,`replay` and `subproc`. Their functionality doesn't indicate a filter either but rather just to retrieve schema from the TP and replay the TP log file.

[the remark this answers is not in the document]
jonathonmcmurray: OK fair enough, guess this is fine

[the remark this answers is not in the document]
mattdeshAQ: I can't see any striping info in the returned object too

PR #556, opened 2023-02-21
hdb attribute retrieval bug fix/adding ignorelist to hdb

mattdeshAQ:
A bug was found preventing the `.proc.getattributes` function from running in the hdb as intended, believed due to the absence of the `packets` table on disk.  To remedy this we've added an `ignorelist` variable defined in the config/settings/hdb.q file to create a list of tables which we don't want to retrieve attributes from.  This list is config driven and therefore can be altered by users to suit their requirement.


ccrease01 (approved):
All good to me

jonathonmcmurray (approved):

PR #560, opened 2023-03-14
Fixes to Data Quality Processes failing postback for checks when TorQ Permissions enabled

mattdeshAQ:
A TorQ user reported an error occurring with the Data Quality Checker (dqc) when permissions logic was enabled.  This error prevented postback of results for checks made by dqc.

The root cause of this error was found to be due to a condition checking the type on the returning data and executing `eval` on it which errors.  A change to the condition was made to execute `value` instead which allows the postback to complete successfully.  This change adjusts the type of the data checked to be above type 104 (projection), which in the tested cases returns the false result of the conditional and in turn a successful postback.

Additionally, the Data Quality processes have been added to the list of permissioned processes.


jonnypress:
@mattdeshAQ there are some pending comments on this - can you check and amend?

code/handlers/permissions.q
```diff
@@ -153,7 +153,7 @@ lamq:{[u;e;b;pr]
   if[count prohibited;'" | " sv .pm.err[`selt] each prohibited];
   $[b; :exe e; :1b]}
 
-exe:{v:$[(100<abs type first x);val;valp]x;
+exe:{v:$[(104=abs type first x);valp;(100<abs type first x);val;valp]x;
```
jonathonmcmurray, line 156: I think we can simplify here something like

```suggestion
exe:{v:$[(104<>a)&100<a:abs type first x;val;valp]x;
```

jonathonmcmurray (commented):
Can we add some test cases to https://github.com/DataIntellectTech/TorQ/blob/master/tests/permissions/permissions.csv to cover the changed functionality please?

tests/permissions/permissions.csv
```diff
@@ -53,3 +53,7 @@ comment,,,,,,,test expressions where first item is lambda from .q
 true,,,q,(`a`b`c!1 4 7)~.pm.expr[`tom;"first each d"],,,test expression with each
 fail,,,q,.pm.expr[`harry;"first each d"],,,harry doesn't have access to d
 
+comment,,,,,,,test projection functionality works with both value and eval
+true,,,q,.pm.expr[`tom;({[x;y]y~`ABCD};`x;`ABCD)],,,passed to value and variable used literally
+true,,,q,.pm.expr[`tom;({[x;y]y~`ABCD}[`x];`ABCD)],,,passed to eval and looks up variable
```
jonathonmcmurray, line 58: I think these comments are a little misleading - in particular the second one describes the _old_ (incorrect) behaviour that has been fixed. Projections are no longer going to be passed to eval, so the symbol will be used literally, the same as in the first. In both cases, it's a symbol rather than a variable.

I think something like

```suggestion
true,,,q,.pm.expr[`tom;({[x;y]y~`ABCD};`x;`ABCD)],,,passing a symbol argument to a lambda function
true,,,q,.pm.expr[`tom;({[x;y]y~`ABCD}[`x];`ABCD)],,,passing a symbol argument to a projection
```

jonathonmcmurray (commented):
Looks good to me, just need to update the comments on the unit tests

jonathonmcmurray (approved):
LGTM

Issue #563, opened 2023-03-28
Merge by part breaks with many small partitions


jonathonmcmurray:
In #448, a change was made so that in the default "merge by part" mode for WDB, intraday partitions will be grouped into "chunks" based on row size, and an entire chunk will be read into memory at once.

This can cause an issue in the case where there are many small intraday positions (e.g. a large number of symbols with only a few records each). The total row count is still below the row limit configured, and so they are put together in one chunk. With a sufficient number of intraday partitions being read at the same time, q can encounter a `Cannot allocate memory` error. (Specific error may vary depending on OS).

I think we probably need to add an additional limit for the number of partitions that will be included in a given chunk i.e. a change in the `.merge.getpartchunks` function in `code/common/merge.q`.

PR #564, opened 2023-03-31
Update to .merge.getpartchunks to limit number partitions in a chunk #563

JamesMassey97:
Issue #563 
Additional limit has been added to .merge.getpartchunks function to limit the number of partitions that can be merged in a given chunk. Previously many small intraday positions could be merged within the row limit and reading in too many partitions at the same time can lead to a cannot allocate memory error in q.

### DIFF
--- a/code/common/merge.q
+++ b/code/common/merge.q
@@ -44,7 +44,7 @@ getpartchunks:{[partdirs;mergelimit]
   /-return list of partitions to be called in batch
   l:(where r={$[z<x+y;y;x+y]}\[0;r;mergelimit]),(select count i from .merge.partsizes)[`x];
   /-where there are more than set partlimit, split the list
-  s:-1_ asc raze {$[(x[y]-x[y-1])<=.merge.partlimit;x[y];x[y], x[y-1] + (v where 0=(v:1+ til x[y] - x[y-1]) mod .merge.partlimit)]}/:[l;til count l];
+  s:-1_distinct asc raze {$[(x[y]-x[y-1])<=.merge.partlimit;x[y];x[y], first each .merge.partlimit cut x[y-1]+til x[y] - x[y-1]]}/:[l;til count l];
   /-return list of partitions
   s cut exec ptdir from t
   };
--- a/code/common/merge.q
+++ b/code/common/merge.q
@@ -1,6 +1,6 @@
 \d .merge
 mergebybytelimit:@[value;`.merge.mergebybytelimit;0b];                        /- merge limit configuration - default is 0b row count limit 1b is byte size limit
-partlimit:@[value;`.merge.partlimit;5]                                        /- limit the number of partitions in a chunk
+partlimit:@[value;`.merge.partlimit;1000]                                     /- limit the number of partitions in a chunk
 
 partsizes:([ptdir:`symbol$()] rowcount:`long$(); bytes:`long$());             /- partsizes table used to keep track of table row count and bytesize estimate when data is written to disk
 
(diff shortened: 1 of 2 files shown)


jonnypress:
This looks ok to me but @jonathonmcmurray can you provide a review please?

code/common/merge.q
```diff
@@ -41,7 +42,11 @@ getpartchunks:{[partdirs;mergelimit]
   /-get list of limits (rowcount or bytesize) to be used to get chunks of partitions to get merged in batch
   r:$[.merge.mergebybytelimit;exec bytes from t;exec rowcount from t];
   /-return list of partitions to be called in batch by merge by part function
-  (where r={$[z<x+y;y;x+y]}\[0;r;mergelimit]) cut exec ptdir from t
+  l:(where r={$[z<x+y;y;x+y]}\[0;r;mergelimit]),(select count i from .merge.partsizes)[`x];
+  /-set a limit to the number of partitions that can be in a chunk
+  s:-1_ asc raze {$[(l[x]-l[x-1])<=.merge.partlimit;l[x];l[x], l[x-1] + (v where 0=(v:1+ til l[x] - l[x-1]) mod .merge.partlimit)]} each til count l;
```
jonathonmcmurray, line 47: does this work? it looks like `l` is referenced inside a lambda without being passed in, which I don't think will work?

code/common/merge.q
```diff
@@ -40,8 +41,12 @@ getpartchunks:{[partdirs;mergelimit]
   t:select from .merge.partsizes where ptdir in partdirs;
   /-get list of limits (rowcount or bytesize) to be used to get chunks of partitions to get merged in batch
   r:$[.merge.mergebybytelimit;exec bytes from t;exec rowcount from t];
-  /-return list of partitions to be called in batch by merge by part function
-  (where r={$[z<x+y;y;x+y]}\[0;r;mergelimit]) cut exec ptdir from t
+  /-return list of partitions to be called in batch
+  l:(where r={$[z<x+y;y;x+y]}\[0;r;mergelimit]),(select count i from .merge.partsizes)[`x];
+  /-where there are more than set partlimit, split the list
+  s:-1_ asc raze {$[(x[y]-x[y-1])<=.merge.partlimit;x[y];x[y], x[y-1] + (v where 0=(v:1+ til x[y] - x[y-1]) mod .merge.partlimit)]}/:[l;til count l];
```
jonathonmcmurray, line 47: I think this can be simplified a bit e.g. something like

```suggestion
  s:-1_ raze {$[(x[y]-x[y-1])<=.merge.partlimit;x[y];.merge.partlimit cut x[y-1]+til x[y] - x[y-1]]}/:[l;til count l];
```

don't think the `asc` is needed with this version as it'll still be in order anyway

JamesMassey97: Yes the new version doesn't need sorted. The output isn't exactly the same format, it gives a matrix rather than list, so I will make changes to it but it should still be a simplified version of the initial commit

JamesMassey97:
```suggestion
  s:-1_distinct asc raze {$[(x[y]-x[y-1])<=.merge.partlimit;x[y];x[y], raze 1#'.merge.partlimit cut x[y-1]+til x[y] - x[y-1]]}/:[l;til count l];
```
This is a simplified version based on the above suggestion. After the cut and indexing into the matrix I still needed to add the last value of l, which then needs to be sorted with asc and in this case remove duplicates. The extra steps are to stop any of the chunks made being an empty list.

jonathonmcmurray: instead of `raze 1#'` I think `first each` would do the same and probably clearer

```suggestion
  s:-1_distinct asc raze {$[(x[y]-x[y-1])<=.merge.partlimit;x[y];x[y], first each .merge.partlimit cut x[y-1]+til x[y] - x[y-1]]}/:[l;til count l];
```

JamesMassey97: Yes that works the same, agree that is a clearer way to do it

jonathonmcmurray (approved):
LGTM

PR #568, opened 2023-05-16
Adding extra param for informsortandreload within endofday function

### LINKED ISSUE #548: WDB - endofdaysort not running
In https://github.com/AquaQAnalytics/TorQ/blob/master/code/processes/wdb.q
Line 205

$[sortenabled;endofdaysort;informsortandreload] . (savedir;pt;tablist;writedownmode;mergelimits;hdbsettings)];

When sortenabled is true the endofdaysort function is passed 6 arguments, but further below on line 391
it is defined with 7 arguments.
The result is that the endofdaysort does not run.

When I added in the missing mergemode it worked.
$[sortenabled;
  endofdaysort[savedir;pt;tablist;writedownmode;mergelimits;hdbsettings;mergemode];
  informsortandreload[savedir;pt;tablist;writedownmode;mergelimits;hdbsettings]
]

Ciaran-Donnelly:
An extra argument (mergemethod) was added to the the informsortandreload function to allow it have the same amount of arguments as the endofdaysort function.  This will allow either to be called successfully depending on the variable sortenabled.
Also this change involved defining the mergemethod variable inside the endofday function instead of the informsortandreload function.

This closes #548

### DIFF
--- a/code/processes/wdb.q
+++ b/code/processes/wdb.q
@@ -193,7 +193,9 @@ savetodisk:{[] savetables[savedir;getpartition[];0b;] each tablelist[]};
 
 /- eod - flush remaining data to disk
 endofday:{[pt;processdata]
-	.lg.o[`eod;"end of day message received - ",spt:string pt];	
+	.lg.o[`eod;"end of day message received - ",spt:string pt];
+        /- set what type of merge method to be used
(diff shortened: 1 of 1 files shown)


jonathonmcmurray (approved):
LGTM

PR #576, opened 2023-06-14
Fix DQE savedown with alternative hdb types

jonathonmcmurray:
Currently, DQE assumes that all hdb processes will have proctype of `hdb. This causes a break in savedown when alternative proctypes are being used.

This change makes a config variable for the hdb types, so that this can be modified via appconfig for specific applications.


jonnypress (approved):
LGTM

PR #577, opened 2023-06-20
Sync reload

CReid98:
Synchronise reloads in a system where there are multiple subscribers and one hdb, such that the hdb will not reload until all subscribers have completed their reload process.


code/processes/gateway.q
```diff
@@ -637,4 +643,15 @@ neg[h](`.gw.asyncexecjpt;(`.q.system;"sleep 10");enlist[`servertype]!enlist`rdb`
...
+// dictionary of handles to reload
+reloadcalls:()!();
+
+// function to add handle to reloadcalls dictionary
+po:{[h] if[.z.u in `wdb;reloadcalls[h]:0b]};
+.z.po:{[f;x] @[f;x;()];.gwreload.po x} @[value;`.z.po;{{}}];
 
+// function to remove handle from reloadcalls dictionary
+pc:{[h] reloadcalls _: h; if[(all .gwreload.reloadcalls) & count .gwreload.reloadcalls;reload[]]};
+.z.pc:{[f;x] @[f;x;()];.gwreload.pc x} @[value;`.z.pc;{{}}];
```
jonathonmcmurray, line 657: I think this section won't be executed/do anything due to the `/` on line 597 - did this work when it was tested?

code/processes/rdb.q
```diff
@@ -114,6 +114,10 @@ endofday:{[date;processdata]
 	};
 	
 reload:{[date]
+	if[.z.w in key .rdb.reloadcalls;
+	        .rdb.reloadcalls[.z.w]:1b;
```
jonathonmcmurray, line 118: mix of tabs & spaces on this line, surrounding lines all use just tabs

code/hdb/hdbstandard.q
```diff
@@ -1,7 +1,26 @@
 // reload function
 reload:{
+
```
jonathonmcmurray, line 3:
```suggestion
```
remove blank line

code/hdb/hdbstandard.q
```diff
@@ -1,7 +1,26 @@
 // reload function
 reload:{
+	
+	if[.z.w in key .hdb.reloadcalls;
+		.hdb.reloadcalls[.z.w]:1b;
+		.lg.o[`reload;"reload call received from handle ", string[.z.w], "; reload calls pending from handles ", ", "sv string where not .hdb.reloadcalls];
```
jonathonmcmurray, line 6: this log line is possibly confusing when there are no pending reload calls? maybe that last part should be conditional on there being pending calls?

code/hdb/hdbstandard.q
```diff
@@ -1,7 +1,26 @@
...
+		if[not all .hdb.reloadcalls;:(::)]];
```
jonathonmcmurray, line 7:
```suggestion
		if[not all .hdb.reloadcalls;:(::)];
	];
```

IMO the closing `]` of a multi-line `if` block should be on it's own line and aligned with the opening `if`

code/hdb/hdbstandard.q
```diff
@@ -1,7 +1,26 @@
...
 // Get the relevant HDB attributes
 .proc.getattributes:{`date`tables!(@[value;`date;`date$()];tables[])}
+
+\d .hdb
+
+// dictionary of handles to reload
+reloadcalls:()!();
+
+// function to add handle to reloadcalls dictionary
+po:{[h] if[.z.u in `wdb`rdb;reloadcalls[h]:0b]};
```
jonathonmcmurray, line 21: I think hardcoding the usernames here is probably the wrong way to go - ideally I think we should probably do this based on proctype, and we should have a config variable for which proctypes (or usernames, if we can't use proctypes for some reason) we need to keep track of (users of TorQ can change proctypes, and the internal usernames used by them etc.)

Reason for using proctypes rather than usernames is that I don't think elsewhere in TorQ usernames have any specific "meaning" and typically any process identification logic is based off proctype - if we do change that here, we have to make sure we document that very clearly & signpost users to the config variable they have to update if they're using different usernames.

code/processes/gateway.q
```diff
@@ -550,9 +550,15 @@ reloadstart:{
  };
 
 reloadend:{
+ if[.z.w in key .gw.reloadcalls;
+  .gw.reloadcalls[.z.w]:1b;
+  .lg.o[`reload;"reload call received from handle ", string[.z.w], "; reload calls pending from handles ", ", "sv string where not .gw.reloadcalls];
+  if[not all .gw.reloadcalls;:(::)]];
+
```
jonathonmcmurray, line 557: same comments apply here as in the hdb script

code/processes/rdb.q
```diff
@@ -114,6 +114,10 @@ endofday:{[date;processdata]
...
+		.lg.o[`reload;"reload call received from handle ", string[.z.w], "; reload calls pending from handles ", ", "sv string where not .rdb.reloadcalls];
+		if[not all .rdb.reloadcalls;:(::)]];
```
jonathonmcmurray, line 120: same as in hdb/gw here

config/settings/hdb.q
```diff
@@ -7,4 +7,3 @@ loadprocesscode:1b              // whether to load the process specific code def
 \d .servers
 CONNECTIONS:()			// list of connections to make at start up
 STARTUP:1b                      // create connections
-
```
jonathonmcmurray, line 10: this seems like an unimportant change, can you remove it from this PR please?

config/settings/rdb.q
```diff
@@ -22,6 +22,7 @@ gc:1b                                       //if true .Q.gc will be called after
 sortcsv:hsym first .proc.getconfigfile["sort.csv"]	//location of csv file
 reloadenabled:0b    //if true, the RDB will not save when .u.end is called but
                     //will clear it's data using reload function (called by the WDB)
+
```
jonathonmcmurray, line 25: unimportant, remove from PR please

tests/syncreload/connections.csv
```diff
@@ -0,0 +1,18 @@
+action,ms,bytes,lang,code,repeat,minver,comment
+beforeany,0,0,q,system "sleep 5",1,,"wait for all procs to go up"
+before,0,0,q,hdbHandle:hopen`::41804:admin:admin,1,,"Get handle to the HDB"
+before,0,0,q,rdb1Handle:hopen`::41802:admin:admin,1,,"Get handle to RDB1"
+before,0,0,q,rdb2Handle:hopen`::41803:admin:admin,1,,"Get handle to RDB2"
+before,0,0,q,wdbHandle:hopen`::41805:admin:admin,1,,"Get handle to WDB"
+before,0,0,q,gwHandle:hopen`::41806:admin:admin,1,,"Get handle to gateway"
```
jonathonmcmurray, line 7: hardcoded ports in unit tests isn't ideal - a future developer should be able to run all unit tests in TorQ easily, without having to faff about with configuring base ports or worrying about port collisions etc.

tests/syncreload/run.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# Handle command-line arguments
+source $KDBTESTS/flagparse.sh
+
+# Path to test directory
+testpath=/home/creid/devTorQ/bin/syncreload
```
jonathonmcmurray, line 7: hardcoded file paths are a no-no - as mentioned above, these need to run without modification for future devs (otherwise, chances are future devs won't bother to run them and the unit tests become useless)

tests/syncreload/connections.csv
jonathonmcmurray, line 1: I think more clear comments are needed in this file - it's not at all clear why numbers of connections should be what they are etc., so it's very unclear what is actually being tested here

tests/syncreload/eod.csv
jonathonmcmurray, line 1: as with the other test CSV, more clear comments needed - not clear why rdb1 & hdb should be updated after calling .u.end on the wdb, but not rdb2 & gw?

code/hdb/hdbstandard.q
```diff
@@ -1,7 +1,28 @@
...
 // Get the relevant HDB attributes
 .proc.getattributes:{`date`tables!(@[value;`date;`date$()];tables[])}
+
+\d .hdb
+
+// dictionary of handles to reload
+reloadcalls:()!();
+
+// function to add handle to reloadcalls dictionary
+po:{[h] if[.proc.proctype in .hdb.connectedProcs;reloadcalls[h]:0b]};
```
jonathonmcmurray, line 23: `.proc.proctype` is the _current_ process's proctype i.e. here it will always be `hdb` - I don't think this is right

code/hdb/hdbstandard.q
```diff
@@ -1,7 +1,28 @@
 // reload function
 reload:{
+	if[.z.w in key .hdb.reloadcalls;
+		.hdb.reloadcalls[.z.w]:1b;
+		$[not all .hdb.reloadcalls; 
+			{.lg.o[`reload;"reload call received from handle ", string[.z.w], "; reload calls pending from handles ", ", "sv string where not .hdb.reloadcalls]; :(::)}[];
+			.lg.o[`reload;"reload call received from handle ", string[.z.w], "; no more reload calls pending"];
```
jonathonmcmurray, line 7: I think less duplication is needed here

```suggestion
		 .lg.o[`reload;"reload call received from handle ", string[.z.w], $[a:not all .hdb.reloadcalls;"; reload calls pending from handles ", ", "sv string where not .hdb.reloadcalls;""]];
		// return early if there are still pending calls
		if[a;:(::)];
```

or something like this

code/processes/gateway.q
```diff
@@ -637,4 +647,17 @@ neg[h](`.gw.asyncexecjpt;(`.q.system;"sleep 10");enlist[`servertype]!enlist`rdb`
 h(`.gw.syncexec;"`$last .z.x";enlist[`tables]!enlist enlist`logmsgXXX)
 h(`.gw.syncexec;"`$last .z.x";`tables`servertype!(enlist`data;`rdb`hdb))
 
+\
```
jonathonmcmurray, line 650: I guess this works, but I think it'd be better to just move the code further up the file? as a general pattern in TorQ (and other q code), once you hit a big block of comments with usage examples etc., you won't expect to find more actual code further down

config/settings/gateway.q
```diff
@@ -9,6 +9,9 @@ querykeeptime:0D00:30		// the time to keep queries in the
 errorprefix:"error: "		// the prefix for clients to look for in error strings
 clearinactivetime:0D01:00	// the time to keep inactive handle data
 
+// list of process types connected for sync reload
+connectedProcs:()
```
jonathonmcmurray, line 13: we don't use camel case anywhere else in TorQ, so let's not introduce it here please

tests/syncreload/connections.csv
CReid98: I'm having some trouble resolving this issue. I've tried getting the baseport from the system variable and then joining it with sv, but this gives `::/41800/:admin:admin. I'm not sure where to go from here.

jonathonmcmurray: So couple of thoughts - 

1. It should be fairly simple to do string manipulation without using `sv` e.g.

```q
`$"::",getenv[`BASEPORT],":admin:admin"
```

2. TorQ has functions for getting connections to other processes based on proctype/procname without manually constructing connection strings etc. - if we use those, we shouldn't have to worry about any of this

PR #582, opened 2023-07-23
String lambdas in permissions framework

erichards97:


erichards97:
image

jonnypress:
Thanks for this @erichards97. This looks good to me, although we'll probably put the formatting back to how it was... I don't want to devolve into a discussion about which style is better, but at the very least we should remain consistent with the remainder of this file (we realise that TorQ isn't consistent with itself). 

We would like @jonathonmcmurray to take a look and if @jprgrant has a view that would be great. I guess one question would be whether we want to extend to other function types e.g. 104, but I think 100 alone is probably good enough for now.

erichards97:
I've reverted the formatting to the incumbent style, I won't make any comments on that here 😁 

Including 104 sounds good to me, otherwise a similar issue occurs:
image
Does it make sense to cover everything `100h <= x <= 111h`?

code/handlers/permissions.q
```diff
@@ -189,8 +189,8 @@ mainexpr:{[u;e;b;pr]
 expr:mainexpr[;;runmode;permissivemode]
 allowed:mainexpr[;;0b;0b]
 
-parsequery:{[q]q:$[10=type q;q;10h=abs type f:first q;destringf[f],1_ q;q]}
-destringf:{$[(x:`$x)in key`.q;.q x;x~`insert;insert;x]}
+parsequery:{[q]q:$[10=type q;q;10h=abs type f:first q;destringf[f],1_ q;q]};
+destringf:{$[(s:`$x)in key`.q;.q s;s~`insert;insert;100h=type f: @[value; x; 0];f;s]};
```
jonathonmcmurray, line 193: I think if we use `value` here, it could mean that we excute code before we've actually checked permissions? e.g. if first element of query is a string expression that _isn't_ a lambda, such as this query:

```q
(".jm.x:123";`)
```

Regardless of later permissions checks etc., the `.jm.x:123` expression would already have been executed here. I think `parse` should work here without potential side effects?

```suggestion
destringf:{$[(s:`$x)in key`.q;.q s;s~`insert;insert;100h=type f: @[parse; x; 0];f;s]};
```

jonathonmcmurray (commented):
Thanks for this @erichards97 

If you have some example test cases, are you able to add them to https://github.com/DataIntellectTech/TorQ/blob/master/tests/permissions/permissions.csv please? (Or we can add some test cases later after the PR gets merged)

erichards97:
Have switched out `value` for `parse`, and added test cases (which pass!)

jonathonmcmurray (approved):
Looks good to me

PR #583, opened 2023-07-23
Handlers loaded from appcode

erichards97:


erichards97:
I think this also represents a semantic change ( / bugfix) to the config _.proc.loadhandlers_
Previously _.proc.loadhandlers_ only controlled whether code was/was not loaded from KDBSERVCODE, meaning handler code was always loaded from KDBCODE regardless of the config value.

Given the comment attached to this config in default.q, I figured this behaviour was incorrect:
_// whether to load the message handler code defined at ${KDBCODE}/handlers_

So now _.proc.loadhandlers_ determines whether handlers are loaded at all.

jonnypress (changes requested):
Thanks for the change @erichards97 ! Agree on the comment r.e. bugfix
I think we need to change this to not create the x variable e.g. 
`if[.proc.loadhandlers; .proc.loaddir each {x where not ~[""; ] each x} getenv each `KDBCODE`KDBSERVCODE`KDBAPPCODE),\: "/handlers"];`

erichards97:
Thanks - have updated

erichards97:
Thanks - tested and applied

jonathonmcmurray (approved):
Looks good, thanks @erichards97 !

PR #585, opened 2023-07-27
Refactor code and config loading section of torq.q

jamiepeters14:
- initial changes to the loading section
- note still need to do a little more thorough testing to ensure these changes don't change any functionality

### DIFF
--- a/torq.q
+++ b/torq.q
@@ -565,18 +565,14 @@ sys:{[cmd]
 	/-@[{result:system x;.lg.o[`system;"successfully executed"];result};cmd;{.lg.e[`system;"failed to execute ",cmd,": ",x];'x}]
 	};
 
-// load config files 
-/
-loadsettings:{[x]{.proc.loadconfig[getenv[x],"/settings/";]} each `default,.proc.parentproctype,.proc.proctype,.proc.procname};
-\
-func:{[env;procenv]
-	/- wihtout this the logging line within the [] breaks?
-	envstring:string env;
-	$[""~getenv[env];
-	.lg.o[`fileload;"environment variable ",envstring," not set, not loading app specific config"];
-	[procenv::getenv[env],"/settings";
-	.lg.o[`fileload;"environment variable ",envstring," set, loading app specific config"];
-	.proc.loadconfig[procenv;] each `default,.proc.parentproctype,.proc.proctype,.proc.procname]
+// Load additional config files, check if the specific configuration environment variable exists and load the settings if it does
+loadaddcnfg:{[envvar;cnfgpath]
+	evstring:string envvar;
+	$[""~getenv[envvar];
+	.lg.o[`fileload;"environment variable ",evstring," not set, not loading app specific config"];
+	(set[cnfgpath;getenv[envvar],"/settings/"];
+	.lg.o[`fileload;"environment variable ",evstring," set, loading app specific config"];
+	.proc.loadconfig[getenv[envvar],"/settings/";] each `default,.proc.parentproctype,.proc.proctype,.proc.procname)
 	];
 	};
 
@@ -588,9 +584,9 @@ if[not `noconfig in key .proc.params;
 	// load TorQ Default configuration module
 	.proc.loadconfig[getenv[`KDBCONFIG],"/settings/";] each `default,.proc.parentproctype,.proc.proctype,.proc.procname;
     // check if KDBSERVCONFIG is set and load Service Layer specific configuration module
-	.proc.func[`KDBSERVCONFIG;`.proc.servconfig]
+	.proc.loadaddcnfg[`KDBSERVCONFIG;`.proc.servconfig]
 	// check if KDBAPPCONFIG is set and load Appliation specific configuration module 
-	.proc.func[`KDBAPPCONFIG;`.proc.appconfig]
+	.proc.loadaddcnfg[`KDBAPPCONFIG;`.proc.appconfig]
 	// Override config from the command line
 	.proc.override[]]
 
--- a/torq.q
+++ b/torq.q
@@ -543,20 +543,17 @@ loadspeccode:{[ext;dir]
    ];
 	};
 
+// Load each section of code from each directory if it exists
 reloadcommoncode:{
-	// Load common code from each directory if it exists
 	loadspeccode["/common"]'[`KDBCODE`KDBSERVCODE`KDBAPPCODE];
 	};
 reloadparentprocesscode:{
-	// Load parentproctype code from each directory if it exists
 	loadspeccode["/",string parentproctype]'[`KDBCODE`KDBSERVCODE`KDBAPPCODE];
 	};
 reloadprocesscode:{
-	// Load proctype code from each directory if it exists
 	loadspeccode["/",string proctype]'[`KDBCODE`KDBSERVCODE`KDBAPPCODE];
 	};
 reloadnamecode:{
-	// Load procname code from each directory if it exists
 	loadspeccode["/",string procname]'[`KDBCODE`KDBSERVCODE`KDBAPPCODE];
 	};
 
@@ -565,6 +562,22 @@ sys:{[cmd]
 	.lg.o[`system;"executing system command: ",cmd];
 	catcherror:{[cmd;error] .lg.e[`system;"failed to execute ",cmd,": ",error];'error};
 	@[{result:system x;.lg.o[`system;"successfully executed"];result};cmd;catcherror[cmd;]]
+	/-@[{result:system x;.lg.o[`system;"successfully executed"];result};cmd;{.lg.e[`system;"failed to execute ",cmd,": ",x];'x}]
+	};
+
+// load config files 
+/
+loadsettings:{[x]{.proc.loadconfig[getenv[x],"/settings/";]} each `default,.proc.parentproctype,.proc.proctype,.proc.procname};
+\
+func:{[env;procenv]
+	/- wihtout this the logging line within the [] breaks?
+	envstring:string env;
+	$[""~getenv[env];
+	.lg.o[`fileload;"environment variable ",envstring," not set, not loading app specific config"];
+	[procenv::getenv[env],"/settings";
+	.lg.o[`fileload;"environment variable ",envstring," set, loading app specific config"];
+	.proc.loadconfig[procenv;] each `default,.proc.parentproctype,.proc.proctype,.proc.procname]
+	];
 	};
 
 \d . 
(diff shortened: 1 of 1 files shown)


torq.q
```diff
@@ -565,6 +562,18 @@ sys:{[cmd]
...
 	catcherror:{[cmd;error] .lg.e[`system;"failed to execute ",cmd,": ",error];'error};
 	@[{result:system x;.lg.o[`system;"successfully executed"];result};cmd;catcherror[cmd;]]
+	/-@[{result:system x;.lg.o[`system;"successfully executed"];result};cmd;{.lg.e[`system;"failed to execute ",cmd,": ",x];'x}]
+	};
+
+// Load additional config files, check if the specific configuration environment variable exists and load the settings if it does
+loadaddcnfg:{[envvar;cnfgpath]
+	evstring:string envvar;
+	$[""~getenv[envvar];
+	.lg.o[`fileload;"environment variable ",evstring," not set, not loading app specific config"];
```
jonathonmcmurray, line 572: I think we can tidy things up a bit here if we use an `if` instead of ` $` and return early in this case e.g.

```suggestion
	if[""~getenv[envvar];
		.lg.o[`fileload;"environment variable ",evstring," not set, not loading app specific config"];
		:();
	];
```

then the rest of the code doesn't need to be wrapped in a block

torq.q
```diff
@@ -565,6 +562,18 @@ sys:{[cmd]
...
+	(set[cnfgpath;getenv[envvar],"/settings/"];
+	.lg.o[`fileload;"environment variable ",evstring," set, loading app specific config"];
+	.proc.loadconfig[getenv[envvar],"/settings/";] each `default,.proc.parentproctype,.proc.proctype,.proc.procname)
```
jonathonmcmurray, line 575: I mentioned above switching to an `if` instead of `$` so this code doesn't need to be wrapped in a block. However atm you don't actually even have it in a block (i.e. that'd be wrapped with `[]`), instead you have it as a list (wrapped with `()`).

Key difference in this context is that a block (`[]`) will execute expressions left-to-right, but a list (`()`) will excutre right-to-left, which may cause behaviour you're not expecting:

```q
q)[-1"first";-1"second"];
first
second
q)(-1"first";-1"second");
second
first
```

(Of course, if we switch to the `if` with the early return, this is all academic anyway)

torq.q
```diff
@@ -565,6 +562,18 @@ sys:{[cmd]
...
 	@[{result:system x;.lg.o[`system;"successfully executed"];result};cmd;catcherror[cmd;]]
+	/-@[{result:system x;.lg.o[`system;"successfully executed"];result};cmd;{.lg.e[`system;"failed to execute ",cmd,": ",x];'x}]
+	};
+
+// Load additional config files, check if the specific configuration environment variable exists and load the settings if it does
+loadaddcnfg:{[envvar;cnfgpath]
+	evstring:string envvar;
+	$[""~getenv[envvar];
+	.lg.o[`fileload;"environment variable ",evstring," not set, not loading app specific config"];
+	(set[cnfgpath;getenv[envvar],"/settings/"];
```
jonathonmcmurray, line 573: why is this needed? I think it's probably simpler/cleaner to not pass this `cnfgpath` variable to the function and instead just _return_ the path from the function, then at invocation we can do something like

```q
.proc.appconfig:.proc.loadaddconfig`KDBAPPCODE;
```

torq.q
```diff
@@ -574,20 +583,10 @@ sys:{[cmd]
 if[not `noconfig in key .proc.params;
 	// load TorQ Default configuration module
 	.proc.loadconfig[getenv[`KDBCONFIG],"/settings/";] each `default,.proc.parentproctype,.proc.proctype,.proc.procname;
```
jonathonmcmurray, line 585: I think we should use the new function for this one as well, for consistency

torq.q
```diff
@@ -596,12 +595,10 @@ if[not `noconfig in key .proc.params;
 .proc.loadprocesscode:@[value;`.proc.loadprocesscode;1b];
 .proc.loadnamecode:@[value;`.proc.loadnamecode;0b];
 .proc.loadhandlers:@[value;`.proc.loadhandlers;1b];
-.proc.logroll:@[value;`.proc.logroll;1]
-.lg.o[`init;".proc.loadcommoncode flag set to ",string .proc.loadcommoncode];
-.lg.o[`init;".proc.loadprocesscode flag set to ",string .proc.loadprocesscode];
-.lg.o[`init;".proc.loadnamecode flag set to ",string .proc.loadnamecode];
-.lg.o[`init;".proc.loadhandlers flag set to ",string .proc.loadhandlers];
-.lg.o[`init;".proc.logroll flag set to ",string .proc.logroll];
+.proc.logroll:@[value;`.proc.logroll;1b]
```
jonathonmcmurray, line 598: is this definitely meant to be a boolean? (it probably is, but just in case it was intentionally a long instead)

also while we're fixing it

```suggestion
.proc.logroll:@[value;`.proc.logroll;1b];
```

torq.q
```diff
@@ -596,12 +595,10 @@ if[not `noconfig in key .proc.params;
...
+// Log these flags
+.proc.logflag:{[x].lg.o[`init; string[x]," flag set to ",string value x]}
+.proc.logflag each `.proc.loadcommoncode`.proc.loadprocesscode`.proc.loadnamecode`.proc.loadhandlers`.proc.logroll
```
jonathonmcmurray, line 601:
```suggestion
.proc.logflag each `.proc.loadcommoncode`.proc.loadprocesscode`.proc.loadnamecode`.proc.loadhandlers`.proc.logroll;
```

torq.q
```diff
@@ -565,6 +562,18 @@ sys:{[cmd]
 	.lg.o[`system;"executing system command: ",cmd];
 	catcherror:{[cmd;error] .lg.e[`system;"failed to execute ",cmd,": ",error];'error};
 	@[{result:system x;.lg.o[`system;"successfully executed"];result};cmd;catcherror[cmd;]]
+	/-@[{result:system x;.lg.o[`system;"successfully executed"];result};cmd;{.lg.e[`system;"failed to execute ",cmd,": ",x];'x}]
```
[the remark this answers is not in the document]
jamiepeters14: oops didn't mean to keep that in there was just comparing to see if I felt we needed to assign cathcherror explicitly and forgot to remove it

torq.q
jamiepeters14: I did test with [] aswell and still felt the functionailty was acting strangely (can't remember exactly why now, I think I was hoping to use the `cnfgpath` variable within `.proc.loadconfig`  but hopefully removing this section from any conditionals or `if` blocks etc will remove this issue) I reverted to a list purely cause it was what was there before but thanks for the clarification, and yeh won't matter changing to an `if` block anyway. Would you say in general it is better practice to use [] purely for readability then?

torq.q
jamiepeters14: this is a clever work around, I purely had it as it was the only way I saw to ensure the variable still got set, although with the change to an `if` block with an escape then this variable can hopefully (still need to check) can be called in `.proc.loadconfig` would this change things in terms of simplicity/tidiness

torq.q
jamiepeters14: the comment below is about similar stuff if you want to reply to just one

torq.q
jamiepeters14: every other instance of it being assigned seems to be with a boolean and its currently set to 1b so I assumed its just a typo, was kind of hoping you would have confirmed haha, any ideas how to be more sure than this?

torq.q
jonathonmcmurray: Yeah generally best to use `[]` to combine multiple expressions within one branch of a `$` conditional (although my preference is to avoid having multiple expressions within one branch of a `$` as much as possible, as it can quickly get quite hard to follow with so many `[]` around - there are often ways to avoid it, such as using `if`s with early returns, or using function calls in the `$` branches instead)

torq.q
jamiepeters14: and have logic in the function to avoid the logging? I just figured it was going to be messier to include it in the function and would outweigh the consistency factor

torq.q
jonathonmcmurray: if it's a boolean everywhere else, I think that's fine 👍

torq.q
jonathonmcmurray: I think it's fine to do the logging for this one as well, makes it more consistent anyway

torq.q
```diff
@@ -567,27 +553,28 @@ sys:{[cmd]
 	@[{result:system x;.lg.o[`system;"successfully executed"];result};cmd;catcherror[cmd;]]
 	};
 
+// Load additional config files, check if the specific configuration environment variable exists and load the settings if it does
+loadaddconfig:{[envvar;cnfgpath]
+	evstring:string envvar;
+	if[""~getenv[envvar];
+	.lg.o[`fileload;"environment variable ",evstring," not set, not loading this config"];
+	:()];
```
jonathonmcmurray, line 561:
```suggestion
	if[""~getenv[envvar];
		.lg.o[`fileload;"environment variable ",evstring," not set, not loading this config"];
		:();
	];
```

just a minor formatting thing - adding indentation here & putting the closing `]` on a new line helps make it immediately clear where the conditional lines end

torq.q
```diff
@@ -567,27 +553,28 @@ sys:{[cmd]
 	@[{result:system x;.lg.o[`system;"successfully executed"];result};cmd;catcherror[cmd;]]
 	};
 
+// Load additional config files, check if the specific configuration environment variable exists and load the settings if it does
+loadaddconfig:{[envvar;cnfgpath]
+	evstring:string envvar;
```
jonathonmcmurray, line 558: I'd probably scrap this and just do `string[envvar]` in the two log lines?

jonathonmcmurray (approved):
Looks good, couple of minor points but otherwise I think this is good to go

torq.q
jamiepeters14: I had tried this before and was running into an error will try it and reply below with the exact issue

jamiepeters14: when trying to start a process in debug mode it gives this error
 `'type `
 `[5] /home/jpeter/torqrepo/deploy/torq.q:188: .lg.format:{[loglevel;proctype;proc;id;message] "|"sv string[(.proc.cp[];.z.h;proctype;proc;loglevel;id)],enlist(),message}`

jonathonmcmurray: Hmm, what's the exact `.lg.o` line you have?

jamiepeters14: nevermind, I didn't have `string env` bracketed so obviously it wouldn't work *facepalm*

PR #586, opened 2023-08-02
Update ldap.q 

jdavison33:
Updating ldap.q to work with Kx's kdblap.so 
Load in kdbldap.so


code/handlers/ldap.q
```diff
@@ -32,6 +43,19 @@ unblock:{[usr]
...
+  updDict:defaultDict,customDict;
+  bindSession:.ldap.bind_s[sess;;;]. updDict defaultKeys;
+  bindSession
+  }
+
+/authenticate:{[dict].ldap.bind[.ldap.sessionID;(enlist dict[`bind_dn])!enlist dict[`pass]]}
+
 login:{[user;pass]                                              / validate login attempt
   incache:.ldap.cache user;                                     / get user from inputs
   dict:`version`server`port`bind_dn`pass!(.ldap.version;.ldap.server;.ldap.port;.ldap.buildDN user;pass);
```
jonathonmcmurray, line 61: if we modify this, it simplifies the below call to `.ldap.bind`

```suggestion
  dict:`dn`cred!(.ldap.buildDN user;pass);
```

then `dict` can be passed directly e.g. `.ldap.bind[.ldap.sessionID;dict]` instead of having to rename dictionary keys.

the other keys here aren't needed for `.ldap.bind`, so we can get rid of them

code/handlers/ldap.q
```diff
@@ -45,34 +69,38 @@ login:{[user;pass]                                              / validate login
...
+  /authenticate:{[dict].ldap.bind[.ldap.sessionID;(enlist `$dict[`bind_dn])!enlist `$dict[`pass]]}
+
   authorised:$[all (                                            / check if previously used details match
     incache[`success];                                          / previous attempt was a success
     incache[`time]>.z.p-.ldap.checktime;                        / previous attemp occured within checktime period
     incache[`pass]~np:md5 pass                                  / same password was used
   );
     1b;
-    @[{.ldap.authenticate[x]`success};dict;0b]                  / attempt authentication
+    authenticate:{[dict].ldap.bind[.ldap.sessionID;(`dn;`cred)!(dict[`bind_dn];`$dict[`pass])]} [dict]                 / attempt authentication
```
jonathonmcmurray, line 80: mentioned above, but this becomes simplified. also don't think you need the lambda, and I don't think you do anything with the stored `authenticate` variable?

```suggestion
    .ldap.bind[.ldap.sessionID;dict]                 / attempt authentication
```

Before we have this in an error trap, probably still want it in an error trap.

code/handlers/ldap.q
```diff
@@ -45,34 +69,38 @@ login:{[user;pass]                                              / validate login
...
   ];
  
-  `.ldap.cache upsert (user;np;`$.ldap.server;.ldap.port;.z.p; (1+0^incache`attempts;0) authorised;authorised;0b);  / upsert details of current attempt
+  
 
-  $[authorised;                                                 / display authentication status message
+  `.ldap.cache upsert (user;np;`$.ldap.server;.ldap.port;.z.p; (1+0^incache`attempts;0) authorised[`ReturnCode] ;$[authorised[`ReturnCode]~0i;1b;0b];0b);  / upsert details of current attempt
```
jonathonmcmurray, line 85: I think indexing into `authorised` like this will fail if the user was in the cache? in that case, `authorised` will currently be a true boolean (`1b`), so we can't index like this (easiest solution might be to change the `1b` on line 53 to a dictionary like ``enlist[`ReturnCode]!enlist 0i``)

Also, ``$[authorised[`ReturnCode]~0i;1b;0b]`` - the conditional is redundant, you can just use ``authorised[`ReturnCode]~0i`` which will return a boolean anyway

code/handlers/ldap.q
```diff
@@ -45,34 +69,38 @@ login:{[user;pass]                                              / validate login
...
     .ldap.out"successfully authenticated user ",;
     .ldap.err"failed to authenticate user ",] dict`bind_dn;
  
   if[.ldap.checklimit<=.ldap.cache[user]`attempts;              / if attempt limit reached then block user
     .[`.ldap.cache;(user;`blocked);:;1b];
     .ldap.out"limit reached, user ",dict[`bind_dn]," has been locked out"];
 
-  :authorised;
+  :$[authorised[`ReturnCode]~0i;1b;0b];
+  show authenticate;
```
jonathonmcmurray, line 96: I think this is leftover from some debugging perhaps? Should be removed

code/handlers/ldap.q
```diff
@@ -45,34 +69,38 @@ login:{[user;pass]                                              / validate login
...
+  $[authorised[`ReturnCode]~0i;                                                 / display authentication status message
     .ldap.out"successfully authenticated user ",;
     .ldap.err"failed to authenticate user ",] dict`bind_dn;
  
   if[.ldap.checklimit<=.ldap.cache[user]`attempts;              / if attempt limit reached then block user
     .[`.ldap.cache;(user;`blocked);:;1b];
     .ldap.out"limit reached, user ",dict[`bind_dn]," has been locked out"];
 
-  :authorised;
+  :$[authorised[`ReturnCode]~0i;1b;0b];
```
jonathonmcmurray, line 95: As above, can just use the comparison without conditional

```suggestion
  :authorised[`ReturnCode]~0i;
```

config/settings/default.q
```diff
@@ -133,10 +133,10 @@ errortolerance:3f		// and to an error state when it hasn't heartbeated in errort
 
 \d .ldap
 
-enabled:0b                                  // whether ldap authentication is enabled
-debug:0i				    // debug level for ldap library: 0i = none, 1i=normal, 2i=verbose
+enabled:1b                                  // whether ldap authentication is enabled
+debug:2i				    // debug level for ldap library: 0i = none, 1i=normal, 2i=verbose
 server:"localhost";                         // name of ldap server
-port:0i;                                    // port for ldap server
+port:10389i;                                    // port for ldap server
```
jonathonmcmurray, line 139: I think return all of these to their old default values

lib/l64/l64/kafkaq.so
jonathonmcmurray, line 1: you appear to have added an extra nested `l64/` directory, please remove

code/handlers/ldap.q
```diff
@@ -2,26 +2,37 @@
 // User attempts are cached
 // This is used to allow .z.pw to be integrated with ldap
 
+
+
```
jonathonmcmurray, line 6: remove added blank lines

```suggestion
```

jonathonmcmurray (commented):
Overall, looks good, just some tidying up here.

Can you attach some test evidence to the PR please? (e.g. screenshots of successfully/unsuccessfully connecting to a TorQ process with valid/invalid usernames & passwords)

I think there are some additional functions in the Kx LDAP library that we're not currently loading - we should probably load all of them into the `.ldap` namespace when we initialise, some of them are likely useful to end users even if we don't directly use them in our authentication code.

jonnypress:
What do we have to do here r.e. licences? We are using the KX LDAP lib which is licenced under Apache 2.0, which itself uses openldap which is under a different licence. Do we know how we have to carry those licences across?

code/handlers/ldap.q
```diff
@@ -14,8 +12,8 @@ port:       @[value;`port;0i];                                     / port for ld
 blocktime:  @[value;`blocktime; 0D00:30:00];                       / time before blocked user can attempt authentication
 checklimit: @[value;`checklimit;3];                                / number of attempts before user is temporarily blocked
 checktime:  @[value;`checktime;0D00:05];                           / period for user to reauthenticate without rechecking LDAP server
-buildDNsuf: @[value;`buildDNsuf;""];                               / suffix used for building bind DN
-buildDN:    @[value;`buildDN;{{"uid=",string[x]}}]; / function to build bind DN
+buildDNsuf: @[value;`buildDNsuf;",ou=people,dc=planetexpress,dc=com"];                               / suffix used for building bind DN
```
jonathonmcmurray, line 15: this is only the suffix for the test LDAP server we were using, default should be left empty

code/handlers/ldap.q
```diff
@@ -25,13 +23,21 @@ initialise:{[lib]                                                     / initiali
...
   .ldap.bind_s:lib 2:(`kdbldap_bind_s;4);
   .ldap.err2string:lib 2:(`kdbldap_err2string;1);
+  .ldap.startTLS:lib 2:(`kdbldap_start_tls;1);
+  .ldap.setGlobalOption:lib 2:(`kdbldap_set_global_option;2);
+  .ldap.getOption:lib 2:(`kdbldap_get_option;2);
+  .ldap.getGlobalOption:lib 2:(`kdbldap_get_global_option;1);
+  .ldap.interactive_bind_s:lib 2:(`kdbldap_interactive_bind_s;5);
+  .ldap.search_s:lib 2:(`kdbldap_search_s;8);
+  .ldap.unbind_s:lib 2:(`kdbldap_unbind_s;1);
+  r:.ldap.init[0i;enlist `$.ldap.schema,"://",.ldap.server,":",string .ldap.port];
```
jonathonmcmurray, line 33: should use `.ldap.sessionID` in this line

```suggestion
  r:.ldap.init[.ldap.sessionID;enlist `$.ldap.schema,"://",.ldap.server,":",string .ldap.port];
```

code/handlers/ldap.q
```diff
@@ -50,29 +78,32 @@ login:{[user;pass]                                              / validate login
     incache[`time]>.z.p-.ldap.checktime;                        / previous attemp occured within checktime period
     incache[`pass]~np:md5 pass                                  / same password was used
   );
-    1b;
-    @[{.ldap.authenticate[x]`success};dict;0b]                  / attempt authentication
+    enlist[`ReturnCode]!enlist 0i;
+    .[.ldap.bind;(.ldap.sessionID;dict);0b]                 / attempt authentication
```
jonathonmcmurray, line 82: I think the error trap needs to return the same as line above if the bind fails i.e. the one item dictionary, not a boolean

code/handlers/ldap.q
```diff
@@ -50,29 +78,32 @@ login:{[user;pass]                                              / validate login
...
-  `.ldap.cache upsert (user;np;`$.ldap.server;.ldap.port;.z.p; (1+0^incache`attempts;0) authorised;authorised;0b);  / upsert details of current attempt
+  
+
+  `.ldap.cache upsert (user;np;`$.ldap.server;.ldap.port;.z.p; (1+0^incache`attempts;0) authorised[`ReturnCode] ;authorised[`ReturnCode]~0i;0b);  / upsert details of current attempt
 
-  $[authorised;                                                 / display authentication status message
+  $[authorised[`ReturnCode]~0i;                                                 / display authentication status message
     .ldap.out"successfully authenticated user ",;
-    .ldap.err"failed to authenticate user ",] dict`bind_dn;
+    .ldap.err"failed to authenticate user ",.ldap.err2string authorised[`ReturnCode]] dict[`bind_dn];
```
jonathonmcmurray, line 91: I think there's a `,` missing here and some brackets (note that there's a trailing comma on each of the log lines to join the username onto the log message. Think the username is now `dn` instead of `bind_dn` as well

```suggestion
    .ldap.err"failed to authenticate user ",.ldap.err2string[authorised[`ReturnCode]],] dict[`dn];
```

code/handlers/ldap.q
```diff
@@ -50,29 +78,32 @@ login:{[user;pass]                                              / validate login
...
  
   if[.ldap.checklimit<=.ldap.cache[user]`attempts;              / if attempt limit reached then block user
     .[`.ldap.cache;(user;`blocked);:;1b];
     .ldap.out"limit reached, user ",dict[`bind_dn]," has been locked out"];
```
jonathonmcmurray, line 95:
```suggestion
    .ldap.out"limit reached, user ",dict[`dn]," has been locked out"];
```

another reference to dictionary key that has been renamed now

jonathonmcmurray (commented):
Couple more little things & that nested `lib/l64/l64/` directory is still there

code/handlers/ldap.q
jdavison33: When having the error trap return a dictionary, would it be enlist[`ReturnCode]!enlist 0i, if im reading the login function right, that would assign authorised to the dictionary enlist[`ReturnCode]!enlist 0i, which would then allow authentification even if the bind breaks?

jonathonmcmurray: Good point - yes, should be a non-zero return code, can probably just pick e.g. `-2i` which is "LDAP_LOCAL_ERROR              local error (failed dynamic memory allocation?)" - I think if something fails here we can consider it a "local error".

jdavison33: Good job, will add that in now

jdavison33:
Starting a test TorQ process, and qconning into the process with username and password that are acceptable for the docker image. Running .z.u returns the user `Philip J. Fry
correctuserqcon

The following message returns on the test process to confirm authentication:
userauthenticated

Using a wrong username/password in qcon and running .z.u and no return:
wrong user

Failed to authenticate message on test process, as well as after multiple attempts the user is locked out
failedauthent

code/handlers/ldap.q
```diff
@@ -5,23 +5,42 @@
...
 server:     @[value;`server;"localhost"];                          / name of ldap server
 port:       @[value;`port;0i];                                     / port for ldap server
 blocktime:  @[value;`blocktime; 0D00:30:00];                       / time before blocked user can attempt authentication
 checklimit: @[value;`checklimit;3];                                / number of attempts before user is temporarily blocked
 checktime:  @[value;`checktime;0D00:05];                           / period for user to reauthenticate without rechecking LDAP server
 buildDNsuf: @[value;`buildDNsuf;""];                               / suffix used for building bind DN
-buildDN:    @[value;`buildDN;{{"uid=",string[x],",",buildDNsuf}}]; / function to build bind DN
+buildDN:    @[value;`buildDN;{{"uid",string[x],",",buildDNsuf}}];  / function to build bind DN
+schema:     @[value;`schema;"ldap"];                               / schema for ldap
+version:    @[value;`version;3];                                    / ldap version number
```
jonathonmcmurray, line 18:
```suggestion
version:    @[value;`version;3];                                   / ldap version number 
```

comment alignment is off by one space

code/handlers/ldap.q
```diff
@@ -5,23 +5,42 @@
...
+lib:        `$getenv[`KDBLIB],"/",string[.z.o],"/kdbldap";         / ldap library location
 debug:      @[value;`debug;0i]                                     / debug level for ldap library: 0i = none, 1i=normal, 2i=verbose
 server:     @[value;`server;"localhost"];                          / name of ldap server
 port:       @[value;`port;0i];                                     / port for ldap server
 blocktime:  @[value;`blocktime; 0D00:30:00];                       / time before blocked user can attempt authentication
 checklimit: @[value;`checklimit;3];                                / number of attempts before user is temporarily blocked
 checktime:  @[value;`checktime;0D00:05];                           / period for user to reauthenticate without rechecking LDAP server
 buildDNsuf: @[value;`buildDNsuf;""];                               / suffix used for building bind DN
-buildDN:    @[value;`buildDN;{{"uid=",string[x],",",buildDNsuf}}]; / function to build bind DN
+buildDN:    @[value;`buildDN;{{"uid",string[x],",",buildDNsuf}}];  / function to build bind DN
```
jonathonmcmurray, line 16: missing the `=`

```suggestion
buildDN:    @[value;`buildDN;{{"uid=",string[x],",",buildDNsuf}}];  / function to build bind DN
```

jonathonmcmurray (commented):
Looks good to me - couple of little things here, and then I think we just need to answer Jonny's licensing question before we merge this

jonathonmcmurray:
One more thing - it looks like this PR doesn't remove the old `torqldap.so`, I think it should be removed given we're not going to be using it anymore

code/handlers/ldap.q
```diff
@@ -5,23 +5,42 @@
...
+lib:        `$getenv[`KDBLIB],"/",string[.z.o],"/kdbldap";         / ldap library location
 debug:      @[value;`debug;0i]                                     / debug level for ldap library: 0i = none, 1i=normal, 2i=verbose
 server:     @[value;`server;"localhost"];                          / name of ldap server
 port:       @[value;`port;0i];                                     / port for ldap server
 blocktime:  @[value;`blocktime; 0D00:30:00];                       / time before blocked user can attempt authentication
 checklimit: @[value;`checklimit;3];                                / number of attempts before user is temporarily blocked
 checktime:  @[value;`checktime;0D00:05];                           / period for user to reauthenticate without rechecking LDAP server
 buildDNsuf: @[value;`buildDNsuf;""];                               / suffix used for building bind DN
-buildDN:    @[value;`buildDN;{{"uid=",string[x],",",buildDNsuf}}]; / function to build bind DN
+buildDN:    @[value;`buildDN;{{"uid=",string[x],",",buildDNsuf}}];  / function to build bind DN
```
jonathonmcmurray, line 16:
```suggestion
buildDN:    @[value;`buildDN;{{"uid=",string[x],",",buildDNsuf}}]; / function to build bind DN
```

comment alignment

jdavison33:
Here's the licenses for both the Kx (Apache) and the OpenLDAP:
https://github.com/KxSystems/ldap/blob/master/LICENSE

https://www.openldap.org/software/release/license.html

LICENSE
```diff
@@ -19,3 +19,254 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
+
```
jonathonmcmurray, line 22: I don't think we should be modifying this - this is the license that applies to TorQ, which we aren't changing

docs/Licenses.txt
```diff
@@ -0,0 +1,251 @@
+ Apache License
```
jonathonmcmurray, line 1: I think we should add headers to both the licenses in this file to be explicitly clear what they apply to, similar to example on https://softwareengineering.stackexchange.com/questions/234511/what-is-the-best-practice-for-arranging-third-party-library-licenses-paperwork

docs/Licenses.txt
jonathonmcmurray, line 1: I think we should call this `Licenses-3rd-party.txt` or something, to make it clear that these are licenses for 3rd party libs

docs/Licenses-3rd-Party.txt
```diff
@@ -0,0 +1,257 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+
+----------------------------------------------------------------------------
+                  The OpenLDAP Public License
+                  Version 2.8, 17 August 2003
+               Applies to:
+                -ldap.q
```
jonathonmcmurray, line 210: I'm not sure this is right, I think that bit is our code? I think the openLDAP license only applies to the openLDAP library, which forms part of Kx's kdbldap.so

jonathonmcmurray (commented):
Looks good to me

@jonnypress @drgdavies any thoughts on this? satisfied with how 3rd party licenses are presented etc.?

jonathonmcmurray (approved):
LGTM

PR #588, opened 2023-08-10
Finspace dev

niallnorris:


code/handlers/finspaceservers.q
```diff
@@ -0,0 +1,25 @@
...
+.servers.FINSPACECLUSTERSFILE:@[value; `.servers.FINSPACECLUSTERSFILE; hsym `];
+.servers.FINSPACECONNECTIONMARKER:@[value; `.servers.FINSPACECONNECTIONMARKER;`finspace_connection_string];
+
+.servers.readclustersfile:{[]
+    if[null .servers.FINSPACECLUSTERSFILE; {.lg.e[`readclustersfile; "no finspace clusters file defined"]}];
+    :("SSSS**"; enlist ",") 0: .servers.FINSPACECLUSTERSFILE;
+    };
+
+.servers.getfinspaceclusters:{[]
+    expclusters:.servers.readclustersfile[];
```
jonathonmcmurray, line 11: maybe should error trap this?

code/handlers/finspaceservers.q
```diff
@@ -0,0 +1,25 @@
...
+
+
+
+.servers.getfinspaceconn:{[ptype; pname]
+    id:.j.j[(ptype;pname)];
+    cluster:first exec cluster_name from .servers.getfinspaceclusters[] where proctype=ptype, procname = pname;
+
+    if[null cluster; .lg.w[`finspaceconn; "no available finspace cluster found for ",id]; :`];
+    conn:@[.aws.get_kx_connection_string; cluster; {[id;e] .lg.e[`finspaceconn; "failed to get connection string for ",id," via aws api - ",e]; :`}[id;]];
+    :`$conn
```
jonathonmcmurray, line 24:
```suggestion
    :`$conn;
```

end an explicit return with a `;`

code/handlers/trackservers.q
```diff
@@ -30,6 +30,7 @@ SOCKETTYPE:@[value;`SOCKETTYPE;enlist[`]!enlist `]
 PASSWORDS:@[value;`PASSWORDS;enlist[`]!enlist `]                                                // dict of host:port!user:pass e.g. `:host:1234!`user:pass
 
 
+
```
jonathonmcmurray, line 33: revert whitespace change

```suggestion
```

code/handlers/trackservers.q
```diff
@@ -176,14 +177,23 @@ retryrows:{[rows]
     //which checks if .proc.getattributes is defined on the nontorqprocess and executes it 
     //only if it is defined
     a:{$[not null x;@[x;({.proc.getattributes[]};::);()!()];()!()]};
-
```
jonathonmcmurray, line 179: revert whitespace change

code/handlers/trackservers.q
```diff
@@ -176,14 +177,23 @@ retryrows:{[rows]
...
     update attributes:a each w,startp:?[null w;0Np;.proc.cp[]] from`.servers.SERVERS where i in rows;
     if[count connectedrows:select from`.servers.SERVERS where i in rows,.dotz.liveh0 w;
     connectcustom[connectedrows]]}
 
+// get the connection string to connect to a given process
+// in most cases this is just the hpup, unless we are connecting to a finspace process
+// in this case we need to generate the connection string using the AWS api
+getconnectionstring:{[proctype;procname;hpup]
+    if[`finspace~ `tcp ^ .servers.SOCKETTYPE proctype; 
+        :hpup ^ .servers.getfinspaceconn[proctype; procname]
```
jonathonmcmurray, line 192:
```suggestion
        :hpup ^ .servers.getfinspaceconn[proctype; procname];
```

end explicit return with a `;`

code/handlers/trackservers.q
```diff
@@ -287,6 +297,10 @@ formathp:{[HOST;PORT;IPCTYPE]
     if[ipctype = `unix;
         hpup:lower `$":unix://",port;
     ];
+    if[ipctype = `finspace;
+        // there is no point in generating the AWS connection string at this point as it will expire after a period of time
+        hpup:.servers.FINSPACECONNECTIONMARKER
+        ];
```
jonathonmcmurray, line 303:
```suggestion
        hpup:.servers.FINSPACECONNECTIONMARKER;
    ];
```

add `;` and align closing `]` to opening `if`

code/processes/rdb.q
```diff
@@ -100,6 +102,10 @@ endofday:{[date;processdata]
 	a:{(x;raze exec {(enlist x)!enlist((#);enlist y;x)}'[c;a] from meta x where not null a)}each tables`.;
 	/-save and wipe the tables
 	writedown[hdbdir;date];
+        /-creates new changeset if this is a finspace application
+        if[finspace;
+                    .aws.create_changeset[database;([]input_path:enlist getenv[`KDBSCRATCH];database_path:enlist "/";change_type:enlist "PUT")];
```
jonathonmcmurray, line 107: I guess here is where we add the loop to wait for changeset to be completed? do we need to/are we able to capture some id etc. from the `.aws.create_changeset` call to track the status of the changeset?

jonathonmcmurray (commented):
@niallnorris I know a bunch of these changes aren't yours originally, but a few minor things that could be tidied up here as well I think

code/processes/rdb.q
niallnorris: Yeah, this API returns the changeset ID as part of it's output and there is another aws API (I think its .aws.get_changeset) which checks the current status of a changeset so I think we have everything in place except the HDB reload API

code/common/finspace.q
```diff
@@ -0,0 +1,6 @@
+// Config for setting Finspace specific parameters
+\d .finspace
+
+enabled:@[value;`enabled;0b];                               //whether the application is finspace or on prem - set to false by default
+database:@[value;`database;"database"];                     //name of the finspace database applicable to a certain RDB cluster - Not used if on prem
+clusters:@[value;`clusters;enlist `cluster];                //list of clusters to be reloaded during the rdb end of day (and possibly other uses)
```
jonathonmcmurray, line 6: this should maybe be called `hdbclusters` or something? if I'm interpreting the comment correctly it'll just be a list of HDB clusters that need to be reloaded, right?

niallnorris: I'm currently using this flag in my feedhandler for creating a connection string to the rdb cluster. I can create a flag a seperate flag for that I suppose. I might just create that within the feedhandler itself as it's not something that will need to be used once a TP is available. I will remain this flag to hdbclusters as it will come in handy for HDB reloads

code/common/finspace.q
```diff
@@ -0,0 +1,44 @@
...
+   ];
+   .lg.o[`checkStatus;"Status: ",(res`status)];
+   :res; 
+ };
+
+// Creates a Finspace changeset during the RDB end of day process
+createChangeset:{[db]
+      .lg.o[`createChangeset;"creating changeset for database: ", db];
+      details:.aws.create_changeset[db;([]input_path:enlist getenv[`KDBSCRATCH];database_path:enlist "/";change_type:enlist "PUT")];
+      .lg.o[`createChangeset;("creating changset ",(details`id)," with initial status of ",(details`status))];
```
jonnypress, line 28: Can we please avoid changing the casing e.g. createChangeset
also apiCall above
Also .lg.o second variable doesn't need ()

code/handlers/finspaceservers.q
```diff
@@ -0,0 +1,25 @@
...
+    };
+
+
+
+.servers.getfinspaceconn:{[ptype; pname]
+    id:.j.j[(ptype;pname)];
+    cluster:first exec cluster_name from .servers.getfinspaceclusters[] where proctype=ptype, procname = pname;
+
+    if[null cluster; .lg.w[`finspaceconn; "no available finspace cluster found for ",id]; :`];
+    conn:@[.aws.get_kx_connection_string; cluster; {[id;e] .lg.e[`finspaceconn; "failed to get connection string for ",id," via aws api - ",e]; :`}[id;]];
```
jonnypress, line 23: I don't think this code is going to error correctly
On an error condition, conn variable will be set to null symbol, which is then cast to a symbol. You can't cast a symbol to a symbol.

code/processes/rdb.q
```diff
@@ -87,6 +87,11 @@ endofday:{[date;processdata]
 	/-add date+1 to the rdbpartition global
 	rdbpartition,:: date +1;
 	.lg.o[`rdbpartition;"rdbpartition contains - ","," sv string rdbpartition];
+        / Need to download sym file to scratch directory if this is Finspace application
+        if[.finspace.enabled;
+                        .lg.o[`createChangeset;"downloading sym file to scratch directory for ",.finspace.database];
+                        .aws.get_latest_sym_file[.finspace.database;getenv[`KDBSCRATCH]];
+        ];
 	/-if reloadenabled is true, then set a global with the current table counts and then escape
 	if[reloadenabled;
```
jonnypress, line 96: We need a check at some point (RDB start up ideally) to make sure that it can't be the case that both reloadenabled = 1b and .finspace.enabled = 1b as they don't work together.

code/common/finspace.q
drgdavies, line 28: Convert these to details[`status] annotation please

code/processes/rdb.q
```diff
@@ -100,6 +105,10 @@ endofday:{[date;processdata]
 	a:{(x;raze exec {(enlist x)!enlist((#);enlist y;x)}'[c;a] from meta x where not null a)}each tables`.;
 	/-save and wipe the tables
 	writedown[hdbdir;date];
+        /-creates new changeset if this is a finspace application
+        if[.finspace.enabled;
+                    changeset:.finspace.createChangeset[.finspace.database];
```
drgdavies, line 110: Not a fan of variable setting inside if statements as it creates a scenario where sometimes a variable exists, sometimes not at run time

PR #593, opened 2023-09-07
Fix projection bug in permissions.q

### LINKED ISSUE #591: Permissions framework throws type error for projections
image

The left pane shows a TorQ process with pm enabled. The right shows a vanilla q process connecting to the TorQ process with admin credentials, which should have full access.

Calling a regular function with args (fn1) works fine, but calling a projected function with args (fn2) throws a type error.

The type error comes from `.pm.pdict`, specifically `fp!a` [here](https://github.com/DataIntellectTech/TorQ/blob/master/code/handlers/permissions.q#L79).
The error is recreated below using the last values that `.pm.pdict` was called with from the first screenshot
image

In other instances, `.pm.pdict` returns a dictonary of parameter names to arguments. What is the intention with projections? Should it be trying to extract the fixed parameter and argument from the projection, and join this to the user provided args?

I'm happy to raise a PR for this but not entirely clear what the expected behaviour should be. Thanks 🙂

jamiepeters14:
Add a section to the conditional to specifically deal with projections in pdict function.

This closes #591

### DIFF
--- a/code/handlers/permissions.q
+++ b/code/handlers/permissions.q
@@ -73,7 +73,7 @@ cloneuser:{[u;unew;p] adduser[unew;ul[0] ;ul[1]; value (string (ul:raze exec aut
 
 / permissions check functions
 / making a dictionary of the parameters and the argument values
-pdict:pdict:{[f;a]
+pdict:{[f;a]
   d:enlist[`]!enlist[::];
   d:d,$[not ca:count a; ();
         f~`select; ();
--- a/code/handlers/permissions.q
+++ b/code/handlers/permissions.q
@@ -79,7 +79,7 @@ pdict:{[f;a]
         f~`select; ();
         (1=count a) and (99h=type first a); first a;
         /if projection first obtain a list of function and fixed parameters (fnfp) 
-        104h=type value f; [fnfp:value[value[f]]; (value[fnfp[0]][1])!(fnfp[1]),a];
+        104h=type value f; [fnfp:value value f; (value[fnfp 0][1])!fnfp[1],a];
         /get paramaters and make a dictionary with the arguments
         101h<>type fp:value[value[f]][1]; fp!a;
         ((),(`$string til ca))!a
--- a/tests/permissions/permissions.csv
+++ b/tests/permissions/permissions.csv
@@ -54,11 +54,11 @@ true,,,q,(`a`b`c!1 4 7)~.pm.expr[`tom;"first each d"],,,test expression with eac
 fail,,,q,.pm.expr[`harry;"first each d"],,,harry doesn't have access to d
 
 comment,,,,,,,test passing a symbol argument to both a lambda function and a projection
-true,,,q,.pm.expr[`tom;({[x;y]y~`ABCD};`x;`ABCD)],,,passing a symbol argument to a lambda function
-true,,,q,.pm.expr[`tom;({[x;y]y~`ABCD}[`x];`ABCD)],,,passing a symbol argument to a projection
+true,,,q,.pm.requ[`tom;({[x;y]y~`ABCD};`x;`ABCD)],,,passing a symbol argument to a lambda function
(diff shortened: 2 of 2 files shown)


tests/permissions/permissions.csv
```diff
@@ -54,11 +54,17 @@ true,,,q,(`a`b`c!1 4 7)~.pm.expr[`tom;"first each d"],,,test expression with eac
 fail,,,q,.pm.expr[`harry;"first each d"],,,harry doesn't have access to d
 
 comment,,,,,,,test passing a symbol argument to both a lambda function and a projection
-true,,,q,.pm.expr[`tom;({[x;y]y~`ABCD};`x;`ABCD)],,,passing a symbol argument to a lambda function
-true,,,q,.pm.expr[`tom;({[x;y]y~`ABCD}[`x];`ABCD)],,,passing a symbol argument to a projection
+true,,,q,.pm.requ[`tom;({[x;y]y~`ABCD};`x;`ABCD)],,,passing a symbol argument to a lambda function
+true,,,q,.pm.requ[`tom;({[x;y]y~`ABCD}[`x];`ABCD)],,,passing a symbol argument to a projection
```
jonathonmcmurray, line 58: I don't think these should be changed to `requ`?

tests/permissions/permissions.csv
```diff
@@ -54,11 +54,17 @@ true,,,q,(`a`b`c!1 4 7)~.pm.expr[`tom;"first each d"],,,test expression with eac
...
 true,,,q,.pm.requ[`tom; ({x+y}[1];2)]~.pm.requ[`tom; ("{x+y}[1]";2)],,,passing projection as string returns same as regular projection
-fail,,,q,.pm.requ[`harry; ({x+a};1)]~2,,,harry cannot run regular lambda which accesses var a
-fail,,,q,.pm.requ[`harry; ("{x+a}";1)]~2,,,harry cannot run string lambda which accesses var a
\ No newline at end of file
+fail,,,q,.pm.expr[`harry; ({x+a};1)]~2,,,harry cannot run regular lambda which accesses var a
+fail,,,q,.pm.expr[`harry; ("{x+a}";1)]~2,,,harry cannot run string lambda which accesses var a
+
+comment,,,,,,,test defined projection
+run,,,q,fn:{x+y}[1],,,creating a projection 
+true,,,q,".pm.pdict[`fn;1]~(enlist[`]!enlist[::]),(`x`y)!1 1",,,check output is as expected
```
jonathonmcmurray, line 68: I don't think we need this test - this is an implementation detail rather than an important behaviour. as long as `.pm.expr` works, for the tests we don't need to worry about _how_ it works

code/handlers/permissions.q
```diff
@@ -72,11 +72,18 @@ cloneuser:{[u;unew;p] adduser[unew;ul[0] ;ul[1]; value (string (ul:raze exec aut
...
+/ making a dictionary of the parameters and the argument values
 pdict:{[f;a]
   d:enlist[`]!enlist[::];
-  /checking for parameters,projection
-  d:d,$[not ca:count a;();f~`select;();(1=count a) and (99h=type first a);first a;101h<>type fp:get[get[f]][1]; fp!a; ((),(`$string til ca))!a];
+  d:d,$[not ca:count a; ();
+        f~`select; ();
+        (1=count a) and (99h=type first a); first a;
+        /if projection first obtain a list of function and fixed parameters (fnfp) 
+        104h=type value f; [fnfp:value[value[f]]; (value[fnfp[0]][1])!(fnfp[1]),a];
```
jonathonmcmurray, line 82: I don't think you need the parantheses here, although the first set perhaps help a little with readability.

Can also lose some square brackets I think

```suggestion
        104h=type value f; [fnfp:value value f; (value[fnfp 0][1])!fnfp[1],a];
```

jonathonmcmurray (commented):
did you test and confirm that the new unit test(s) fail without your changes (i.e. on current master)?

jonathonmcmurray (approved):
Looks good

PR #595, opened 2023-09-08
Only initialise LDAP if enabled

jdavison33:
.ldap.initialise was called twice meaning that regardless if ldap was enabled it was still called.

Start a test TorQ process with ldap enabled, and qcon in using correct user name and password
.z.u returns username:
zu good
The q process returns authenticated message (if .ldap.debug is set to 2i in default.q):
fry authenticated
The cache is updated:
authenticated cache

qcon using wrong credentials:
.z.u doesn't return a value:
zu bad
The q process returns a failed to authenticate message
failed to authenticate fr
Continued attempts will block the user
blocked cache

Fixed bug in .ldap.cache where attempts counter was wrong for authenticated users


jonathonmcmurray:
Replaced by #596

PR #596, opened 2023-09-08
Remove repeated line and fix cache bug

jdavison33:
.ldap.initialise was called twice meaning that regardless if ldap was enabled it was still called.

Start a test TorQ process with ldap enabled, and qcon in using correct user name and password
.z.u returns username:
zu good
The q process returns authenticated message (if .ldap.debug is set to 2i in default.q):
fry authenticated
The cache is updated:
authenticated cache

qcon using wrong credentials:
.z.u doesn't return a value:
zu bad
The q process returns a failed to authenticate message
failed to authenticate fr
Continued attempts will block the user
blocked cache

Fixed bug in .ldap.cache where attempts counter was wrong for authenticated users

### DIFF
--- a/code/handlers/ldap.q
+++ b/code/handlers/ldap.q
@@ -39,7 +39,6 @@ initialise:{[lib]                                                     / initiali
  };
 
 sessionID:0i
-.ldap.initialise hsym .ldap.lib
 
 cache:([user:`$()]; pass:(); server:`$(); port:`int$(); time:`timestamp$(); attempts:`long$(); success:`boolean$(); blocked:`boolean$());  / create table to store login attempts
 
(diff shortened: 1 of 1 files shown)


code/handlers/ldap.q
```diff
@@ -84,7 +83,7 @@ login:{[user;pass]                                              / validate login
     .[.ldap.bind;(.ldap.sessionID;dict);enlist[`ReturnCode]!enlist -2i]                 / attempt authentication
   ];
  
-  `.ldap.cache upsert (user;np;`$.ldap.server;.ldap.port;.z.p; $[0=authorised[`ReturnCode];0;1+0^incache`attempts] ;authorised[`ReturnCode]~0i;0b);  / upsert details of current attempt
+  `.ldap.cache upsert (user;np;`$.ldap.server;.ldap.port;.z.p; 1+0^incache`attempts ;authorised[`ReturnCode]~0i;0b);  / upsert details of current attempt
```
jonathonmcmurray, line 86: Is this change correct? Do we want to keep track of successful attempts? I think this was probably intended to track failed attempts

jonathonmcmurray (commented):
I'm not sure if the change around tracking attempts is needed?

code/handlers/ldap.q
jdavison33: I had changed it because the counter for attempts would always be zero for authenticated users, whereas the change now just counts the number of attempts as it did for the failed attempts, although if only counting failed attempts is the idea then the change isn't necessary

jonathonmcmurray: I think it's probably just for counting failed attempts, so a user can be locked out after X failed attempts. WIth it like this, someone could successfully login 100 times, fail once and be blocked, which I don't think we want

jdavison33: That makes sense actually, counting failed attempts only then

jonathonmcmurray (approved):
LGTM

PR #598, opened 2023-09-20
Can specify multiple LDAP servers

erichards97:
The Kx LDAP library allows you to pass in a list of LDAP servers to initialise against. I believe this is done for redundancy - e.g. passing in a list of `(bad server address; good server address)` still works.

The TorQ wrapper doesn't currently support passing multiple servers in, as it constructs the address by joining schema - host - port. Since there's a chance people may use different combinations of schemas and ports, I thought it best to do away with these variables and require the user to provide `.ldap.servers` - a list of `schema://server:port`, which gets passed through to the underlying LDAP library.


jonathonmcmurray (approved):
Thanks for this @erichards97 , looks good!

erichards97:
Thanks @jonathonmcmurray 😃 I think it needs someone from DI to hit the big merge button.

PR #601, opened 2023-09-28
Dotz bug

zanbuchanan:


code/handlers/logusage.q
```diff
@@ -124,9 +124,9 @@ if[enabled;
...
 	.dotz.set[`.z.wc;p1[`wc;.z.wc;]];
 	.dotz.set[`.z.ws;p2[`ws;.z.ws;]];
-	.dotz.set[`.z.exit;p2[`exit;$[.finspace.enabled;.aws_z.exit;.z.exit];]];
+	.dotz.set[`.z.exit;p2[`exit;.dotz.getcommand[`.z.exit];]];
 	.dotz.set[`.z.pg;p2[`pg;.z.pg;]];
-	.dotz.set[`.z.pi;p2[`pi;$[.finspace.enabled;.aws_z.pi;.z.pi];]];
+	.dotz.set[`.z.pi;p2[`pi;.dotz.getcommand[`.z.pi];]];
 	.dotz.set[`.z.ph;p2[`ph;.z.ph;]];
 	.dotz.set[`.z.pp;p2[`pp;.z.pp;]];
 	.dotz.set[`.z.ps;p3[`ps;.z.ps;]];]
```
jonathonmcmurray, line 132: do we need to do the same for the rest of these?

torq.q
```diff
@@ -490,6 +491,8 @@ loadf0:{[reload;x]
   if[not[reload]&x in loadedf;.lg.o[`fileload;"already loaded ",x];:()];
   .lg.o[`fileload;"loading ",x];
   // error trapped loading of file
+  if[.finspace.enabled & x like "/opt/kx/app/db/*";
```
jonathonmcmurray, line 494: having a hardcoded path here doesn't sit right with me, do we think this is a temporary measure or will this be here going forward? If it's going to be kept, can this be based on an env variable or something?

or perhaps we could create a more generic "blocklist" of locations to skip when loading, that could be configured in the FinTorQ app or something (in `env.q` maybe)?

torq.q
```diff
@@ -418,7 +418,8 @@ readprocfile:{[file]
 	// exit if no port passed via command line or specified in config
 	if[null[output`port]&0i=system"p";
 		.err.ex[`readprocfile;"No port passed via -p flag or found in ",string[file],". Parameters are host: ", string[output`host], ", proctype: ", string[output`proctype], ", procname: ",string output`procname;1]]; 
-	if[not[output[`port] = system"p"]& 0i = system"p";
+	// for finspace do nothing as port assignment is handled by AWS and not used for connections on our side
+	if[not[.finspace.enabled]&not[output[`port] = system"p"]& 0i = system"p";
```
jonathonmcmurray, line 422: this will already only be true if `0i = system"p"` i.e. no port is set

so if port assignment is handled by AWS, how are we triggering this code? is there no port set automatically in dev clusters? can we instead just add a command line parameter of `-p 0W` if the port number doesn't matter?

zanbuchanan: For non DEV clusters this will not get hit, but the DEV cluster seem to set it's port later and this gets triggered. I'll test adding -p to see if this can be removed

torq.q
```diff
@@ -490,6 +493,8 @@ loadf0:{[reload;x]
   if[not[reload]&x in loadedf;.lg.o[`fileload;"already loaded ",x];:()];
   .lg.o[`fileload;"loading ",x];
   // error trapped loading of file
+  if[max like[x;]each .proc.blocklist;
```
jonathonmcmurray, line 496: I think `any` makes more logical sense than `max` here (will do the same thing)

```suggestion
  if[any like[x;]each .proc.blocklist;
```

PR #602, opened 2023-09-28
Finspacelogs

cmclaverty:


torq.q
```diff
@@ -190,7 +197,7 @@ getapplication:{$[0 = count a:@[{read0 x};hsym last getconfigfile"application.tx
 // Setting .finspace.enabled here so that the .dotz lib can pick it up
 // standard TorQ - gets reset to false in code/common/finspace.q
 // FinTorQ - gets reset to true in finTorq-App/appconfig/settings/default.q
-format:$[.finspace.enabled:"true"~getenv[`KDBFINSPACE];
+format:$[(`jsonlogs in key .proc.params) or ("true"~getenv[`KDBFINSPACE]);
```
zanbuchanan, line 200: If you need this flag ("true"~getenv[`KDBFINSPACE]) then just set .finspace.enabled here and delete line 183. less duplicate code

torq.q
```diff
@@ -178,6 +179,12 @@ application:""
 getversion:{$[0 = count v:@[{raze string exec version from (("SS ";enlist ",")0: x) where app=`TorQ};hsym`$getenv[`KDBCONFIG],"/dependency.csv";version];version;v]}
 getapplication:{$[0 = count a:@[{read0 x};hsym last getconfigfile"application.txt";application];application;a]}
 
+//Check if running in finspace and if so set necessary flag
+.finspace.enabled:"true"~getenv[`KDBFINSPACE]
```
zanbuchanan, line 183: I would delete this line and just set .finspace.enabled on line 200 since you are keeping the flag there

torq.q
zanbuchanan: Also, you don't need brackets on the right argument. So:

(`jsonlogs in key .proc.params) | .finspace.enabled:"true"~getenv[`KDBFINSPACE];

code/handlers/logusage.q
```diff
@@ -44,9 +44,10 @@ write:{
 // extension function to extend the logging e.g. publish the log message
 ext:{[x]}
 
-// format the string to be written to the file, in json if in finspace
-format:$[.finspace.enabled;{.j.j (`p`id`time`zcmd`proctype`procname`type`ip`user`handle`txtc`meminfo`length`errorcheck)!x};
-                           {"|" sv -3!'x}
+// format the string to be written to the file
+format:$[(`jsonlogs in key .proc.params) or ("true"~getenv[`KDBFINSPACE]);
```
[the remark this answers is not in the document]
zanbuchanan: You also don't need brackets on the right of this arg either

zanbuchanan (commented):
Just a few points

jonathonmcmurray (approved):
LGTM

PR #603, opened 2023-10-04
Update tzinfo file, generated from Java 21 - tzdata 2023c

erichards97:
The current tzinfo file is outdated, e.g. it had the DST change as 2015.10.25 for Europe/Istanbul, when it was actually 2015.11.08

New tzinfo generated using the Java and q code here: https://code.kx.com/q/kb/timezones/

image


erichards97:
Please don't merge yet - there still seems to be an issue with historic Europe/Istanbul timezones

erichards97:
There was an issue in the Java code used to generate tzinfo.csv, which I've raised with Kx [here](https://github.com/KxSystems/cookbook/issues/9). It only considered DST changes, but not when a timezone changed its base offset or what happened when DST was no longer used in a timezone.

I have an updated copy of the Java code to generate tzinfo.csv, which looks at when the offset changes rather than just considering DST. I'm hoping to share this code on the linked PR.

I have committed an updated copy of tzinfo in this PR. A few changes to note:

- Some three-letter timezone codes have been dropped (pasted below). This is due to a change from _java.util.TimeZone_ --> _java.time.ZoneId_. My understanding is that these three-letter codes are considered bad practice, so hopefully this isn't an issue.

- There is only one adjustment value in tzinfo, rather than separate adjustments for the timezone + DST. These individual adjustments weren't used in timezone.q so hopefully this also isn't an issue.
- The timezones cover 1900-2100, rather than 1970-2030.

Comparing the old and new files for the overlapping date range shows they largely match:
image

The larger counts for the old file are due to the dropped three-letter codes, mentioned previously.

EST,HST,MST,ACT,AET,AGT,ART,AST,BET,BST,CAT,CNT,CST,CTT,EAT,ECT,IET,IST,JST,MIT,NET,NST,PLT,PNT,PRT,PST,SST,VST

jonnypress:
Thanks for this @erichards97. Is there anyway we can get the three letter codes back? I highly suspect that people will be using some of the more common shortcodes. I presume all the shortcodes have a 1-to-1 mapping with a city, so maybe for example a solution would be to copy America/New_York to EST on load? Or put a mapping into the tz functionality to handle the shortcodes? My guess would be most of the usage in torq is via the functions rather than accessing the table directly.

erichards97:
Done!
image

erichards97:
tzinfo.csv generation code fyi - https://github.com/KxSystems/cookbook/pull/10

jonathonmcmurray:
Thanks for this @erichards97

PR #604, opened 2023-10-06
Update torQemail.so

jdavison33:
Update torQemail.so to prevent an attachment being attachment to every email sent using TorQ email


jdavison33:
Before, when sending any email, it had the following attachment:
email attachment

With the updated torQemail file, the attachment is removed
no attachment

jonathonmcmurray (approved):

PR #623, opened 2023-12-12
Bugfix .save namespace in rdb 

jamiepeters14:
Fixing the defaulting of two functions in the .save namespace in the rdb equivalent to the wdb changes in #553.


jonathonmcmurray (approved):
LGTM

PR #625, opened 2023-12-15
EOP function for WDB/ RDB's for diconnecting from TP and removing unwanted data

tom-watters-DI:
## Issue Number
625
## Pull Request Label
Enhancement
## Team Member(s) who solved this issue
Tom Watters
## Summary
Added a data cleanup function for use in finspace by the RDB's and WDB's being deactivated state. 
## Details
When EOP is called this function will close the connection to the tickerplant and clear all unwanted data after the given index in .u.icounts being passed in. 
## Checklist from Issue of Things Solved
- Phase 2 feature

### DIFF
--- a/code/common/finspace.q
+++ b/code/common/finspace.q
@@ -41,3 +41,10 @@ notifyhdb:{[cluster;changeset]
       .aws.update_kx_cluster_databases[string[cluster];.aws.sdbs[.aws.db[.finspace.database;changeset[`id];.aws.cache["CACHE_1000";"/"]]];.aws.sdep["NO_RESTART"]]
       // TODO - Also need to figure out the ideal logic if a changeset fails to create. Possibly recreate and re-run notifyhd
    }
+
+// function to close connection to TP and remove unwanted data in WDB and RDB's 
+eopdatacleanup:{[dict]
+    hclose each exec w from .servers.SERVERS where proctype = `segmentedtickerplant;
+    // function to parse icounts dict and remove all data after a given index for RDB and WDB's 
+    {[sym;ind]delete from sym where i >= ind}'[key dict;first each value dict]
+ }
--- a/code/common/finspace.q
+++ b/code/common/finspace.q
@@ -44,7 +44,8 @@ notifyhdb:{[cluster;changeset]
 
 // function to close connection to TP and remove unwanted data in WDB and RDB's 
 eopdatacleanup:{[dict]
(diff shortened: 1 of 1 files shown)


code/common/finspace.q
```diff
@@ -41,3 +41,10 @@ notifyhdb:{[cluster;changeset]
       .aws.update_kx_cluster_databases[string[cluster];.aws.sdbs[.aws.db[.finspace.database;changeset[`id];.aws.cache["CACHE_1000";"/"]]];.aws.sdep["NO_RESTART"]]
       // TODO - Also need to figure out the ideal logic if a changeset fails to create. Possibly recreate and re-run notifyhd
    }
+
+// function to close connection to TP and remove unwanted data in WDB and RDB's 
+eopdatacleanup:{[dict]
+    hclose each exec w from .servers.SERVERS where proctype = `segmentedtickerplant;
```
jonathonmcmurray, line 47: I don't think we should hardcode the proctype - if you have a look at the subscription logic in e.g. rdb etc., I think there's a config variable called `tickerplanttypes` or something similar. It's possible that users could change the proctype etc.

I think we need to either (i) use the `tickerplanttypes` config variable or (ii) use the proctype(s)/handle(s) from the `.sub.SUBSCRIPTIONS` table

code/common/finspace.q
```diff
@@ -41,3 +41,10 @@ notifyhdb:{[cluster;changeset]
...
+    // function to parse icounts dict and remove all data after a given index for RDB and WDB's 
+    {[sym;ind]delete from sym where i >= ind}'[key dict;first each value dict]
```
jonathonmcmurray, line 49: why is the `first each` needed? is it not a single value for each table?

I also think it's confusing to use `sym` as the argument in this lambda, that's the name of a table, right? so something like `t` or `table` would probably be more clear?

tom-watters-DI: From my testing i saw that the `value dictionaryname` enlists each keyed entry, hence the `first each` as we are itterating through the list of lists 

will update the `sym` var to `t` as per your suggestion

code/common/finspace.q
tom-watters-DI: Please see the latest approach in the recent commit, spoken to zan and she has suggested we pull the handle itself to from `.sub.SUBSCIPTIONS`

jonathonmcmurray (approved):

PR #626, opened 2024-01-11
Fix Gateway syncexec bug

jamiepeters14:
Fixing a bug where using .gw.syncexec to query the gateway using qcon or HTTP results in an empty return, now the correct result will be returned if synccallsallowed is set to 1b and an error message if it is set to 0b.

### DIFF
--- a/code/processes/gateway.q
+++ b/code/processes/gateway.q
@@ -451,7 +451,10 @@ syncexecjt:{[query;servertype;joinfunction;timeout]
  asyncexecjpts[query;servertype;joinfunction;();timeout;1b];
  // defer response
  e:@[{-30!x;1b};(::);0b];
- if[not e;[.lg.o[`syncexecjt;"failed to defer; query passed to syncexecjpre36"];:syncexecjpre36[query;servertype;joinfunction]]] 
+ if[not e;
+     .lg.o[`syncexecjt;"failed to defer; query passed to syncexecjpre36"];
+     :syncexecjpre36[query;servertype;joinfunction];
+  ];
  }; 
 
 $[.z.K < 3.6;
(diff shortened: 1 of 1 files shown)


code/processes/gateway.q
```diff
@@ -450,7 +450,8 @@ syncexecjt:{[query;servertype;joinfunction;timeout]
  // doesn't make sense to allow specification of a callback for sync requests
  asyncexecjpts[query;servertype;joinfunction;();timeout;1b];
  // defer response
- @[-30!;(::);()];
+ e:@[{-30!x;1b};(::);0b];
+ if[not e;[.lg.o[`syncexecjt;"failed to defer; query passed to syncexecjpre36"];:syncexecjpre36[query;servertype;joinfunction]]]
```
jonathonmcmurray, line 454:
```suggestion
 if[not e;
     .lg.o[`syncexecjt;"failed to defer; query passed to syncexecjpre36"];
     :syncexecjpre36[query;servertype;joinfunction];
  ];
```

Don't need to use a block (square brackets) inside an `if`, and a bit more readable if split over multiple lines

code/processes/gateway.q
```diff
@@ -450,7 +450,8 @@ syncexecjt:{[query;servertype;joinfunction;timeout]
  // doesn't make sense to allow specification of a callback for sync requests
  asyncexecjpts[query;servertype;joinfunction;();timeout;1b];
  // defer response
- @[-30!;(::);()];
+ e:@[{-30!x;1b};(::);0b];
```
jonathonmcmurray, line 453: if this fails and we've already called `asyncexecjpts` above, does that mean query is already in `.gw.queryqueue`? I think possibly that'll mean the query will be run even though the result won't be sent.

Maybe better to move the call to `asyncexecjpts` until after this & only do that if the defer has succeeded?

jamiepeters14: would you add this as another if like if[e..] or change the original if to a $ conditional, and keep the [] you wanted to remove for clarity?

jamiepeters14: have moved this to after now

jonathonmcmurray (approved):
LGTM

PR #627, opened 2024-01-12
Updating TorQ for Finspace Integration

David2464:
Summary
A number of variables and functions have been moved into a separate file to keep in line with the order in which files are loaded.

Details
- [x] Commented out lines relating to the tickerplant which result in errors if left present.

- [x] Moved various variables and functions from code/processes/wdb.q to code/wdb/writedown.q. This again was necessary to keep in line with the order in which files are loaded.

- [x] Provided lines of code to the wdb.q, sort.q and sortworker.q so that they all load in the code in the new file.


config/settings/sort.q
```diff
@@ -49,3 +49,8 @@ eodwaittime:0D00:00:10.000		// time to wait for async calls to complete at eod
 \d .servers
 CONNECTIONS:`hdb`tickerplant`rdb`gateway	// list of connections to make at start up
 STARTUP:1b					// create connections
+
+\d .proc
+loadprocesscode:1b  // whether to load the process specific code defined at ${KDBCODE}/{process type}
+reloadcode[`wdb]
```
jonathonmcmurray, line 55: I don't think this should be in config - this will I think load code when loading config, this means things will be loaded in a non-standard order which could cause subtle issues in some places.

Why is this needed? Could we instead set the parentproctype of sort/sortworker to be `wdb` (in process.csv), and then the `wdb` process code would be loaded in these processes as well?

code/processes/wdb.q
```diff
@@ -43,8 +43,8 @@ hdbtypes:@[value;`hdbtypes;`hdb];                                          /-lis
 rdbtypes:@[value;`rdbtypes;`rdb];                                          /-list of rdb types to look for and call in rdb reload
 gatewaytypes:@[value;`gatewaytypes;`gateway];                              /-list of gateway types to inform at reload
 tickerplanttypes:@[value;`tickerplanttypes;`tickerplant];                  /-list of tickerplant types to try and make a connection to
-tpconnsleepintv:@[value;`tpconnsleepintv;10];                              /-number of seconds between attempts to connect to the tp
-tpcheckcycles:@[value;`tpcheckcycles;0W];                                  /-number of attempts to connect to tp before process is killed 
+/-tpconnsleepintv:@[value;`tpconnsleepintv;10];                              /-number of seconds between attempts to connect to the tp
```
jonathonmcmurray, line 46: is this no longer used/needed? if so, remove it completely rather than commenting it out

why is it not used/needed anymore? does this have a negative impact on non-finspace installations?

code/processes/wdb.q
```diff
@@ -88,9 +78,8 @@ eodwaittime:@[value;`eodwaittime;0D00:00:10.000];                          /-len
 	`u#exec w from .servers.getservers[`proctype;sortworkertypes;()!();1b;0b]]}]
 
 /- fix any backslashes on windows
-savedir:.os.pthq savedir;
+savedir:.os.pthq savedir;
```
jonathonmcmurray, line 81: added whitespace

code/processes/wdb.q
```diff
@@ -516,7 +477,7 @@ startup:{[]
 	.lg.o[`init;"partition has been set to [savedir]/[", (string partitiontype),"]/[tablename]/", $[writedownmode~`partbyattr;"[parted column(s)]/";""]];
 	if[saveenabled;
 		//check if tickerplant is available and if not exit with error 
-		.servers.startupdepcycles[.wdb.tickerplanttypes;.wdb.tpconnsleepintv;.wdb.tpcheckcycles]; 
+		/-.servers.startupdepcycles[.wdb.tickerplanttypes;.wdb.tpconnsleepintv;.wdb.tpcheckcycles];
```
jonathonmcmurray, line 480: remove if unneeded, but is it unneeded? possibly the condition for the `if` here needs to be changed/added to to prevent this running when we don't want it to, but I think in a typical non-finspace installation we probably still want this to run?

code/processes/wdb.q
```diff
@@ -22,13 +22,11 @@ mode:@[value;`mode;`saveandsort];                                          /-the
                                                                            /-                                              save mode process.  When this is triggered it will sort the
                                                                            /-                                              data on disk, apply attributes and the trigger a reload on the
                                                                            /-                                              rdb and hdb processes
-
```
jonathonmcmurray, line 25: unrelated whitespace change

code/processes/wdb.q
```diff
@@ -22,13 +22,11 @@ mode:@[value;`mode;`saveandsort];                                          /-the
...
 writedownmode:@[value;`writedownmode;`default];                            /-the wdb process can periodically write data to disc and sort at EOD in two ways:
                                                                            /- 1. default                   -       the data is partitioned by [ partitiontype ]
                                                                            /-                                      at EOD the data will be sorted and given attributes according to sort.csv before being moved to hdb
                                                                            /- 2. partbyattr                -       the data is partitioned by [ partitiontype ] and the column(s) assigned the parted attributed in sort.csv
                                                                            /-                                      at EOD the data will be merged from each partiton before being moved to hdb
-
```
jonathonmcmurray, line 31: unrelated whitespace change

jonathonmcmurray (approved):
Looks good, just a couple of whitespace changes to revert then good to merge

jonathonmcmurray (approved):
LGTM

PR #628, opened 2024-01-12
Rts start stall

andrewMorrison95:
1. Added rdbready variable in code/common/finspace.q in order to flag wether or not the next periods rdb is up and ready
2. Added newrdbup[] to finspace.q to change the rdbready variable when the next process completes start up and signals it is ready
3. Added newrdbready to .rdb namespace (declared in code/processes/rdb.q) which , at the end of start up, connects to the old processes and sends them a signal to call .finspace.newrdbup to change their respective instances of .finspace.rdbready.
(tested on homer & finspace. works as expected)

### DIFF
--- a/code/common/finspace.q
+++ b/code/common/finspace.q
@@ -50,7 +50,7 @@ eopdatacleanup:{[dict]
     // function to parse icounts dict and remove all data after a given index for RDB and WDB's 
     {[t;ind]delete from t where i >= ind}'[key dict;first each value dict];
  }
-//send signal to the old rts that new processes are running and ready to take over at start of new period
+//set rdbready to true after signal received from the old rdb, that new processes are running and ready to take over at start of new period
 newrdbup:{[]
         .lg.o[`newrdbup;"received signal from next period rdb, setting rdbready to true"]
         @[`.finspace;`rdbready;:;1b];
--- a/code/common/finspace.q
+++ b/code/common/finspace.q
@@ -5,6 +5,7 @@ enabled:@[value;`enabled;0b];                               //whether the applic
 database:@[value;`database;"database"];                     //name of the finspace database applicable to a certain RDB cluster - Not used if on prem
 
 hdbclusters:@[value;`hdbclusters;enlist `cluster];          //list of clusters to be reloaded during the rdb end of day (and possibly other uses)
+rdbready:@[value;`rdbready;0b];                             //whether or not the rdb is running and ready to take over at the next period- set to false by default
 
 / Runs a .aws api until a certain status has been received
 checkstatus:{[apicall;status;frequency;timeout]
@@ -49,3 +50,8 @@ eopdatacleanup:{[dict]
     // function to parse icounts dict and remove all data after a given index for RDB and WDB's 
     {[t;ind]delete from t where i >= ind}'[key dict;first each value dict];
  }
+//send signal to the old rts that new processes are running and ready to take over at start of new period
+newrdbup:{[]
+        .lg.o[`newrdbup;"received signal from next period rdb, setting rdbready to true"]
+        @[`.finspace;`rdbready;:;1b];
+        };
--- a/code/processes/rdb.q
+++ b/code/processes/rdb.q
@@ -244,3 +244,15 @@ $[.rdb.connectonstart;
 /-GMT offset rounded to nearest 15 mins and added to roll time
 .timer.repeat[.eodtime.nextroll-00:01+{00:01*15*"j"$(`minute$x)%15}(.proc.cp[]-.z.p);0W;1D;
   (`.rdb.timeoutreset;`);"Set rdb timeout to 0 for EOD writedown"];
+
+/-send a signal to the old rdb and wdb (excluding the most recently started process) that the new rdb is ready for the next period.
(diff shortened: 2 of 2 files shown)


code/common/finspace.q
```diff
@@ -48,4 +49,9 @@ eopdatacleanup:{[dict]
     hclose each distinct exec w from .sub.SUBSCRIPTIONS;
     // function to parse icounts dict and remove all data after a given index for RDB and WDB's 
     {[t;ind]delete from t where i >= ind}'[key dict;first each value dict];
- }
\ No newline at end of file
+ }
+//set rdbready to true after signal received from the old rdb, that new processes are running and ready to take over at start of new period
+newrdbup:{[]
+        .lg.o[`newrdbup;"received signal from next period rdb, setting rdbready to true"]
```
zanbuchanan, line 55: indenting doesn't match the rest of the file

code/processes/rdb.q
```diff
@@ -244,3 +244,15 @@ $[.rdb.connectonstart;
 /-GMT offset rounded to nearest 15 mins and added to roll time
 .timer.repeat[.eodtime.nextroll-00:01+{00:01*15*"j"$(`minute$x)%15}(.proc.cp[]-.z.p);0W;1D;
   (`.rdb.timeoutreset;`);"Set rdb timeout to 0 for EOD writedown"];
+
+/-send a signal to the old rdb and wdb (excluding the most recently started process) that the new rdb is ready for the next period.
+.rdb.newrdbready:{[]
+  if[1<count h:exec w from .servers.SERVERS where proctype=`wdb;
+  times:raze{enlist[@[;".proc.starttimeUTC";()]x]!enlist[x]} each h;
```
zanbuchanan, line 251: Indenting needed for the if statement.

code/processes/rdb.q
```diff
@@ -244,3 +244,15 @@ $[.rdb.connectonstart;
...
+  .servers.removerows exec i from `.servers.SERVERS where w=times max key times];
```
zanbuchanan, line 252: Can we have a comment above this to show why we are doing it.

zanbuchanan (commented):
Just indenting and comment notes :)

code/common/finspace.q
andrewMorrison95: indentation seems to be inconsistent in this one, i've adjusted it so that its consistent with the notifyhdb and createchangeset functions above it

code/processes/rdb.q
andrewMorrison95: should be fixed now

code/processes/rdb.q
andrewMorrison95: how does that look, does it need more detail?

code/common/finspace.q
```diff
@@ -5,6 +5,7 @@ enabled:@[value;`enabled;0b];                               //whether the applic
 database:@[value;`database;"database"];                     //name of the finspace database applicable to a certain RDB cluster - Not used if on prem
 
 hdbclusters:@[value;`hdbclusters;enlist `cluster];          //list of clusters to be reloaded during the rdb end of day (and possibly other uses)
+rdbready:@[value;`rdbready;0b];                             //whether or not the rdb is running and ready to take over at the next period- set to false by default
```
jonathonmcmurray, line 8: this will be set in all processes, not just rdb processes - is that intended?

code/processes/rdb.q
```diff
@@ -244,3 +244,15 @@ $[.rdb.connectonstart;
 /-GMT offset rounded to nearest 15 mins and added to roll time
 .timer.repeat[.eodtime.nextroll-00:01+{00:01*15*"j"$(`minute$x)%15}(.proc.cp[]-.z.p);0W;1D;
   (`.rdb.timeoutreset;`);"Set rdb timeout to 0 for EOD writedown"];
+
+/-send a signal to the old rdb and wdb (excluding the most recently started process) that the new rdb is ready for the next period.
+.rdb.newrdbready:{[]
+	if[1<count h:exec w from .servers.SERVERS where proctype=`wdb;
+		times:raze{enlist[@[;".proc.starttimeUTC";()]x]!enlist[x]} each h;
```
jonathonmcmurray, line 251:
```suggestion
		times:(@[;".proc.starttimeUTC";()] each h)!h;
```

I think this should do the same

code/processes/rdb.q
```diff
@@ -244,3 +244,15 @@ $[.rdb.connectonstart;
...
+		/-don't want to send signal to new wdb to indicate the next period rdb is ready - remove it from table
+  		.servers.removerows exec i from `.servers.SERVERS where w=times max key times];
```
jonathonmcmurray, line 253: do we want to do `.servers.removerows` here? I think that'll completely remove the newest wdb from `.servers.SERVERS` which is maybe an issue if we need that connection later? (could be wrong here, maybe it's fine)

andrewMorrison95: David's new function will need it, good spot. Will adjust

code/common/finspace.q
andrewMorrison95: yes for all rdbs/wdbs

jonathonmcmurray (approved):

PR #631, opened 2024-01-18
Updating Torq for compatibility with finspace

David2464:
I have added a new file named endofperiod.q in code/rdb.

This file contains the standard endofperiod function in TorQ. This change has only been made so that the endofperiod function can be overridden when finspace is integrated with it.

### DIFF
--- a/code/processes/wdb.q
+++ b/code/processes/wdb.q
@@ -467,21 +467,6 @@ replayupd:{[f;t;d]
 		savetables[savedir;getpartition[];0b;t]]	
 	}[upd];
 
-startup:{[] 
-	.lg.o[`init;"searching for servers"];
-	.servers.startup[];
-	if[writedownmode~`partbyattr;
-		.lg.o[`init;"writedown mode set to ",(string .wdb.writedownmode)]
-		];
-	.lg.o[`init;"partition has been set to [savedir]/[", (string partitiontype),"]/[tablename]/", $[writedownmode~`partbyattr;"[parted column(s)]/";""]];
-	if[saveenabled;
-		//check if tickerplant is available and if not exit with error 
-			if[not .finspace.enabled;                                                                                                                      /-TODO Remove when tickerplant fixed in finspace
-			.servers.startupdepcycles[.wdb.tickerplanttypes;.wdb.tpconnsleepintv;.wdb.tpcheckcycles];
-		];		
-		subscribe[]; 
-        ];
-	}	
 / - if there is data in the wdb directory for the partition, if there is remove it before replay
 / - is only for wdb processes that are saving data to disk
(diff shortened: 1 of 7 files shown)


eugenetemlock (approved):

jonathonmcmurray (approved):
LGTM

code/rdb/rdbstandard.q
```diff
@@ -11,3 +11,8 @@ moveandclear:{[fromNS;toNS;tab]
  if[tab in key fromNS;
   set[` sv (toNS;tab);0#fromNS tab];
   eval(!;enlist fromNS;();0b;enlist enlist tab)]}
+
+\d .
+
+/-set the upd function in the top level namespace
+upd:.rdb.upd
```
jonathonmcmurray, line 18: have you made sure this works? I think `.rdb.upd` is defined in `processes/rdb.q`, which will be loaded _after_ this file, so this may fail. I think `.rdb.upd` will also usually be defined in config somewhere, but the default definition is in `processes/rdb.q` so if a user _doesn't_ have a definition in config, I think this will fail - we should probably error trap this & put a default here to be safe (same as, or maybe instead of, the one in `processes/rdb.q`)

code/wdb/origstartup.q
```diff
@@ -8,7 +8,7 @@ startup:{[]
         .lg.o[`init;"partition has been set to [savedir]/[", (string partitiontype),"]/[tablename]/", $[writedownmode~`partbyattr;"[parted column(s)]/";""]];
         if[saveenabled;
                 //check if tickerplant is available and if not exit with error
-                        if[not .finspace.enabled;                                                                                                                      /-TODO Remove when tickerplant fixed in finspace
+		if[not .finspace.enabled;                                                       /-TODO Remove when tickerplant fixed in finspace
```
jonathonmcmurray, line 11: whitespace change in this line (I think it was spaces before, and is now tabs)

code/processes/wdb.q
```diff
@@ -527,6 +527,6 @@ upd:.wdb.replayupd;
 / - clear any wdb data in the current partition
 .wdb.clearwdbdata[];
 /- initialise the wdb process
-.wdb.startup[];
+.wdb.startup[`];
```
jonathonmcmurray, line 530: is this change necessary? I don't think this argument will be used, so don't think this should be needed?

jonathonmcmurray (approved):
couple of very minor things, but LGTM

code/processes/wdb.q
zanbuchanan: Ah, that was my fault, I didn't realise it would run without it so I suggested it while one a call with David (just did a mini test and I now understand :) )

PR #632, opened 2024-01-22
Remove cluster.csv for finspace connection strings

eugenetemlock:
## Author

## Summary

change finspaceservers API to ignore clusters.csv and match processname against the name of the cluster itself. Use the AWS native Q API to list clusters

### DIFF
--- a/code/handlers/finspaceservers.q
+++ b/code/handlers/finspaceservers.q
@@ -1,22 +1,16 @@
 .servers.FINSPACEDISC:@[value; `.servers.FINSPACEDISC; 0b];
 .servers.FINSPACECLUSTERSFILE:@[value; `.servers.FINSPACECLUSTERSFILE; hsym `];
 
-.servers.readclustersfile:{[]
-    if[null .servers.FINSPACECLUSTERSFILE; {.lg.e[`readclustersfile; "no finspace clusters file defined"]}];
-    :("SSSS**"; enlist ",") 0: .servers.FINSPACECLUSTERSFILE;
+.servers.listfinspaceclusters:{
+    :@[.aws.list_kx_clusters; `; {.lg.e[`listfinspaceclusters; "Failed to get finspace clusters using the finspace API - ",x]}];
     };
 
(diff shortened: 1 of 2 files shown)


code/handlers/finspaceservers.q
```diff
@@ -1,22 +1,16 @@
...
-    :expclusters ij `cluster_name`cluster_type xkey select `$cluster_name, `$cluster_type, `$status from availclusters where status like "RUNNING";
-    };
-
-.servers.getfinspaceconn:{[ptype; pname]
-    id:.j.j[(ptype;pname)];
-    cluster:first exec cluster_name from .servers.getfinspaceclusters[] where proctype=ptype, procname = pname;
+.servers.getfinspaceconn:{[pname]
+    id:.Q.s1 pname;
+    runningclusters:select `$cluster_name, `$cluster_type, status from .servers.listfinspaceclusters[] where status like "RUNNING";
+    cluster:first exec cluster_name from runningclusters where pname = cluster_name;
```
jonathonmcmurray, line 11: I think this can be simplified?

```suggestion
    cluster:first exec `$cluster_name from .servers.listfinspaceclusters[] where status like "RUNNING",cluster_name=pname;
```

(also minor thing but I think it's more conventional to put the column name on the left of a where clause comparison and the local variable being compared to on the right)

code/handlers/finspaceservers.q
```diff
@@ -1,22 +1,16 @@
...
 
     if[null cluster; .lg.w[`finspaceconn; "no available finspace cluster found for ",id]; :`];
     conn:@[.aws.get_kx_connection_string; cluster; {[id;e] .lg.e[`finspaceconn; "failed to get connection string for ",id," via aws api - ",e]; :`}[id;]];
-    :`$conn;
-    };
+    :`$conn;
```
jonathonmcmurray, line 15: added whitespace

eugenetemlock:
@jonathonmcmurray , updated the code to reflect your comments

PR #633, opened 2024-02-01
Unregister from gw finspace torQ compatibiity

eugenetemlock:
## Author

## Summary

TorQ changes required for: https://github.com/DataIntellectTech/TorQ-Amazon-FinSpace-Starter-Pack/pull/90

## Changes

- move definition of addserversfromconnectiontable to gatewaylib.q, so it can be overwritten in other TorQ Apps while retaining original definition for standard TorQ
- addserverresult will ignore whether the server is active or not IF .finspace.enabled, and populate .gw.results accordingly
- adding deletecluster logic to finspace.q

### DIFF
--- a/code/common/finspace.q
+++ b/code/common/finspace.q
@@ -58,7 +58,8 @@ newrdbup:{[]
 
  // clustername must be of type string
 deletecluster:{[clustername]
-  if[10h<>fType:type clustername; .lg.o[`deletecluster;"clustername must be of type string: 10h, got ",-3!fType]; :(::)];
+  if[not any (10h;-11h)=fType:type clustername; .lg.e[`deletecluster;"clustername must be of type string or symbol: 10h -11h, got ",-3!fType]; :(::)];
+  if[-11h~fType; clustername:string clustername];
   .lg.o[`deletecluster;"Going to delete ",$[""~clustername;"current cluster";"cluster named: ",clustername]];
   .aws.delete_kx_cluster[clustername]; // calling this on an empty string deletes self
   // TODO ZAN Error trap
--- a/code/common/finspace.q
+++ b/code/common/finspace.q
@@ -58,6 +58,7 @@ newrdbup:{[]
 
  // clustername must be of type string
 deletecluster:{[clustername]
+  if[10h<>fType:type clustername; .lg.o[`deletecluster;"clustername must be of type string: 10h, got ",-3!fType]; :(::)];
   .lg.o[`deletecluster;"Going to delete ",$[""~clustername;"current cluster";"cluster named: ",clustername]];
   .aws.delete_kx_cluster[clustername]; // calling this on an empty string deletes self
   // TODO ZAN Error trap
--- a/code/common/finspace.q
+++ b/code/common/finspace.q
@@ -56,7 +56,6 @@ newrdbup:{[]
       @[`.finspace;`rdbready;:;1b];
  };
 
- // clustername must be of type string
 deletecluster:{[clustername]
   if[not any (10h;-11h)=fType:type clustername; .lg.e[`deletecluster;"clustername must be of type string or symbol: 10h -11h, got ",-3!fType]; :(::)];
   if[-11h~fType; clustername:string clustername];
--- a/code/common/finspace.q
+++ b/code/common/finspace.q
@@ -55,3 +55,11 @@ newrdbup:{[]
       .lg.o[`newrdbup;"received signal from next period rdb, setting rdbready to true"];
       @[`.finspace;`rdbready;:;1b];
  };
+
+ // clustername must be of type string
+deletecluster:{[clustername]
+  .lg.o[`deletecluster;"Going to delete ",$[""~clustername;"current cluster";"cluster named: ",clustername]];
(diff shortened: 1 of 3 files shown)


code/common/finspace.q
```diff
@@ -55,3 +55,11 @@ newrdbup:{[]
       .lg.o[`newrdbup;"received signal from next period rdb, setting rdbready to true"];
       @[`.finspace;`rdbready;:;1b];
  };
+
+ // clustername must be of type string
```
jonathonmcmurray, line 59: should we add a check for this & throw a descriptive error if it isn't a string?

code/processes/gateway.q
```diff
@@ -193,7 +193,9 @@ deleteresult:{[queryid] .gw.results : (queryid,()) _ .gw.results}
 
 // add a result coming back from a server
 addserverresult:{[queryid;results]
- serverid:first exec serverid from .gw.servers where active, handle=.z.w;
+ serverid:$[@[get;`.finspace.enabled;0b];
```
jonathonmcmurray, line 196:
```suggestion
 serverid:$[@[value;`.finspace.enabled;0b];
```

conventionally `get` is for file access, for resolving a symbol to variable content use `value`

code/processes/gateway.q
```diff
@@ -193,7 +193,9 @@ deleteresult:{[queryid] .gw.results : (queryid,()) _ .gw.results}
...
+    first exec serverid from .gw.servers where handle=.z.w;
```
jonathonmcmurray, line 197: does a server get removed from `.gw.servers` when it shuts down? if not we may have an issue when a new server comes in and is assigned same handle (I think this possibility of multiple servers in `.gw.servers` with same handle is probably why the `active` flag was included in this query to begin with)

code/common/finspace.q
eugenetemlock: added the following: `if[10h<>fType:type clustername; .lg.o[`deletecluster;"clustername must be of type string: 10h, got ",-3!fType]; :(::)]; `

code/processes/gateway.q
eugenetemlock: resolved now

code/processes/gateway.q
eugenetemlock: Hi @jonathonmcmurray ,
   You raise a good point. The function responsible for adding new servers to ```.gw.servers```, ```.gw.addserversfromconnectiontable``` is redefined in corresponding PR for TorQ-Finspace repo so that it only adds new servers with handles not already included in ```.gw.servers```.  Which means an inactive server with a live handle will not be readded into the table in Finspace. But this is only the case for if we are using the TorQ-Finspace app, hence the check if finspace is enabled

eugenetemlock: I see. I don't think that scenario is very probable, but let me revisit it just in case

eugenetemlock: > does a server get removed from `.gw.servers` when it shuts down? if not we may have an issue when a new server comes in and is assigned same handle (I think this possibility of multiple servers in `.gw.servers` with same handle is probably why the `active` flag was included in this query to begin with)

I'm sorry Jonathan, I went on a tangent without addressing the original question. Yes, on .z.pc, .gw.pc gets called, which basically sets the handle for that server to null - so if rdb3 get assigned handle 5, it should be added as an active server to ```.gw.servers``` as the handle associated with rdb1 is now set to null. Although in Finspace, it isn't always behaving as I expect due to what appears to me as some delays between cluster beginning it's shutdown and the kdb process actually disconnecting?

code/processes/gateway.q
```diff
@@ -281,6 +292,11 @@ removeserverhandle:{[serverh]
 
  // mark the server as inactive
  update handle:0Ni, active:0b, disconnecttime:.proc.cp[] from `.gw.servers where handle=serverh;
+ // finspace check
+ if[@[value;`.finspace.enabled;0b] and @[value;`.finspace.dereginprog;0b];
+   .finspace.deregserverids:.finspace.deregserverids _ serverid;
+   .finspace.dereginprog:"b"$count .finspace.deregserverids _ 0N;
```
jonathonmcmurray, line 298: can golf away 2 chars ⛳ 

```suggestion
   .finspace.dereginprog:0<count .finspace.deregserverids _ 0N;
```

code/common/finspace.q
```diff
@@ -55,3 +55,12 @@ newrdbup:{[]
...
+deletecluster:{[clustername]
+  if[10h<>fType:type clustername; .lg.o[`deletecluster;"clustername must be of type string: 10h, got ",-3!fType]; :(::)];
```
jonathonmcmurray, line 61: if we get a symbol, can we just `string` it here rather than throwing an error? (fine to throw an error for other datatypes, and I think this should probably be an error log rather than out log? i.e. `.lg.e`)

jonathonmcmurray (approved):
looks good

code/common/finspace.q
eugenetemlock: resolved

code/processes/gateway.q
eugenetemlock: resolved

PR #639, opened 2024-03-18
Update get attributes functions

andrewMorrison95:
updating the getattributes functions for both hdb and rdb to be dynamic and return correct attributes for both int and date partitioned systems.

### DIFF
--- a/code/rdb/rdbstandard.q
+++ b/code/rdb/rdbstandard.q
@@ -1,5 +1,5 @@
 // Get the relevant RDB attributes
-.proc.getattributes:{`date`tables!(.rdb.rdbpartition[],();tables[])}
+.proc.getattributes:{`partition`tables!(.rdb.getpartition[],();tables[])}
 
 \d .rdb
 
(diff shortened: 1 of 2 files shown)


jonathonmcmurray (approved):

code/rdb/rdbstandard.q
```diff
@@ -1,5 +1,5 @@
 // Get the relevant RDB attributes
-.proc.getattributes:{`date`tables!(.rdb.rdbpartition[],();tables[])}
+.proc.getattributes:{`partition`tables!(.rdb.getpartition[],();tables[])}
```
zanbuchanan, line 2: Why does .rdb.rdbpartition need changed? Doesn't .rdb.getpartition just return .rdb.rdbparititon?

andrewMorrison95: not sure honestly, i just know that it throws a type error with .rdb.rdbpartition[] when int partitioned. I thought it was because .rdb.rdbpartition is a variable and its being invoked as if its a function? but then it works with date. getpartition[] works in both cases though so i went with it

PR #641, opened 2024-03-25
Dacces api int update

andrewMorrison95:
1) Add logic to cusotmfuncs to check if the hdb is int or date partitioned and return the correct hdbtimerange accordingly.
2) Alter getdata to append the correct type of partition column onto table in event both processes are queried through gateway. 
3) Alter attributes routing function in gateway to adjust the upper limit partition for each process to ensure queries are mapped correctly.
4)Alter adjust queries function in gateway to extract partitions from attributes and deal with both int partitions and date partitions.

### DIFF
--- a/code/dataaccess/customfuncs.q
+++ b/code/dataaccess/customfuncs.q
@@ -28,7 +28,7 @@ rollover:{[tabname;hdbtime;prc]
 partitionrange:{[tabname;hdbtimerange;prc;timecol]
     // Get the partition fields from default rollover 
     hdbtimerange:rollover[tabname;;prc] each hdbtimerange+00:00;
-    partfield:@[value;`.Q.pf;`]
+    partfield:@[value;`.Q.pf;`];
     C:?[.checkinputs.tablepropertiesconfig;((=;`tablename;(enlist tabname));(=;`proctype;(enlist prc)));();(1#`ptc)!1#`primarytimecolumn];
     // Output the partitions allowing for non-primary timecolumn
     @[hdbtimerange;1;+;any timecol=raze C[`ptc]];
--- a/code/dataaccess/customfuncs.q
+++ b/code/dataaccess/customfuncs.q
@@ -26,7 +26,7 @@ rollover:{[tabname;hdbtime;prc]
 //- offset times for non-primary time columns
 // example @[`date$(starttime;endtime);1;+;not `time~`time]
 partitionrange:{[tabname;hdbtimerange;prc;timecol]
-    // Get the partition fields from default rollover
+    // Get the partition fields from default rollover 
     hdbtimerange:rollover[tabname;;prc] each hdbtimerange+00:00;
     partfield:$[()~key`.Q.pf;`;.Q.pf];
     C:?[.checkinputs.tablepropertiesconfig;((=;`tablename;(enlist tabname));(=;`proctype;(enlist prc)));();(1#`ptc)!1#`primarytimecolumn];
--- a/code/dataaccess/customfuncs.q
+++ b/code/dataaccess/customfuncs.q
@@ -26,11 +26,14 @@ rollover:{[tabname;hdbtime;prc]
 //- offset times for non-primary time columns
 // example @[`date$(starttime;endtime);1;+;not `time~`time]
 partitionrange:{[tabname;hdbtimerange;prc;timecol]
-    // Get the partition fields from default rollover 
+    // Get the partition fields from default rollover
     hdbtimerange:rollover[tabname;;prc] each hdbtimerange+00:00;
+    partfield:$[()~key`.Q.pf;`;.Q.pf];
     C:?[.checkinputs.tablepropertiesconfig;((=;`tablename;(enlist tabname));(=;`proctype;(enlist prc)));();(1#`ptc)!1#`primarytimecolumn];
     // Output the partitions allowing for non-primary timecolumn
-    :@[hdbtimerange;1;+;any timecol=raze C[`ptc]]};
+    @[hdbtimerange;1;+;any timecol=raze C[`ptc]];
+    if[partfield=`int;hdbtimerange:`long$`timestamp$hdbtimerange];
+    :hdbtimerange};
 
 // Gets the last rollover
 lastrollover:{:rollover[x;.proc.cp[];`hdb]};
--- a/code/dataaccess/customfuncs.q
+++ b/code/dataaccess/customfuncs.q
@@ -28,7 +28,7 @@ rollover:{[tabname;hdbtime;prc]
 partitionrange:{[tabname;hdbtimerange;prc;timecol]
     // Get the partition fields from default rollover 
     hdbtimerange:rollover[tabname;;prc] each hdbtimerange+00:00;
-    partfield:$[()~key`.Q.pf;`;.Q.pf];
+    partfield:@[value;`.Q.pf;`]
     C:?[.checkinputs.tablepropertiesconfig;((=;`tablename;(enlist tabname));(=;`proctype;(enlist prc)));();(1#`ptc)!1#`primarytimecolumn];
     // Output the partitions allowing for non-primary timecolumn
     @[hdbtimerange;1;+;any timecol=raze C[`ptc]];
--- a/code/dataaccess/getdata.q
+++ b/code/dataaccess/getdata.q
@@ -4,51 +4,55 @@
 getdata:{[inputparams]
   if[.proc.proctype in key inputparams;inputparams:inputparams .proc.proctype];
   requestnumber:.requests.initlogger[inputparams];
-// [input parameters dict] generic function acting as main access point for data retrieval
+  // [input parameters dict] generic function acting as main access point for data retrieval
   if[1b~inputparams`getquery;:.dataaccess.buildquery[inputparams]];
   // validate input passed to getdata
   usersdict:inputparams;
   inputparams:@[.dataaccess.checkinputs;inputparams;.requests.error[requestnumber;]];
-  // log success of checkinputs 
+  // log success of checkinputs
   .lg.o[`getdata;"getdata Request Number: ",(string requestnumber)," checkinputs passed"];
   // extract validated parameters from input dictionary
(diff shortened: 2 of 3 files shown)


code/gateway/dataaccess.q
```diff
@@ -171,27 +173,45 @@ partdict:{[input]
...
-        // converts dates dictionary to timestamps/datetimes
-        dates:$[-15h~tp;{"z"$x};::]{(0D+x 0;x[1]+1D-1)}'[dates];
+    // get the tablename
+    tabname:options[`tablename];
+    // remove duplicate servertypes from the gw.servers
+    servers:select from .gw.servers where i=(first;i)fby servertype;
+    // extract the procs which have the table defined
+    servers:select from servers where {[x;tabname]tabname in @[x;`tables]}[;tabname] each attributes;
+    // create a dictionary of the attributes against servertypes
+    procdict:(exec servertype from servers)!(exec attributes from servers)@'(key each exec attributes from servers)[;0];
```
jonathonmcmurray, line 183: if I'm reading this right, this assumes that the partition type will always be the first key in the attributes dict, is that a safe assumption? generally I don't like enforcing a rigid ordering for dictionaries, one of the big benefits of dictionaries over lists is that we _don't_ have to have things in a particular order as we can lookup by key instead of index.

Perhaps instead we could add a `partitiontype` key to the attributes dict, and use that?

code/gateway/dataaccess.q
```diff
@@ -171,27 +173,45 @@ partdict:{[input]
...
+    // if the response is a dictionary index into the tablename
+    procdict:@[procdict;key procdict;{[x;tabname]if[99h=type x;:x[tabname]];:x}[;tabname]];
+    // create list of all available partitions
+    possParts:raze value procdict;
```
jonathonmcmurray, line 187:
```suggestion
    possparts:raze value procdict;
```

no camelCase in TorQ please

code/gateway/dataaccess.q
```diff
@@ -171,27 +173,45 @@ partdict:{[input]
...
+
+    //group partitions to relevant process
+    partitions:group key[part]where each{within[y;]each value x}[part]'[possParts];
+    partitions:possParts{(min x;max x)}'[partitions];
+    partitions:`timestamp$partitions;
+
+    // adjust the times to account for period end time when int partitioned
+    c:first[x:@[partitions;`hdb]],-1+ first[@[partitions;`rdb]];
+    d:first[@[partitions;`rdb]],options `endtime;
+    partitions:@[@[partitions;`hdb;:;c];`rdb;:;d];
```
jonathonmcmurray, line 197:
```suggestion
    partitions:@[partitions;`hdb`rdb;:;(c;d)];
```

think this should work

code/gateway/dataaccess.q
```diff
@@ -171,27 +173,45 @@ partdict:{[input]
...
+    possParts:raze value procdict;
+
+    //group partitions to relevant process
+    partitions:group key[part]where each{within[y;]each value x}[part]'[possParts];
+    partitions:possParts{(min x;max x)}'[partitions];
+    partitions:`timestamp$partitions;
+
+    // adjust the times to account for period end time when int partitioned
+    c:first[x:@[partitions;`hdb]],-1+ first[@[partitions;`rdb]];
+    d:first[@[partitions;`rdb]],options `endtime;
```
jonathonmcmurray, line 196:
```suggestion
    c:first[x:partitions`hdb],-1+ first[partitions`rdb];
    d:first[partitions`rdb],options `endtime;
```

`@` isn't needed here, IMO clearer without

does `x` get used again? if not, don't set it here

code/gateway/dataaccess.q
```diff
@@ -171,27 +173,45 @@ partdict:{[input]
...
+    partitions:@[@[partitions;`hdb;:;c];`rdb;:;d];
+
+   // if start/end time not a date, then adjust dates parameter for the correct types
+    if[not a:-12h~tp:type start:options`starttime;
+        // converts partitions dictionary to timestamps/datetimes
+        partitions:$[-15h~tp;{"z"$x};::]{(0D+x 0;x[1]+1D-1)}'[partitions];
```
jonathonmcmurray, line 202:
```suggestion
        partitions:$[-15h~tp;"z"$;]{(0D+x 0;x[1]+1D-1)}'[partitions];
```

code/gateway/dataaccess.q
```diff
@@ -171,27 +173,45 @@ partdict:{[input]
...
+    partitions:group key[part]where each{within[y;]each value x}[part]'[possParts];
+    partitions:possParts{(min x;max x)}'[partitions];
+    partitions:`timestamp$partitions;
+
+    // adjust the times to account for period end time when int partitioned
+    c:first[x:@[partitions;`hdb]],-1+ first[@[partitions;`rdb]];
+    d:first[@[partitions;`rdb]],options `endtime;
+    partitions:@[@[partitions;`hdb;:;c];`rdb;:;d];
+
+   // if start/end time not a date, then adjust dates parameter for the correct types
```
jonathonmcmurray, line 199: re-align comment

code/dataaccess/customfuncs.q
```diff
@@ -28,9 +28,12 @@ rollover:{[tabname;hdbtime;prc]
 partitionrange:{[tabname;hdbtimerange;prc;timecol]
     // Get the partition fields from default rollover 
     hdbtimerange:rollover[tabname;;prc] each hdbtimerange+00:00;
+    partfield:$[()~key`.Q.pf;`;.Q.pf];
     C:?[.checkinputs.tablepropertiesconfig;((=;`tablename;(enlist tabname));(=;`proctype;(enlist prc)));();(1#`ptc)!1#`primarytimecolumn];
     // Output the partitions allowing for non-primary timecolumn
-    :@[hdbtimerange;1;+;any timecol=raze C[`ptc]]};
+    @[hdbtimerange;1;+;any timecol=raze C[`ptc]];
+    if[partfield=`int;hdbtimerange:`long$`timestamp$hdbtimerange];
```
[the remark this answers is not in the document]
andrewMorrison95: Yes timerange is actually dates in this func (rollover returns dates), so need to cast to timestamp before casting to long

code/gateway/dataaccess.q
andrewMorrison95: I think you are right, this part is existing logic taken from the .dataaccess.partdict[] function. Can certainly look at changing it but presumably would need to change that existing funciton as well then?

jonathonmcmurray: I think that's OK, I don't think it should be a massive change hopefully

code/dataaccess/getdata.q
```diff
@@ -24,32 +24,36 @@ getdata:{[inputparams]
   if[(.proc.proctype=`rdb);
   // change defaulttime.date to date on rdb process query result
     if[(`$(string .checkinputs.getdefaulttime inputparams),".date") in (cols table);
-      table:?[(cols table)<>`$(string .checkinputs.getdefaulttime[inputparams]),".date";cols table;`date] xcol table];    
-  // adds date column when all columns are quried from the rdb process for both keyed and unkeyed results
-    if[(1 < count inputparams`procs) & (all (cols inputparams`tablename) in (cols table));   
-        table:update date:.z.d from table;                                                    
-      if[98h=type table;table:`date xcols table]                                              
+      table:?[(cols table)<>`$(string .checkinputs.getdefaulttime[inputparams]),".date";cols table;`date] xcol table];
```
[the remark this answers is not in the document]
andrewMorrison95: I don't think so, as I think the vector conditional only ever evaluates to false if the inputparams  table has a date col and check inputs finds it. So date would be the correct return in that case and we must be in a date partitioned system. May be misunderstanding that but it hasn't cause any issues in my testing at least.

code/gateway/dataaccess.q
```diff
@@ -171,27 +173,45 @@ partdict:{[input]
...
+    servers:select from servers where {[x;tabname]tabname in @[x;`tables]}[;tabname] each attributes;
+    // create a dictionary of the attributes against servertypes
+    procdict:(exec servertype from servers)!(exec attributes from servers)@'(key each exec attributes from servers)[;0];
+    // if the response is a dictionary index into the tablename
+    procdict:@[procdict;key procdict;{[x;tabname]if[99h=type x;:x[tabname]];:x}[;tabname]];
+    // create list of all available partitions
+    possParts:raze value procdict;
+
+    //group partitions to relevant process
+    partitions:group key[part]where each{within[y;]each value x}[part]'[possParts];
```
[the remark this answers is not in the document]
andrewMorrison95: Cant get this to work, \:/: returns incorrect dictionary and /:\: returns a type error

code/gateway/dataaccess.q
jonathonmcmurray: formatting is a bit messed up there https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax

I _think_ what you're saying makes sense but hard to see exactly what you're saying

andrewMorrison95: sorry hopefully that has fixed it

jonathonmcmurray: yep looks good - I think you can probably skip the lambda within something like

```q
procdict:(exec servertype from servers)!(exec attributes from servers)[;`partition]
```

code/gateway/dataaccess.q
jonathonmcmurray: this still looks wrong I think, 3 space indent here, 4 space indent on line below

code/gateway/dataaccess.q
andrewMorrison95: reverted it back as retested and the distionary was incorrect format (rdb time range was incorrect and it adds on a key that is (`hdb,`rdb)

code/gateway/dataaccess.q
andrewMorrison95: should be fixed

andrewMorrison95:
All comments should now be addressed (hopefully)

PR #642, opened 2024-04-08
Bypassing .z.pd Function

David2464:
Added finspace flags to skip over the use of .z.pd


code/common/finspace.q
```diff
@@ -39,8 +39,8 @@ notifyhdb:{[cluster;changeset]
       // Ensuring that the changeset has successfully created before doing the HDB reload
       current:.finspace.checkstatus[(`.aws.get_changeset;.finspace.database;changeset[`id]);("COMPLETED";"FAILED");00:01;0wu];
       .lg.o[`notifyhdb;("notifying ",string[cluster]," to repoint to changeset ",changeset[`id])];
-      awsdb:.aws.db[.finspace.database;changeset[`id];.aws.cache["CACHE_1000";"/"];""];
-      .aws.update_kx_cluster_databases[string[cluster];.aws.sdbs[awsdb];.aws.sdep["NO_RESTART"]]
+      awsdb:.aws.db[.finspace.database;changeset[`id];();"dndataview"];
```
eugenetemlock, line 42: hardcoded dataview name NG

code/common/finspace.q
```diff
@@ -39,8 +39,8 @@ notifyhdb:{[cluster;changeset]
...
+      .aws.update_kx_cluster_databases[string[cluster];.aws.sdbs[awsdb];.aws.sdep["ROLLING"]]
```
eugenetemlock, line 43: have we concluded we will only be supporting ROLLING deployment strategies?

code/processes/wdb.q
```diff
@@ -311,44 +323,45 @@ merge:{[dir;pt;tableinfo;mergelimits;hdbsettings;mergemethod]
...
-  /-call the posteod function
-  .save.postreplay[hdbsettings[`hdbdir];pt];
-  $[permitreload; 
-    doreload[pt];
-    if[gc;.gc.run[]];
-    ];
-  };
-	
+  if[not .finspace.enabled;
+          $[(0 < count .z.pd[]) and ((system "s")<0);
```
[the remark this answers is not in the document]
eugenetemlock: Alternatively, one can wrap the second half (deleting the path and reloading) in a helper function, then have a gaurd where if .finspace.enabled we can call the helper function and exit before calling the merge logic.

code/common/finspace.q
eugenetemlock: Yeah, if nothing's been concluded yet, best to keep it as is (within the master branch) or add an if-else statement. It's a bit of a pain, but you can update the function to explicitly use ROLLING in your local branches

David2464: No worries that makes sense, I'll put it back to ROLLING for the meantime.

David2464: Sorry, i mean NO_RESTART

code/common/finspace.q
```diff
@@ -39,8 +39,8 @@ notifyhdb:{[cluster;changeset]
       // Ensuring that the changeset has successfully created before doing the HDB reload
       current:.finspace.checkstatus[(`.aws.get_changeset;.finspace.database;changeset[`id]);("COMPLETED";"FAILED");00:01;0wu];
       .lg.o[`notifyhdb;("notifying ",string[cluster]," to repoint to changeset ",changeset[`id])];
-      awsdb:.aws.db[.finspace.database;changeset[`id];.aws.cache["CACHE_1000";"/"];""];
-      .aws.update_kx_cluster_databases[string[cluster];.aws.sdbs[awsdb];.aws.sdep["NO_RESTART"]]
+      awsdb:.aws.db[.finspace.database;changeset[`id];();".finspace.dataview"];
```
eugenetemlock, line 42: I think you don't need the quotations around .finspace.database. 

Also maybe if we are unconcluded about whether we will support NO_RESTARTS, it may be preferable to encapsulate this in an if-else. Here's what I've been working with for example:

```
      usecache:"/opt/kx/app/scratch"~getenv[`KDBSCRATCH];
      awsdb:$[usecache;
          .aws.db[.finspace.database;changeset[`id];.aws.cache["CACHE_1000";"/"];""];
          .aws.db[.finspace.database;changeset[`id];();.finspace.dataview]
        ];
      .aws.update_kx_cluster_databases[string[cluster];.aws.sdbs[awsdb];.aws.sdep $[usecache;"NO_RESTART";"ROLLING"]]
```

And just locally change the useCache flag locally depending on your use-case

code/processes/wdb.q
```diff
@@ -151,26 +151,37 @@ savetables:$[writedownmode~`partbyattr;savetablesbypart;savetables];
 
 savetodisk:{[] savetables[savedir;getpartition[];0b;] each tablelist[]};
 
-/- eod - flush remaining data to disk
 endofday:{[pt;processdata]
-	.lg.o[`eod;"end of day message received - ",spt:string pt];
+        .lg.o[`eod;"end of day message received - ",spt:string pt];
```
jonathonmcmurray, line 155: can we keep whitespace-only changes to separate PRs please? makes it less clear where there is meaningful change

code/processes/wdb.q
```diff
@@ -311,44 +324,46 @@ merge:{[dir;pt;tableinfo;mergelimits;hdbsettings;mergemethod]
...
-  /-call the posteod function
-  .save.postreplay[hdbsettings[`hdbdir];pt];
-  $[permitreload; 
-    doreload[pt];
-    if[gc;.gc.run[]];
-    ];
-  };
-	
+  /- .z.pd funciton in finspace will cause an error. Add in this flag to skip over the use of .z.pd. This should be temporary and will be removed when issue resolved by AWS.
+  if[not .finspace.enabled;
```
jonathonmcmurray, line 330: does this skip the entire EOD merge when finspace is enabled? is that what's desired?

If so, probably better to just return early if it's enabled e.g.

```q
  if[.finspace.enabled;:()];
```

and leave the rest of function unchanged

code/processes/wdb.q
```diff
@@ -311,44 +324,46 @@ merge:{[dir;pt;tableinfo;mergelimits;hdbsettings;mergemethod]
...
+                          ];
+                  merge[dir;pt;;mergelimits;hdbsettings;mergemethod] peach flip (key tablist;value tablist);
+                  /-clear out in memory table, .merge.partsizes, and call sort worker processes to do the same
+                  .lg.o[`eod;"Delete from partsizes"];
+                  delete from `.merge.partsizes;
+                  {(neg x)({.lg.o[`eod;"Delete from partsizes"];
+                  delete from `.merge.partsizes;
+                  /- run a garbage collection if enabled
+                  if[gc;.gc.run[]]};`);
+                          (neg x)(::)} each .z.pd[];
```
jonathonmcmurray, line 346: is there a meaningful change here that I'm missing? or just formatting change?

code/processes/wdb.q
```diff
@@ -311,44 +324,46 @@ merge:{[dir;pt;tableinfo;mergelimits;hdbsettings;mergemethod]
...
+                          ];
+                  [.lg.o[`merge;"merging on main"];reloadsymfile[.Q.dd[hdbsettings `hdbdir;`sym]];
```
jonathonmcmurray, line 348: as above, is there a meaningful change?

code/processes/wdb.q
```diff
@@ -151,26 +151,37 @@ savetables:$[writedownmode~`partbyattr;savetablesbypart;savetables];
...
-	
+        /- create a dictionary of tables and merge limits, byte or row count limit depending on settings
+        .lg.o[`merge;"merging partitons by ",$[.merge.mergebybytelimit;"byte estimate";"row count"]," limit"];
+        mergelimits:(tablelist[],())!($[.merge.mergebybytelimit;{(count x)#mergenumbytes};{[x] mergenumrows^mergemaxrows[x]}]tablelist[]),();
+        tablist:tablelist[]!{0#value x} each tablelist[];
+        / Need to download sym file to scratch directory if this is Finspace application
+        if[.finspace.enabled;
+                        .lg.o[`createchangeset;"downloading sym file to scratch directory for ",.finspace.database];
+                        .aws.get_latest_sym_file[.finspace.database;getenv[`KDBSCRATCH]];
+        ];
```
jonathonmcmurray, line 166: do we need to do this if save isn't enabled? (i.e. perhaps this should be inside the `if[saveenabled;` below, or maybe even inside the `endofdaysave` function?)

code/processes/wdb.q
David2464: It does not skip the entire eodmerge function. It skips over the first if statement where .z.pd is used. So the function essentially starts at the comment "If path exists, delete it"

code/processes/wdb.q
David2464: Suppose yeah if save isn't enabled then there shouldn't be a changeset created/any saving done.

code/processes/wdb.q
David2464: The only change to the eodmerge and eodsortdate functions are the addition of a  if[not .finspace.enabled; flag in order to skip over the section of the function which uses the .z.pd function. The format is changed just as a result of the extra if statement.

jonathonmcmurray: > The format is changed just as a result of the extra if statement.

this specific line used to be one line and is now two, so I don't think that's completely correct

code/processes/wdb.q
jonathonmcmurray: right, but it's inside that first statement that the actual `merge` function gets called, so we're skipping the merge, right?

David2464: Oh sorry yes i see, it does skip over the merge.

PR #643, opened 2024-04-08
Create deldir.q

andrewMorrison95:
adding function to delete contents of a directory - use case in finspace to clean up scratch directory in both wdb and rdb


code/common/deldir.q
jonathonmcmurray, line 1: possibly makes sense to put this in `os.q`?

andrewMorrison95: Yes makes sense. Should it be in the top level namespace though? should .os be reserved to system commands?

jonathonmcmurray: definitely not top level namespace (nothing should really be defined there by TorQ I don't think). I think it's fine to put it in `.os` namespace, there are other functions there that don't call any system commands

PR #648, opened 2024-04-17
Update hdel func

andrewMorrison95:
update hdeldir function so that it only deletes the contents of the directory and not the parent directory itself.

### DIFF
--- a/code/common/os.q
+++ b/code/common/os.q
@@ -9,9 +9,8 @@ deldir:{system("rm -r ";"rd /s /q ")[NT],pth x}
 hdeldir:{[dirpath]
  diR:{$[11h=type d:key x;raze x,.z.s each` sv/:x,/:d;d]};
  nuke:hdel each desc 1_diR@;
- .lg.o[`deldir;"deleting from  directory"]};
- nuke hsym`$dirpath;
-}
+ .lg.o[`deldir;"deleting from  directory"];
+ nuke hsym`$dirpath}
 md:{if[not Fex x;system"mkdir \"",pth[x],"\""]};
 ren:{system("mv ";"move ")[NT],pth[x]," ",pth y}
 cpy:{system("cp ";"copy ")[NT],pth[x]," ",pth y}
--- a/code/common/os.q
+++ b/code/common/os.q
@@ -7,7 +7,7 @@ ext:{`$(x?".")_(x:string x;x)[i:10h=type x]}
 del:{system("rm ";"del ")[NT],pth x}
 deldir:{system("rm -r ";"rd /s /q ")[NT],pth x}
 hdeldir:{[dirpath]
- dirpath:@[{[x] $[10h~a:type x;x;-11h~a;string x;'`err]};dirpath;{.lg.e[`hdeldir;"Invalid data type for directory path"]}];
+ dirpath:$[10h=a:type dirpath;dirpath;-11h=a;string dirpath;'`type];
  if[()~dirpath;:()];
  diR:{$[11h=type d:key x;raze x,.z.s each` sv/:x,/:d;d]};
  nuke:hdel each desc 1_diR@;
--- a/code/common/os.q
+++ b/code/common/os.q
@@ -9,7 +9,7 @@ deldir:{system("rm -r ";"rd /s /q ")[NT],pth x}
 hdeldir:{[dirpath]
  diR:{$[11h=type d:key x;raze x,.z.s each` sv/:x,/:d;d]};
  nuke:hdel each desc 1_diR@;
- .lg.o[`deldir;"deleting from  directory"];
+ .lg.o[`deldir;"deleting from  directory: ", raze string dirpath];
  nuke hsym`$dirpath}
 md:{if[not Fex x;system"mkdir \"",pth[x],"\""]};
 ren:{system("mv ";"move ")[NT],pth[x]," ",pth y}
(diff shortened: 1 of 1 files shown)


code/common/os.q
```diff
@@ -8,9 +8,9 @@ del:{system("rm ";"del ")[NT],pth x}
 deldir:{system("rm -r ";"rd /s /q ")[NT],pth x}
 hdeldir:{[dirpath]
  diR:{$[11h=type d:key x;raze x,.z.s each` sv/:x,/:d;d]};
- nuke:hdel each desc diR@;
- nuke hsym`$dirpath;
- .lg.o[`deldir;"deleting from  directory"]}
+ nuke:hdel each desc 1_diR@;
+ .lg.o[`deldir;"deleting from  directory: ",raze string dirpath];
```
zanbuchanan, line 12: isn't dirpath already a string? So you are stringing a string which is why you are needing to raze?

zanbuchanan: Since you have a type check above you no longer need `raze string` here. You can assume it's a string at this point.
```suggestion
 .lg.o[`deldir;"deleting from  directory: ",dirpath];
```

code/common/os.q
```diff
@@ -7,10 +7,12 @@ ext:{`$(x?".")_(x:string x;x)[i:10h=type x]}
 del:{system("rm ";"del ")[NT],pth x}
 deldir:{system("rm -r ";"rd /s /q ")[NT],pth x}
 hdeldir:{[dirpath]
+ dirpath:@[{[x] $[10h~a:type x;x;-11h~a;string x;'`err]};dirpath;{.lg.e[`hdeldir;"Invalid data type for directory path"]}];
+ if[()~dirpath;:()];
  diR:{$[11h=type d:key x;raze x,.z.s each` sv/:x,/:d;d]};
- nuke:hdel each desc diR@;
- nuke hsym`$dirpath;
- .lg.o[`deldir;"deleting from  directory"]}
+ nuke:hdel each desc 1_diR@;
```
zanbuchanan, line 13: Is there a case where we could want to also delete the parent dir? Not in finspace but in std torq maybe?
Perhaps this should be a bool input to the function.
`deleteparentdir` or something. Would be false for finspace. And we would only do the `1_` if the bool is false.

code/common/os.q
andrewMorrison95: yes sorry that slipped me by meant to remove that thanks

code/common/os.q
```diff
@@ -7,10 +7,12 @@ ext:{`$(x?".")_(x:string x;x)[i:10h=type x]}
 del:{system("rm ";"del ")[NT],pth x}
 deldir:{system("rm -r ";"rd /s /q ")[NT],pth x}
 hdeldir:{[dirpath]
+ dirpath:@[{[x] $[10h~a:type x;x;-11h~a;string x;'`err]};dirpath;{.lg.e[`hdeldir;"Invalid data type for directory path"]}];
```
zanbuchanan, line 10: This would make it exit the function on error without the need for the check below (you can delete line 11).
```suggestion
 dirpath:@[{[x] $[10h~a:type x;x;-11h~a;string x;'`err]};dirpath;{.lg.e[`hdeldir;"Invalid data type for directory path"];:()}];
```

But, do we want to exit like this or do we actually want to just throw the error and catch it where this function is called?

```suggestion
 dirpath:$[10h~a:type dirpath;x;-11h~a;string x;[.lg.e[`hdeldir;"Invalid data type for directory path: type ",string a];'"Invalid type"]];
```

Or if you don't want to throw you can still just exit here:

```suggestion
 dirpath:$[10h~a:type dirpath;x;-11h~a;string x;[.lg.e[`hdeldir;"Invalid data type for directory path: type ",string a];:()]];
```

jonathonmcmurray: I agree that error should be thrown, no need to log error here IMO - keep it simple

```suggestion
 dirpath:$[10h~a:type dirpath;dirpath;-11h~a;string dirpath;'`type];
```

kdb devs are familiar enough with `'type` errors IMO that it's fine to throw that as the error

jonathonmcmurray (approved):

code/common/os.q
```diff
@@ -7,10 +7,13 @@ ext:{`$(x?".")_(x:string x;x)[i:10h=type x]}
 del:{system("rm ";"del ")[NT],pth x}
 deldir:{system("rm -r ";"rd /s /q ")[NT],pth x}
 hdeldir:{[dirpath]
+ dirpath:$[10h=a:type dirpath;dirpath;-11h=a;string dirpath;'`type];
+ if[()~dirpath;:()];
```
zanbuchanan, line 11: Won't the line above throw an error if it's the wrong type? Will the code ever get to this line and dirpath:()?

zanbuchanan: Oh, and are we catching this error in your other code? and logging it? Or will this kill the function?

code/common/os.q
```diff
@@ -7,10 +7,13 @@ ext:{`$(x?".")_(x:string x;x)[i:10h=type x]}
...
  diR:{$[11h=type d:key x;raze x,.z.s each` sv/:x,/:d;d]};
- nuke:hdel each desc diR@;
- nuke hsym`$dirpath;
- .lg.o[`deldir;"deleting from  directory"]}
+ filelist:diR hsym`$dirpath;
+ if[not pdir;filelist:1_filelist];
```
zanbuchanan, line 14: pdir isn't passed in as a parameter. Is it initialised anywhere?

zanbuchanan (changes requested):
pdir can't be set and isn't initialised (from what I can see)

zanbuchanan (approved):
Looking good!

PR #650, opened 2024-04-19
datecheck function period fix

Lu-Devenny:
Fixed period for datecheck function so it is everyday at midnight, rather than every 20 seconds.


jonathonmcmurray (commented):
Looks good - have you tested and made sure this works fully as expected? (i.e. all reports get run at appropriate times every day)

jonathonmcmurray (approved):

jonnypress:
@Lu-Devenny can you squash and merge this please?

PR #653, opened 2024-05-03
Log to cloudwatch after changeset create

eugenetemlock:
## Author

## Summary

log a message to cloudwatch (to trigger new HDB cluster creation) instead of updating the current HDB

### DIFF
--- a/code/common/finspace.q
+++ b/code/common/finspace.q
@@ -56,7 +56,7 @@ notifyhdb:{[cluster;changeset]
       current:.finspace.checkstatus[(`.aws.get_changeset;.finspace.database;changeset[`id]);("COMPLETED";"FAILED");00:01;0wu];
       //.lg.o[`notifyhdb;("notifying ",string[cluster]," to repoint to changeset ",changeset[`id])];
       .lg.o[`notifyhdb;("changeset ",changeset[`id]," ready, bringing up new hdb cluster")];
-      .lg.o[`notifyhdb;"new rdb ready. create new hdb"];
+      .lg.o[`notifyhdb;"new changeset ready. create new hdb"];
       // TODO - Also need to figure out the ideal logic if a changeset fails to create. Possibly recreate and re-run notifyhd
    }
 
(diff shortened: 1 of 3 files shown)


code/common/finspace.q
```diff
@@ -54,9 +54,9 @@ notifyhdb:{[cluster;changeset]
       
       // Ensuring that the changeset has successfully created before doing the HDB reload
       current:.finspace.checkstatus[(`.aws.get_changeset;.finspace.database;changeset[`id]);("COMPLETED";"FAILED");00:01;0wu];
-      .lg.o[`notifyhdb;("notifying ",string[cluster]," to repoint to changeset ",changeset[`id])];
-      awsdb:.aws.db[.finspace.database;changeset[`id];.finspace.cache;.finspace.dataview];
-      .aws.update_kx_cluster_databases[string[cluster];.aws.sdbs[awsdb];.aws.sdep[.finspace.hdbreloadmode]]
+      //.lg.o[`notifyhdb;("notifying ",string[cluster]," to repoint to changeset ",changeset[`id])];
```
zanbuchanan, line 57: Can we just delete this line if it is no longer needed? We can always recreate it later if we need it

zanbuchanan (approved):

code/common/finspace.q
```diff
@@ -54,9 +54,8 @@ notifyhdb:{[cluster;changeset]
       
       // Ensuring that the changeset has successfully created before doing the HDB reload
       current:.finspace.checkstatus[(`.aws.get_changeset;.finspace.database;changeset[`id]);("COMPLETED";"FAILED");00:01;0wu];
-      .lg.o[`notifyhdb;("notifying ",string[cluster]," to repoint to changeset ",changeset[`id])];
-      awsdb:.aws.db[.finspace.database;changeset[`id];.finspace.cache;.finspace.dataview];
-      .aws.update_kx_cluster_databases[string[cluster];.aws.sdbs[awsdb];.aws.sdep[.finspace.hdbreloadmode]]
+      .lg.o[`notifyhdb;("changeset ",changeset[`id]," ready, bringing up new hdb cluster")];
+      .lg.o[`notifyhdb;"new changeset ready. create new hdb"];
```
jonathonmcmurray, line 58: do we need both of these? seems a bit redundant?

eugenetemlock: Yes, I can see if being a bit redundant. Right now, the contents of the second log are the what is monitored on AWS in order to trigger the creation of the new HDB cluster (the former are just more informative). I suppose I could point CloudWatch to monitor instances of the latter part of the first log message (ie. "ready, bring up new hdb cluster") instead

eugenetemlock: Resolved now

jonathonmcmurray (approved):

PR #656, opened 2024-06-12
new writedown mode partbyenum and new intraday process

Papamaci444:
TorQ IDB process requirements:
- works with both default and partbyenum writedown modes
- needs a WDB(and a HDB) to be operational
	- the WDB gets written to a different partition at rollover but the location of the sym is static(as it is in the HDB)
- on startup an IDB:
	- connects to the WDB and retrieves settings required
		- HDB, WDB location, current partition, writedown mode
	- registers itself with the WDB process
		- after EOD rollover the WDB initialises an empty partition(with new partition) with all tables present -> WDB is(almost) always ready to be loaded
	- loads the WDB and the HDB sym file
		- records size of sym file on disk
		- records counts of intraday partitions
- can be triggered for intraday reload.
	- reloads the sym file ONLY if its size on disk has changed
	- reloads the DB ONLY if the number of partitions has changed - only in case of partbyenum writedown mode
	- the WDB process triggers the intraday reload on all IDBs which it is aware of
		- this should work as all IDBs register themselves with the WDB process
	- current partition is the input parameter for intradayreload:{[partition]
		- example directory structure: wdb/2024.09.05/0/trade/.d (for partbyenum writedown mode)
									   wdb/2024.09.05/trade/.d (for default writedown mode)
									   hdb/sym
- has a maptoint function defined to support queries when partbyenum writedown method is used
- queries can be routed to IDB via the gateway
	- the gateway does loadbalancing between different IDB instances
- the user has to be mindful when to query the IDB
	- it will have data overlapping with the RDB
		- the gateway/dataaccess api does not solve this issue
- IDB fails fast - if there is an error loading the DB -> it fails


code/processes/idb.q
```diff
@@ -0,0 +1,46 @@
...
+    .lg.o[`reload;"reload command has been called remotely with partition: ", string pt];
+    .lg.o[`reload;"Doing nothing."];
+    .lg.o[`reload;"Finished reloading IDB"];
+ };
+
+/- loads the idb and the sym file. Called by the wdb process after every writedown.
+intradayreload:{[pt]
+    ptdir: ` sv  .idb.savedir,(`$string[pt]),`;
+    if[0=count key ptdir;:()];
+    system"l ", 1_string ptdir;
```
jonnypress, line 14: Could we have an error trap / logging around this load command

config/settings/wdb.q
```diff
@@ -58,7 +63,7 @@ tpcheckcycles:0W
 
 // Server connection details
 \d .servers
-CONNECTIONS:`hdb`tickerplant`rdb`gateway`sort                                               // list of connections to make at start up
+CONNECTIONS:`hdb`tickerplant`rdb`gateway`sort`idb                                           // list of connections to make at start up
```
jonnypress, line 66: I think this should probably be driven from the variables above, e.g. .wdb.hdbtypes,.wdb.tickerplanttypes etc. 
(I know this was probably wrong in the original code- not something you have added!)

config/settings/idb.q
```diff
@@ -0,0 +1,13 @@
...
+\d .servers
+CONNECTIONS:()                                                                              // list of connections to make at start up
+STARTUP:1b                                                                                  // create connections
+
+\d .proc
+loadprocesscode:1b                                                                          // whether to load the process specific code defined at ${KDBCODE}/{process type}
+
+\d .idb
+savedir:hsym`$getenv[`TORQHOME],"/wdbhdb"                                                   // location of the wdb data
+hdbdir:hsym`$getenv[`TORQHOME],"/hdb"                                                       // location of the wdb data
```
jonnypress, line 13: I think these should be relative to TORQDATA rather than TORQHOME
I think what you have aligns with the original version of TorQ, but setenv has grown over time I think and we now have KDBHDB/KDBWDB which are relative to TORQDATA. We've left the original config as it was for backward compatability. 

For consistency we could make KDBIDB, or just make it in here relative to TORQDATA. I leave it up to you.

jonnypress:
Also - given we are adding a new process type, need to extend out the documentation to include that process type (what it does, why, how etc.)

code/processes/wdb.q
```diff
@@ -123,14 +127,34 @@ upserttopartition:{[dir;tablename;tabdata;pt;expttype;expt]
...
+	.lg.o[`save;"saving ",(string tablename)," data to partition ",
+				/- create directory location for selected partiton
+				string directory:` sv .Q.par[dir;pt;`$string i],tablename,`];
+	/- upsert selected data matched on partition to specific directory
+	.lg.o[`save;"directory: ",string directory];
+	.[
+	 upsert;
+	 (directory;r:?[tabdata;{(x;y;(),z)}[in;;]'[expttype;expt];0b;()]);
+	 {[e] .lg.e[`savetablesbypartenum;"Failed to save table to disk : ",e];'e}
+	 ];
```
jonathonmcmurray, line 145: I think it's a little more readable if we define `r` outside of the apply (and add a comment to explain what's happening in the functional select)

```suggestion
	// select from tabdata where expttype in expt
	r:?[tabdata;{(x;y;(),z)}[in;;]'[expttype;expt];0b;()];
	.[upsert;(directory;r);
	 {[e] .lg.e[`savetablesbypartenum;"Failed to save table to disk : ",e];'e}];
```

I'm also wondering about the each-both - from line 135 I'm assuming we're only dealing with one unique partition value here, so do we need the iteration? could it just be something like this?

```q
r:?[tabdata;(=;expttype;first expt);0b;()];
```

code/processes/wdb.q
```diff
@@ -147,11 +171,46 @@ savetablesbypart:{[dir;pt;forcesave;tablename]
 		if[gc;.gc.run[]];
 	];
 	};
+
+savetablesbypartenum:{[dir;pt;forcesave;tablename]
+	savetablesbypartenumcol[dir;pt;forcesave;tablename;enumcol];
+ };
```
jonathonmcmurray, line 177: would it make sense to just have this be a projection?

```suggestion
savetablesbypartenum:savetablesbypartenumcol[;;;;enumcol];
```

(defined after `savetablesbypartenumcol`)

code/processes/wdb.q
```diff
@@ -286,12 +345,14 @@ endofdaysortdate:{[dir;pt;tablist;hdbsettings]
...
+merge:{[dir;pt;tableinfo;mergelimits;hdbsettings;mergemethod;writedownmode]
   setcompression[hdbsettings[`compression]];
   /- get tablename
   tabname:tableinfo[0];
-  /- get list of partition directories for specified table 
-  partdirs:` sv' tabledir,/:k:key tabledir:.Q.par[hsym dir;pt;tabname];
+  /- get list of partition directories for specified table - partbyattr and partbyenum use different folder structure
+  partdirs:$[writedownmode in `partbyenum;
+			 p where 0<count each key each p:` sv' ((-1_` vs p),/:key p:.Q.par[hsym dir;pt;`]),\: tabname;
+			 ` sv' tabledir,/:key tabledir:.Q.par[hsym dir;pt;tabname]];
```
jonathonmcmurray, line 355: indentation inconsistent with other lines in this file, I think your lines use tabs and the rest use spaces

jonathonmcmurray (commented):
Looks good, a few initial comments from me

code/processes/idb.q
Papamaci444: Sure, added now.

config/settings/idb.q
Papamaci444: You are right, added KDBHDB/KDBWDB.

code/common/os.q
```diff
@@ -22,3 +22,4 @@ run:{system"q ",x}
 kill:{[p]@[(`::p);"\\\\";1];}
 sleep:{x:string x; system("sleep ",x;"timeout /t ",x," >nul")[NT]}
 pthq:{[x] $[10h=type x;ssr [x;"\\";"/"];`$ -1 _ ssr [string (` sv x,`);"\\";"/"]]}
+symlink:{[pathtofile;pathtolink] .lg.o[`symlink;"creating symlink for: ",string[pathtofile]," as:",string pathtolink]; system"ln -s ",.os.pth[pathtofile]," ",.os.pth[pathtolink]}
```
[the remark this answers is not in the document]
Papamaci444: For now I copied the createalias function to .os, modified it and used that.

code/idb/idbstandard.q
```diff
@@ -0,0 +1,14 @@
+// reload function
+reload:{[pt]
+	.lg.o[`reload; "reloading IDB for partition: ",string pt];
```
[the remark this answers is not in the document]
Papamaci444: Yeah I was doing a rework while you were reviewing. Removed this function and added some new more meaningful ones.

code/idb/idbstandard.q
```diff
@@ -0,0 +1,14 @@
...
+	//`sym set get .idb.dbpath};
+ };
+
+\d .idb
+gmttime:@[value;`gmttime;1b];                                              /-define whether the process is on gmttime or not
+partitiontype:@[value;`partitiontype;`date];                               /-set type of partition (defaults to `date)
+getpartition:@[value;`getpartition;                                        /-function to determine the partition value
+			   {{@[value;`.idb.currentpartition;
+				   (`date^partitiontype)$(.z.D,.z.d)gmttime]}}];
```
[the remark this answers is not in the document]
Papamaci444: I tried to correct all the mismatching indentation in new version, thanks!

code/idb/idbstandard.q
```diff
@@ -0,0 +1,14 @@
+// reload function
+reload:{[pt]
+	.lg.o[`reload; "reloading IDB for partition: ",string pt];
+	//`sym set get .idb.dbpath};
+ };
+
+\d .idb
+gmttime:@[value;`gmttime;1b];                                              /-define whether the process is on gmttime or not
```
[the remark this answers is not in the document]
Papamaci444: You are right, gmttime looks useless, I've changed it to .proc ones.

code/processes/idb.q
```diff
@@ -0,0 +1,46 @@
+/-default parameters
+\d .idb
+
+reload:{[pt]
+    .lg.o[`reload;"reload command has been called remotely with partition: ", string pt];
+    .lg.o[`reload;"Doing nothing."];
+    .lg.o[`reload;"Finished reloading IDB"];
+ };
```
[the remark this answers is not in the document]
Papamaci444: Nothing anymore.

code/processes/wdb.q
```diff
@@ -123,14 +127,34 @@ upserttopartition:{[dir;tablename;tabdata;pt;expttype;expt]
...
+/- function to upsert to specified directory using enumerated extra partitioning
+upserttopartitionenum:{[dir;tablename;tabdata;pt;expttype;expt]
+	hdbsym:` sv hdbsettings[`hdbdir],`sym;
+	/- enumerate current extra partition agains the hdb sym file
+	i:get[hdbsym]?hdbsym?first expt;
+	.lg.o[`save;"saving ",(string tablename)," data to partition ",
+				/- create directory location for selected partiton
+				string directory:` sv .Q.par[dir;pt;`$string i],tablename,`];
+	/- upsert selected data matched on partition to specific directory
+	.lg.o[`save;"directory: ",string directory];
```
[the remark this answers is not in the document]
Papamaci444: You found some debug code of mine, sorry for that!

code/processes/idb.q
```diff
@@ -0,0 +1,46 @@
...
+    system"l ", 1_string ptdir;
+    symfilepath: ` sv .idb.savedir,`sym;
+    @[load; symfilepath; {.lg.e[`reload;"failed to reload sym file: ",x]}];
+    .idb.currentpartition:pt;
+ };
+
+initpartition:{[pt]
+    .lg.o[`init;"Initializing partition: ", string pt];
+    .lg.o[`init;"Nothing to do"];
+ };
```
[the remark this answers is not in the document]
Papamaci444: I had some debate whether idb should be initialising the db or the wdb process and left this function here for future rework. In new version I decided to push all the db work to wdb and leave idb as a "read only" process.
Removed this function.

config/settings/wdb.q
Papamaci444: Removed this line from code, was unnecessarily(because of what you just said).

code/processes/wdb.q
```diff
@@ -1,4 +1,3 @@
-/-TorQ wdb process - based upon w.q
```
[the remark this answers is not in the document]
Papamaci444: Hmm, it is not. Added it back.

code/idb/idbstandard.q
```diff
@@ -0,0 +1,15 @@
...
+getpartition:@[value;`getpartition;                                        /-function to determine the partition value
+               {{@[value;`.idb.initialpartition;
+                   (`date^partitiontype)$.proc.cd[]]}}];
+
+initialpartition:getpartition[];                                           /-at startup the IDB will try to load this partition. Later on the WDB instructs the IDB which partition to load.
+setupsuccess:0b;                                                           /-the IDB is only considered to be successfully setup if it was able to load the underlying database(there is data and sym file)
+
+\d .proc
+trap:1b;                                                                   /-the IDB won't fail, it just signals error on its stderr. This means that the process will be still alive
+                                                                           /-when there is no database behind. This will prevent IDB failures during EOD merging of the WDB.
```
jonathonmcmurray, line 15: do we want this? I see from comment it's to allow for case where there is no database - can we not just detect that & not error in that case? this changes the standard behaviour of TorQ processes which is to fail fast if there are any errors during start up (which could be any number of things besides there being no database to load)

code/processes/wdb.q
```diff
@@ -28,11 +28,15 @@ writedownmode:@[value;`writedownmode;`default];                            /-the
                                                                            /-                                      at EOD the data will be sorted and given attributes according to sort.csv before being moved to hdb
                                                                            /- 2. partbyattr                -       the data is partitioned by [ partitiontype ] and the column(s) assigned the parted attributed in sort.csv
                                                                            /-                                      at EOD the data will be merged from each partiton before being moved to hdb
+                                                                           /- 3. partbyenum                -       the data is partitioned by [ partitiontype ] and a symbol column with parted attribution assigned in sort.csv
+                                                                           /-                                      at EOD the data will be merged from each partiton before being moved to hdb
+partwritemodes:`partbyattr`partbyenum;
```
jonathonmcmurray, line 33: this section at the top of scripts in TorQ is conventionally reserved for config - I don't think this belongs here

code/processes/idb.q
```diff
@@ -0,0 +1,45 @@
...
+    @[system; "l ", 1_string ptdir; {.lg.e[`load;"failed to reload db from folder: ",string[y]," error: ",x];.idb.setupsuccess:0b}[;ptdir]];
+    symfilepath: ` sv .idb.savedir,`sym;
+    @[load; symfilepath; {.lg.e[`load;"failed to reload sym file: ",x];.idb.setupsuccess:0b}];
+ };
+
+/- reloads the db. Called by wdb process midday/eod.
+intradayreload:{[pt]
+    /- first check if IDB is properly setup, if not, try to setup again
+    if[not .idb.setupsuccess;setup[.idb.savedir;pt]];
+    loaddb[pt];
```
jonathonmcmurray, line 17: if `setup` was called, this has already been done I think - is there harm in redoing?

code/processes/idb.q
```diff
@@ -0,0 +1,45 @@
+/-default parameters
+\d .idb
+
+/- loads the idb and the sym file for given partition
+loaddb:{[pt]
+    ptdir: ` sv  .idb.savedir,(`$string[pt]),`;
```
jonathonmcmurray, line 6: for better or worse, we generally don't use fully-qualified names in TorQ when within the same namespace

```suggestion
    ptdir: ` sv  savedir,(`$string[pt]),`;
```

(and similar elsewhere)

code/processes/idb.q
```diff
@@ -0,0 +1,45 @@
...
+    .idb.setupsuccess:1b;
+    .lg.o[`init;"setup has been called for db dir: ",string[savedir]," and partition: ", string pt];
+    .idb.savedir:savedir;
+    symlinkpath:` sv .idb.savedir,`sym;
+    hdbsymfile:` sv .idb.hdbdir,`sym;
+    /- we create a link here to the hdb symfile - this only has to be done once, and its place is NOT the partition folder
+    /- as that gets manipulated by the wdb continuously
+    $[0=count key symlinkpath;.os.createalias[hdbsymfile;symlinkpath];
+      .lg.o[`init;"symbolic link already exists: ",string symlinkpath]];
+    if[0=count key symlinkpath;.lg.e[`init;"IDB setup failed. IDB is still not setup."];.idb.setupsuccess:0b];
```
jonathonmcmurray, line 33:
```suggestion
    $[()~key symlinkpath;.os.createalias[hdbsymfile;symlinkpath];
      .lg.o[`init;"symbolic link already exists: ",string symlinkpath]];
    if[()~key symlinkpath;.lg.e[`init;"IDB setup failed. IDB is still not setup."];.idb.setupsuccess:0b];
```

code/processes/wdb.q
```diff
@@ -102,149 +107,212 @@ mergemaxrows:{[tabname] mergenumrows^mergenumtab[tabname]}
...
+                /- create directory location for selected partiton
+                string directory:` sv .Q.par[dir;pt;`$string i],tablename,`];
+    /- selecting rows of table with the corresponding symbol
+    r:?[tabdata;enlist(in;first expttype;expt);0b;()];
+    /- upsert selected data matched on partition to specific directory
+    .[
+     upsert;
+     (directory;r);
+     {[e] .lg.e[`savetablesbypartenum;"Failed to save table to disk : ",e];'e}
+     ];
```
jonathonmcmurray, line 147: I think breaking this over so many lines makes it less readable (it's not immediately clear that these lines are a single expression). Perhaps:

```suggestion
    .[upsert;(directory;r);
      {[e] .lg.e[`savetablesbypartenum;"Failed to save table to disk : ",e];'e}];
```

code/processes/wdb.q
```diff
@@ -102,149 +107,212 @@ mergemaxrows:{[tabname] mergenumrows^mergenumtab[tabname]}
...
+    };
+
+/- function to upsert to specified directory using enumerated extra partitioning
+upserttopartitionenum:{[dir;tablename;tabdata;pt;expttype;expt]
+    hdbsym:` sv hdbsettings[`hdbdir],`sym;
+    /- enumerate current extra partition agains the hdb sym file
+    i:get[hdbsym]?hdbsym?first expt;
+    .lg.o[`save;"saving ",(string tablename)," data to partition ",
+                /- create directory location for selected partiton
+                string directory:` sv .Q.par[dir;pt;`$string i],tablename,`];
```
jonathonmcmurray, line 139:
```suggestion
    /- create directory location for selected partiton
    directory:` sv .Q.par[dir;pt;`$string i],tablename,`;
    .lg.o[`save;"saving ",(string tablename)," data to partition ",string directory];
```

code/processes/wdb.q
```diff
@@ -102,149 +107,212 @@ mergemaxrows:{[tabname] mergenumrows^mergenumtab[tabname]}
...
+    ];
+    .lg.o[`track;"appending details to partsizes"];
+    /-key in partsizes are directory to partition, need to drop training slash in directory key
+    .merge.partsizes[first ` vs directory]+:(count r;-22!r);
+    };
+
+/- function to upsert to specified directory using enumerated extra partitioning
+upserttopartitionenum:{[dir;tablename;tabdata;pt;expttype;expt]
+    hdbsym:` sv hdbsettings[`hdbdir],`sym;
+    /- enumerate current extra partition agains the hdb sym file
```
jonathonmcmurray, line 135:
```suggestion
    /- enumerate current extra partition against the hdb sym file
```

code/processes/wdb.q
```diff
@@ -102,149 +107,212 @@ mergemaxrows:{[tabname] mergenumrows^mergenumtab[tabname]}
...
+    i:get[hdbsym]?hdbsym?first expt;
```
jonathonmcmurray, line 136:
```suggestion
    i:`long$hdbsym?first expt;
```

is this the same thing? (I'm not 100% sure what the intent is here, if it's to get the integer value of the enumerated sym, I think this should be the same & possibly more efficient as we don't have to `get` the sym file)

or possibly we've already enumerated this data, in which case even simpler I think

```suggestion
    i:`long$first expt;
```

code/processes/wdb.q
```diff
@@ -102,149 +107,212 @@ mergemaxrows:{[tabname] mergenumrows^mergenumtab[tabname]}
...
+        $[sortenabled;endofdaysort;informsortandreload] . (savedir;pt;tablist;writedownmode;mergelimits;hdbsettings;mergemethod);
+        if[.finspace.enabled;changeset:.finspace.createchangeset[.finspace.database]];
+        ];
+    .lg.o[`eod;"deleting data from ",$[r:writedownmode in partwritemodes;"partsizes";"tabsizes"]];
+    $[r;@[`.merge;`partsizes;0#];@[`.wdb;`tabsizes;0#]];
+    /-notify all finspace hdbs
+    if[.finspace.enabled;.finspace.notifyhdb[;changeset] each .finspace.hdbclusters];
+    .wdb.currentpartition:pt+1;
+    /- in case of partbyenum writedown mode we want to initialise the new partition under 0 with all the table schemas
+    /- then notify idb processes of the new db
```
jonathonmcmurray, line 244: why do we want to do this?

code/processes/wdb.q
```diff
@@ -528,21 +642,23 @@ getsortparams:{[]
 
 
 /- make sure to request connections for all the correct types
-.servers.CONNECTIONS:(distinct .servers.CONNECTIONS,.wdb.hdbtypes,.wdb.rdbtypes,.wdb.gatewaytypes,.wdb.tickerplanttypes,.wdb.sorttypes,.wdb.sortworkertypes) except `
+.servers.CONNECTIONS:distinct .wdb.hdbtypes,.wdb.rdbtypes,.wdb.gatewaytypes,.wdb.tickerplanttypes,.wdb.sorttypes,.wdb.sortworkertypes,.wdb.idbtypes;
```
jonathonmcmurray, line 645: why have we removed `.servers.CONNECTIONS` here? the reason for it being there is in case a user has configured additional processes to connect to in config etc., I think we should still have it here

```suggestion
.servers.CONNECTIONS:(distinct .servers.CONNECTIONS,.wdb.hdbtypes,.wdb.rdbtypes,.wdb.gatewaytypes,.wdb.tickerplanttypes,.wdb.sorttypes,.wdb.sortworkertypes,.wdb.idbtypes) except `
```

code/wdb/writedown.q
```diff
@@ -2,13 +2,12 @@
 
 /-Required variables for savetables function
 compression:@[value;`compression;()];                                      /-specify the compress level, empty list if no required
-savedir:@[value;`savedir;`:temphdb];                                       /-location to save wdb data
-hdbdir:@[value;`hdbdir;`:hdb];                                             /-move wdb database to different location
+savedir:hsym @[value;`savedir;`:temphdb];                                       /-location to save wdb data
+hdbdir:hsym @[value;`hdbdir;`:hdb];                                             /-move wdb database to different location
 
-hdbsettings:(`compression`hdbdir)!(compression;hdbdir);
+hdbsettings:(`compression`hdbdir)!(compression;hsym hdbdir);
```
jonathonmcmurray, line 8: are these changes necessary?

config/settings/wdb.q
```diff
@@ -58,9 +62,8 @@ tpcheckcycles:0W
 
 // Server connection details
 \d .servers
-CONNECTIONS:`hdb`tickerplant`rdb`gateway`sort                                               // list of connections to make at start up
```
jonathonmcmurray, line 61: why is this removed?

jonathonmcmurray (commented):
A few comments - haven't looked at docs or unit tests yet (think you mentioned docs might be updated further)

code/common/merge.q
```diff
@@ -27,12 +27,18 @@ checkpartitiontype:{[tablename;extrapartitiontype]
...
+                .lg.o[`checksymbol;"parted column does have a symbol type in ",(string tablename)," table"];
+                .lg.e[`checksymbol;"parted column ",string[extrapartitiontype]," does not have a symbol type in ",(string tablename)," table"]];
+        };
 
 
 /- function to get list of distinct combiniations for partition directories
 /- functional select equivalent to: select distinct [ extrapartitiontype ] from [ tablenme ]
 getextrapartitions:{[tablename;extrapartitiontype]
-        value each ?[tablename;();1b;extrapartitiontype!extrapartitiontype]
+        value each ?[tablename;();1b;(!) . 2#enlist extrapartitiontype,()]
```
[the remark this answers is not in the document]
Papamaci444: It was failing when extrapartitiontype was not a list. True the original is way more readable.
Maybe I could add a
extrapartitiontype:extrapartitiontype,();

code/common/merge.q
```diff
@@ -27,12 +27,18 @@ checkpartitiontype:{[tablename;extrapartitiontype]
                 .lg.o[`checkpart;"all parted columns defined in sort.csv are present in ",(string tablename)," table"]];
 	};
 
+/- function to check if the extra partition column has a symbol type
+checksymboltype:{[tablename;extrapartitiontype]
```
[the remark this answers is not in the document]
Papamaci444: In the savetablesbypartenumcol function in wdb.q there is a check being made if the selected column really has a symbol type. Has a similar function to what checkpartitiontype does in merge.q

code/idb/idbstandard.q
Papamaci444: I now changed that completely. The IDB will register itself with the WDB and retrieve the parameters needed. Also the WDB will initialise an empty db each time after EOD.

code/processes/idb.q
Papamaci444: Reworked as well. This will first check a change in sym file size/partition count then do a reload accordingly.

code/processes/idb.q
Papamaci444: Removed the whole, there is no need for symlink in the current version.

code/wdb/writedown.q
Papamaci444: No they are not, the only reason why I added them is that I've bumped into some errors where the code did not use hsym with these variables therefore concatenations were incorrect(.Q.dd, ` sv).
Then I looked at the purpose of these variables and as I see they only used to point to certain directories, then maybe something gets appended to them - then why not have them hsym by default? This is their only purpose.

config/settings/wdb.q
Papamaci444: It was simply just overwritten in processes/wdb.q anyways.
Looks like it created the list of connections based on process types later anyways.
Not sure then if it makes sense to make it as a setting.

code/processes/wdb.q
Papamaci444: Yeah done!

config/settings/wdb.q
Papamaci444: Okay I think I see now why this is there. Not the fact that those types are listed but rather that is makes it possible for the users to add extra stuff here.
Added back!

code/processes/wdb.q
Papamaci444: Yes you might be right! Added back.

code/processes/idb.q
```diff
@@ -0,0 +1,93 @@
...
+    symfilehaschanged[];
+    .lg.o[`load;"sym file has changed, reloading"];
+    @[load;symfilepath; {.lg.e[`load;"failed to load sym file: ",string[symfilepath]," error: ",x];'x}];
+ };
+
+/-force loads IDB
+loadidb:{[]
+    partitioncounthaschanged[];
+    .lg.o[`load;"number of partitions on disk has changed, reloading"];
+    @[system; "l ", 1_string idbdir; {.lg.e[`load;"failed to load IDB: ",string[idbdir]," error: ",x];'x}];
```
jonathonmcmurray, line 19: (same comment as above)

code/processes/idb.q
```diff
@@ -0,0 +1,93 @@
...
+    currentpartition::pt;
+    idbdir::.Q.dd[savedir; currentpartition];
+    loaddb[];
+ };
+
+/- reloads the db. Called by wdb process midday/eod.
+intradayreload:{[]
+    starttime:.proc.ct[];
+    if[symfilehaschanged[];loadsym[]];
+    if[partitioncounthaschanged[];loadidb[]];
```
jonathonmcmurray, line 41: I think there's some redundancy here - it probably doesn't add much time etc., but we call these checks again inside the loadsym/loadidb functions. Any way we can refactor this to only do the checks once?

code/processes/idb.q
```diff
@@ -0,0 +1,93 @@
...
+    writedownmode::params[3];
+    idbdir::.Q.dd[savedir; $[writedownmode~`default;`;currentpartition]];
+    .lg.o[`init;"Current settings: db folder: ",string[idbdir],", sym file: ",string[symfilepath],", writedownmode: ", string writedownmode];
+ };
+
+init:{[]
+    .lg.o[`init; "searching for servers"];
+    .servers.startup[];
+    .lg.o[`init;"getting connection handle to the WDB"];
+    w:first exec w from .servers.getservers[`proctype;`wdb;()!();1b;1b];
```
jonathonmcmurray, line 72: I think you can use `.servers.gethandlebytype` here https://github.com/DataIntellectTech/TorQ/blob/master/code/handlers/trackservers.q#L106

Something like 

```suggestion
    w:.servers.gethandlebytype[`wdb;`any];
```

possibly we should also make the proctype configurable i.e. something like

```suggestion
    w:.servers.gethandlebytype[.idb.wdbtypes;`any];
```

and then set `.idb.wdbtypes` in the standard way for TorQ config variables

jonathonmcmurray (approved):
few minor comments/questions but I think this looks good

code/processes/idb.q
```diff
@@ -0,0 +1,93 @@
...
+
+/-these parameters are only used once their value has been set with values retrieved from the WBD.
+writedownmode:idbdir:savedir:currentpartition:symfilepath:`;
+symsize:partitionsize:0;
+
+/-force loads sym file
+loadsym:{[]
+    symfilehaschanged[];
+    .lg.o[`load;"sym file has changed, reloading"];
+    @[load;symfilepath; {.lg.e[`load;"failed to load sym file: ",string[symfilepath]," error: ",x];'x}];
```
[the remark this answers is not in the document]
Papamaci444: Oh it just remained from a copy-paste from the version when trap error was on. Removing now.

code/processes/wdb.q
```diff
@@ -101,150 +108,173 @@ mergemaxrows:{[tabname] mergenumrows^mergenumtab[tabname]}
...
+savetodisk:{[]
+    savetables[savedir;getpartition[];0b;] each tablelist[];
+    /- we have to let the idbs know of the changes in the wdbhdb. using filldb[] to make sure it is a db with all the tables
+    if[writedownmode in `partbyenum`default;filldb[];notifyidbs[`.idb.intradayreload;enlist();0b]]};
 
-savetodisk:{[] savetables[savedir;getpartition[];0b;] each tablelist[]};
+/- send an intraday reload message to idbs:
+notifyidbs:{[func;params;islogging]
+    ws:exec w from .servers.getservers[`proctype;`idb;()!();1b;0b];
+    if[islogging;.lg.o[`reload;"found ",(string count ws)," idb(s) to trigger function ",string func]];
```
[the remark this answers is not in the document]
Papamaci444: Changed code now to only call this function when writedown has happened(there was a table with rows changed enough that had to be written down)

code/processes/wdb.q
```diff
@@ -273,25 +303,29 @@ endofdaysortdate:{[dir;pt;tablist;hdbsettings]
...
   };
 
-merge:{[dir;pt;tableinfo;mergelimits;hdbsettings;mergemethod]    
+merge:{[dir;pt;tableinfo;mergelimits;hdbsettings;mergemethod;writedownmode]
   setcompression[hdbsettings[`compression]];
   /- get tablename
   tabname:tableinfo[0];
-  /- get list of partition directories for specified table 
-  partdirs:` sv' tabledir,/:k:key tabledir:.Q.par[hsym dir;pt;tabname];
+  /- get list of partition directories for specified table - partbyattr and partbyenum use different folder structure
```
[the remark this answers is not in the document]
Papamaci444: Edited comment now. Basically there are 2 types of folder structure - partbyattr/default use the same while partbyenum is different.

PR #660, opened 2024-09-17
Replicating the schema extract logic for attr_stripped schemas

GGabi:
Essentially a duplicate of: https://github.com/DataIntellectTech/TorQ/pull/659

Adds a test to ensure existing functionality works as intended for the SegmentedTickerplant as that's the only component that uses the attribute stripped schemas.


tests/schemas/insert_data.q
Papamaci444, line 1: BTW I think I haven't seen a single file defined in TorQ that uses underscore notation!

GGabi: Oh, I can change that if it's a style issue?

Papamaci444 (approved):

tests/schemas/insert_data.q
jonathonmcmurray: not super important for a filename in tests, but yeah, we avoid underscores generally

code/common/pubsub.q
```diff
@@ -104,17 +104,22 @@ closesub:{[h]
   delhandlef[;h]each t;
  };
 
+/ Scope value to inside the function so we can iterate over a set of tables without loading
+/ them all into memory at once
+/ Takes a symbol to a global table
+extractschema: {t: value x; $[.Q.qp t; t; 0#t]};
```
jonathonmcmurray, line 110:
```suggestion
extractschema:{t:value x; $[.Q.qp t; t; 0#t]};
```

mostly a personal preference stylistic thing, but we generally don't have spaces after `:` anywhere in torq

jonathonmcmurray (approved):
minor comments, but looks good

code/common/pubsub.q
GGabi: Got it, I'll make the changes, thanks

Papamaci444 (approved):
Thanks for fixing these, approved!

PR #663, opened 2024-10-02
Update wdb.q

jreillyDI:
Added sync flushing to wdb in case of eodwaittime = 0

### DIFF
--- a/code/processes/wdb.q
+++ b/code/processes/wdb.q
@@ -240,10 +240,8 @@ handler:{
 /- notify the gateway that we are done
 flushend:{
     if[not @[value;`.wdb.reloadcomplete;0b];
-     $[eodwaittime>0;
-      @[{neg[x]"";neg[x][]};;()] each key reloadsummary;
-      @[{x"";x[]};;()] each key reloadsummary
-     ]
+     if[eodwaittime>0;
+      @[{neg[x]"";neg[x][]};;()] each key reloadsummary];
      informgateway(`reloadend;`);
      .lg.o[`sort;"end of day sort is now complete"];
      reloadcomplete::1b];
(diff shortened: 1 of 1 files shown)


code/processes/wdb.q
```diff
@@ -256,8 +259,9 @@ doreload:{[pt]
     /-inform gateway of reload start
     informgateway(`reloadstart;`);
     getprocs[;pt] each reloadorder;
-    if[eodwaittime>0;
+    $[eodwaittime>0;
         .timer.one[timeouttime::.proc.cp[]+eodwaittime;(value;".wdb.flushend[]");"release all hdbs and rdbs as timer has expired";0b];
+        .timer.one[.proc.cp[];(value;".wdb.flushend[]");"flush all hdbs and rdbs"; 0b]
```
jonathonmcmurray, line 264: do we need to do this as a timer? can we not just call it directly?

```suggestion
        .wdb.flushend[]
```

code/processes/wdb.q
```diff
@@ -240,7 +240,10 @@ handler:{
 /- notify the gateway that we are done
 flushend:{
     if[not @[value;`.wdb.reloadcomplete;0b];
-     @[{neg[x]"";neg[x][]};;()] each key reloadsummary;
+     $[eodwaittime>0;
+      @[{neg[x]"";neg[x][]};;()] each key reloadsummary;
+      @[{x"";x[]};;()] each key reloadsummary
```
jonathonmcmurray, line 245: is this needed? if we sent sync requests, we shouldn't have to flush the handles?

code/processes/wdb.q
jreillyDI: Apologies, I think I had misunderstood the purpose of having the timer function here

code/processes/wdb.q
jreillyDI: Hadn't thought of that. Apologies.

jonathonmcmurray (approved):
LGTM

PR #667, opened 2024-10-17
IDB intraday reload fix for default writedown mode, fixed row counts

Papamaci444:
The existing IDB implementation had 2 issues:
1) bug: default writedown method incorrectly created db path.
2) row counts were not updating
This change includes the fix for these 2.


Papamaci444:
IDB intraday reload fix, row counts fix

Issue #669, opened 2024-10-23
`.merge.getpartchunks` uses unfiltered `.merge.partsizes` table after filtering


jonathonmcmurray:
https://github.com/DataIntellectTech/TorQ/blob/33cb5db49ba0028aa118e55e8c01408bb8124fb9/code/common/merge.q#L51
- this line references .merge.partsizes which has already been filtered in local variable `t` - the filtered version should be used here as the the unfiltered table contains intraday partitions for tables not currently being merged, so cut positions generated are incorrect and we can get a `'domain` error due to out of range cut positions.

PR #674, opened 2024-10-28
#669 correct .merge.getpartchunks to use the filtered view

JosefBaines:
Use filtered version of partsizes within getpartchunks.


jonathonmcmurray (approved):
LGTM

PR #675, opened 2024-10-28
IDB WDB connection retry 

JosefBaines:
Alter IDB init function to continually attempt to connect to the WDB rather than causing the proc to fail.


jonathonmcmurray (approved):
LGTM

Issue #672, opened 2024-10-23
IDB reload occurs before previous partition deleted in writedownmode:`default


jonathonmcmurray:
For ``writedownmode:`default``, it looks like the IDB rollover reload is happening before the old partition is deleted from wdbhdb, then the partition gets deleted & idb still thinks it's there

PR #676, opened 2024-10-29
#672 ensure idb process is reloaded after wdb savedown completed

JosefBaines:
Also allow for no idb present

### DIFF
--- a/code/processes/wdb.q
+++ b/code/processes/wdb.q
@@ -256,7 +256,7 @@ doreload:{[pt]
     reloadcomplete::0b;
     /-inform gateway of reload start
     informgateway(`reloadstart;`);
-    getprocs[;pt] each reloadorder where reloadorder in .servers.SERVERS[`proctype];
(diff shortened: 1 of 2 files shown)


code/processes/wdb.q
```diff
@@ -256,7 +256,7 @@ doreload:{[pt]
     reloadcomplete::0b;
     /-inform gateway of reload start
     informgateway(`reloadstart;`);
-    getprocs[;pt] each reloadorder;
+    getprocs[;pt] each reloadorder where reloadorder in .servers.SERVERS[`proctype];
```
jonathonmcmurray, line 259:
```suggestion
    getprocs[;pt] each reloadorder inter .servers.SERVERS[`proctype];
```

🏌️

jonathonmcmurray (approved):
looks good

code/processes/wdb.q
JosefBaines: Sorry didn't notice this - will add here!

Issue #671, opened 2024-10-23
`.merge.mergebypart` error trap references variables not passed to it


jonathonmcmurray:
https://github.com/DataIntellectTech/TorQ/blob/33cb5db49ba0028aa118e55e8c01408bb8124fb9/code/common/merge.q#L75

This error trap lambda references `dest` and `chunks` but these aren't passed in, so it fails if called. also doesn't output the actual error, which seems unhelpful

PR #677, opened 2024-10-30
#671 correct .merge.mergebypart error trap on upsert

JosefBaines:
Also added in returning the specific error - just as a bonus

### DIFF
--- a/code/common/merge.q
+++ b/code/common/merge.q
@@ -72,7 +72,7 @@ mergebypart:{[tablename;dest;partchunks]
    .lg.o[`merge;"upserting ",(string count chunks)," rows to ",string dest];
    /-merge columns to permanent storage
    .[upsert;(dest;chunks);                     
-     {.lg.e[`merge;"failed to merge to ", string[dest], " from segments ", (", " sv string chunks)];}];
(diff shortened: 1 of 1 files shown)


code/common/merge.q
```diff
@@ -72,7 +72,7 @@ mergebypart:{[tablename;dest;partchunks]
    .lg.o[`merge;"upserting ",(string count chunks)," rows to ",string dest];
    /-merge columns to permanent storage
    .[upsert;(dest;chunks);                     
-     {.lg.e[`merge;"failed to merge to ", string[dest], " from segments ", (", " sv string chunks)];}];
+     {[e;d;p].lg.e[`merge;raze "failed to merge to ", string[d], " from segments ", (", " sv string p), " Error is - ",string[]e]}[;dest;partchunks]];
```
jonathonmcmurray, line 75: do we need the `raze`? if we're constructing a single string should be needed I don't think?

JosefBaines: hmm fair point - I _think_ this was due to the partchunks - let me review here

JosefBaines: string[]e

^ epic fail on this as well

jonathonmcmurray (approved):

PR #678, opened 2024-11-03
Add kdbtplog to segmentedchainedtickerplant

erichards97:
`kdbtplog` was recently added to `settings/segmentedtickerplant.q`

If you have a `segmentedchainedtickerplant` process that does not inherit `segmentedtickerplant` as a parentproctype, it will not pick up this config, and the process will fail to startup as `kdbtplog` is undefined.

Therefore adding a copy to `settings/segmentedchainedtickerplant.q`. This brings it in line with the rest of the ctp config, where it has its own copy of each of the `.stplg` variables.

Thanks

### DIFF
--- a/config/settings/segmentedchainedtickerplant.q
+++ b/config/settings/segmentedchainedtickerplant.q
@@ -21,7 +21,8 @@ multilogperiod:0D01;
 errmode:1b;
 batchmode:`defaultbatch;                          // [autobatch|defaultbatch|immediate]
 customcsv:hsym first .proc.getconfigfile["stpcustom.csv"];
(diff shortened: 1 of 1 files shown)


jonathonmcmurray:
Hi @erichards97, thanks for this

We would generally recommend having the segmentedtickerplant as the parentproctype of segmentedchainedtickerplant - I would expect there could be other breakages by not having this set (for example, most of the code for the segementedtickerplant is in `code/segmentedtickerplant/` and would have to be explicitly loaded if not using the parentproctype)

Is there a particular use case where you need to set the parentproctype differently?

PR #681, opened 2024-11-21
#672 make sort process call initmissingtables and notifyidbs instead …

ThomasGillespie:
Fix bug where idb rollover fails to find new partition

### DIFF
--- a/code/processes/wdb.q
+++ b/code/processes/wdb.q
@@ -307,7 +307,7 @@ endofdaysortdate:{[dir;pt;tablist;hdbsettings]
   .lg.o[`mvtohdb;"Moving partition from the temp wdb ",(dw:.os.pth -1 _ string .Q.par[dir;pt;`])," directory to the hdb directory ",hw:.os.pth -1 _ string .Q.par[hdbsettings[`hdbdir];pt;`]];
   .lg.o[`mvtohdb;"Attempting to move ",(", "sv string key hsym`$dw)," from ",dw," to ",hw];
   .[movetohdb;(dw;hw;pt);{.lg.e[`mvtohdb;"Function movetohdb failed with error: ",x]}];
-  
+
   /-call the posteod function
   .save.postreplay[hdbsettings[`hdbdir];pt];
   if[permitreload;
@@ -419,7 +419,7 @@ endofdaysort:{[dir;pt;tablist;writedownmode;mergelimits;hdbsettings;mergemethod]
         endofdaysortdate[dir;pt;key tablist;hdbsettings]
     ];
     /- run steps to rollover idb
-    idbreload[pt+1];
+    idbreload[currentpartition+1];
     /- reset compression level (.z.zd)
     resetcompression[16 0 0]
     };
@@ -523,9 +523,8 @@ fixpartition:{[subto]
 
 /- for writedown modes partbyenum/default we make sure that partition 0/currentpartition has all the tables.
 /- In that case we can use .Q.chk later to fill the db making it useable for intraday processes
-/- currentpt - boolean; if true initialises partition for value of currentpartition; if false initialises partition for next value of currentpartition
-initmissingtables:{[currentpt]
-    $[currentpt;pt:currentpartition;pt:currentpartition+1];
+/- pt - date; partition for which the function should initialise
+initmissingtables:{[pt]
     .lg.o[`fixpartition;"Adding missing tables(empty) to partition ",string pt];
     inittable[;pt] each tablelist[];
     filldb[pt];
@@ -597,7 +596,7 @@ idbreload:{[pt]
     .lg.o[`idb;"starting idb reload"];
     if[writedownmode in `partbyenum`default;
         .lg.o[`eod;"initialising wdbhdb for partition: ",string[pt]];
-        $[.proc.proctype~`sort;{ws:exec w from .servers.getservers[`proctype;`wdb;()!();1b;0b];[first ws](`.wdb.initmissingtables;[1b])}[];initmissingtables[0b]];
+        $[.proc.proctype~`sort;{[pt]ws:exec w from .servers.getservers[`proctype;`wdb;()!();1b;0b];[first ws](`.wdb.initmissingtables;[pt])}[pt];initmissingtables[pt]];
         .lg.o[`eod;"notifying idbs for newly created partition"];
         notifyidbs[`.idb.rollover;pt]];
     .lg.o[`idb;"idb reload complete"];
--- a/code/processes/wdb.q
+++ b/code/processes/wdb.q
@@ -524,20 +524,21 @@ fixpartition:{[subto]
 
 /- for writedown modes partbyenum/default we make sure that partition 0/currentpartition has all the tables.
 /- In that case we can use .Q.chk later to fill the db making it useable for intraday processes
-initmissingtables:{[]
-    .lg.o[`fixpartition;"Adding missing tables(empty) to partition ",string currentpartition];
-    inittable each tablelist[];
-    filldb[];
+initmissingtables:{[currentPt]
+    $[currentPt;pt:currentpartition;pt:currentpartition+1];
+    .lg.o[`fixpartition;"Adding missing tables(empty) to partition ",string pt];
+    inittable[;pt] each tablelist[];
+    filldb[pt];
  }
 
-filldb:{[]
+filldb:{[pt]
     /- for all enumerated partitions we want to make sure that all tables are present
-    .Q.chk[.Q.par[hsym savedir; currentpartition; `]];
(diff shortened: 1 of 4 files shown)


code/processes/wdb.q
```diff
@@ -482,7 +479,7 @@ informsortandreload:{[dir;pt;tablist;writedownmode;mergelimits;hdbsettings;merge
         ];
         [.lg.e[`informsortandreload;"can't connect to the sortandreload - no sortandreload process detected"];
          // try to run the sort locally
-         endofdaysort[dir;pt;tablist;writedownmode;mergelimits;hdbsettings;mergemethod]]];
+         endofdaysort[dir;pt;tablist;writedownmode;mergelimits;hdbsettings;mergemethod;]]];
```
jonathonmcmurray, line 482: is this change intentional? I think this will break this function call?

code/processes/wdb.q
```diff
@@ -313,7 +307,10 @@ endofdaysortdate:{[dir;pt;tablist;hdbsettings]
   .lg.o[`mvtohdb;"Moving partition from the temp wdb ",(dw:.os.pth -1 _ string .Q.par[dir;pt;`])," directory to the hdb directory ",hw:.os.pth -1 _ string .Q.par[hdbsettings[`hdbdir];pt;`]];
   .lg.o[`mvtohdb;"Attempting to move ",(", "sv string key hsym`$dw)," from ",dw," to ",hw];
   .[movetohdb;(dw;hw;pt);{.lg.e[`mvtohdb;"Function movetohdb failed with error: ",x]}];
-
+  
+  
+  idbReload[pt];
+
```
jonathonmcmurray, line 313:
```suggestion

  idbReload[pt];

```

(remove extra blank line and whitespace)

code/processes/wdb.q
```diff
@@ -527,20 +524,21 @@ fixpartition:{[subto]
 
 /- for writedown modes partbyenum/default we make sure that partition 0/currentpartition has all the tables.
 /- In that case we can use .Q.chk later to fill the db making it useable for intraday processes
-initmissingtables:{[]
-    .lg.o[`fixpartition;"Adding missing tables(empty) to partition ",string currentpartition];
-    inittable each tablelist[];
-    filldb[];
+initmissingtables:{[currentPt]
```
jonathonmcmurray, line 527: we don't use camel case in TorQ

```suggestion
initmissingtables:{[currentpt]
```

also can you add a comment about what this argument is for? not immediately clear

code/processes/wdb.q
```diff
@@ -594,6 +592,16 @@ getsortparams:{[]
     ];
     };
 
+idbReload:{[pt]
+    .lg.o[`idb;"starting idb reload"];
+    if[writedownmode in `partbyenum`default;
+        .lg.o[`eod;"initialising wdbhdb for partition: ",string[currentpartition]];
+        $[.proc.proctype~`sort;{ws:exec w from .servers.getservers[`proctype;`wdb;()!();1b;0b];[first ws](`.wdb.initmissingtables;[1b])}[];initmissingtables[0b]];
```
jonathonmcmurray, line 599: what about sortworker processes? do they need to be handled?

why is the arg `1b` if running it from sort process but `0b` if running directly on wdb?

code/processes/wdb.q
```diff
@@ -594,6 +592,16 @@ getsortparams:{[]
     ];
     };
 
+idbReload:{[pt]
```
jonathonmcmurray, line 595:
```suggestion
idbreload:{[pt]
```

code/processes/wdb.q
ThomasGillespie: I don't believe the sortworker procs will ever be involved here but I could be wrong


When the initmissingtables function was called from the wdb with the new set up, it was called before currentpartition is updated to the next day

The new argument for the function allows the function to be called for either the current value of currentpartition or the next one.

Honestly not too sure on this, so I am open to suggestions

jonathonmcmurray: > When the initmissingtables function was called from the wdb with the new set up, it was called before currentpartition is updated to the next day

Sorry I missed this comment earlier, I'm not really getting it. Why is it called before currentpartition is updated if called from wdb but after if called from sort? is there any chance of a timing issue here where if sort is very fast it could call it before it gets updated etc.?

this feels quite messy to me

ThomasGillespie: It seems to me that if the sorting is done by the sort proc, then the wdb has already moved on and updated currentpartiton, however when the wdb does the sort currentpartition hasn't been updated as the endofday method is blocked

code/processes/wdb.q
```diff
@@ -527,20 +523,22 @@ fixpartition:{[subto]
 
 /- for writedown modes partbyenum/default we make sure that partition 0/currentpartition has all the tables.
 /- In that case we can use .Q.chk later to fill the db making it useable for intraday processes
-initmissingtables:{[]
-    .lg.o[`fixpartition;"Adding missing tables(empty) to partition ",string currentpartition];
-    inittable each tablelist[];
-    filldb[];
+/- currentpt - boolean; if true initialises partition for value of currentpartition; if false initialises partition for next value of currentpartition
```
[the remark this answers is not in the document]
ThomasGillespie: same as I have said above, basically if currentpartition hasn't been updated yet but we still need to initialise the wdbhdb partition for the next value of current partition we need to make this function have the ability to do it

ThomasGillespie:
There might be a better way to to the initmissingtables to make sure it is right, going to look into this and try to come up with something better

code/processes/wdb.q
jonathonmcmurray: so should the initilisation of next partition not be tied to wdb moving to next partition, rather than being tied to idb reload?

jonathonmcmurray (commented):
Couple of little things, but looking good

code/processes/wdb.q
```diff
@@ -594,6 +591,17 @@ getsortparams:{[]
     ];
     };
 
+/- funtion to initialise partition at EOD and notifyidbs to rollover, if .wdb.mode is not a sort mode, initmissingtables done on sort proc
```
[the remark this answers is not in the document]
ThomasGillespie: https://github.com/DataIntellectTech/TorQ/pull/681/files#diff-6a3cbc5795f37aeb51ed652b0a6abb04020354619518f84a6c8423d0f888ca55R599
it's this line, if the proc type the function is running on is sort then the wdb mode is not a sort mode

code/processes/wdb.q
```diff
@@ -594,6 +591,17 @@ getsortparams:{[]
...
+idbreload:{[pt]
+    .lg.o[`idb;"starting idb reload"];
+    if[writedownmode in `partbyenum`default;
+        .lg.o[`eod;"initialising wdbhdb for partition: ",string[pt]];
+        $[.proc.proctype~`sort;{[pt]ws:exec w from .servers.getservers[`proctype;`wdb;()!();1b;0b];[first ws](`.wdb.initmissingtables;[pt])}[pt];initmissingtables[pt]];
```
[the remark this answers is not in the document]
ThomasGillespie: with multiple sort processes it does run on both, but it is OK.
with multiple wdb processes it only sends the command to one, but that is all that is needed for the initmissingtables

code/processes/wdb.q
```diff
@@ -75,7 +77,7 @@ eodwaittime:@[value;`eodwaittime;0D00:00:10.000];                          /-len
 .save.savedownmanipulation:@[value;`.save.savedownmanipulation;()!()];           /-a dict of table!function used to manipulate tables at EOD save
 .save.postreplay:@[value;`.save.postreplay;{{[d;p] }}];                          /-post EOD function, invoked after all the tables have been written down
 
-/ - end of default parameters
+ - end of default parameters
```
jonathonmcmurray, line 80: this shouldn't have been removed

code/processes/wdb.q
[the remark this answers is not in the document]
jonathonmcmurray: ok, but my point is this comment refers to `.wdb.mode`, and we're not checking a variable by that name?

code/processes/wdb.q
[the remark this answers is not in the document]
jonathonmcmurray: > with multiple wdb processes it only sends the command to one, but that is all that is needed for the initmissingtables

this seems like an assumption based on the assumption that all wdbs will be subscribed to & saving the same tables? if we have e.g. two wdb proctypes (lets say `wdbtrade` and `wdbquote`) and each proctype subscribes & saves one table, do we not need to initmissingtables on both?

jonathonmcmurray (approved):

Issue #684, opened 2024-12-04
Add a warning message if environment variable is not set when reading torq process file


robsketch:
If you reference an environment variable which is not set in the process.csv, there is no indication from the logs that this is the case, which could have helped when debugging the below issue.

The issue being in this case, if you set a port number to be something like {KDBBASEPORT}+n and KDBBASEPORT is not set, then the port number will be empty and eventually parsed to a null integer in the underlying kdb table read via torq.q.

Consequently, the log message is, in my opinion, unclear that this would have been the problem given my start line and the resulting log message:

`q /home/rsketch/deploy/TorQ/latest/torq.q -proctype randomproc -procname randomproc1 -U /home/rsketch/deploy/TorQApp/latest/appconfig/passwords/accesslist.txt -load /home/rsketch/deploy/TorQApp/latest/code/processes/randomproc.q  -procfile /home/rsketch/deploy/TorQApp/latest/appconfig/process.csv -debug
`

`2024.12.04D15:17:43.030222829|homer.aquaq.co.uk|torq|/home/rsketch/deploy/TorQ/latest/torq.q_2196535_0|ERR|readprocfile|No port passed via -p flag or found in /home/rsketch/deploy/TorQApp/latest/appconfig/process.csv. Parameters are host: localhost, proctype: randomproc, procname: randomproc1`

Proposed change is to add additional logging in the function which parses the environment variables to warn if a given environment variable is set or not, and to hopefully make the log message from torq.q a bit more clear as well.

PR #685, opened 2024-12-04
Add warning if environment variable is not set and improve logging

robsketch:
Attempts to address issue #684


jonathonmcmurray (approved):
Looks good - can we also remove the version from `code/common/` to avoid any future confusion please?

PR #686, opened 2024-12-04
fix idbreload issues

michaelturkington:
Think I've understood the context here
* wdb.q change simple enough, just using the local instead of the global variable
* idb just cleaned up clients suggested conditional, removed their "1b" option to conform with original logic of the function.

### DIFF
--- a/code/processes/idb.q
+++ b/code/processes/idb.q
@@ -56,7 +56,7 @@ symfilehaschanged:{[]
 /- checks if count of partitions has changed since last reload of the IDB. Records new partition count if changed.
 /- the default writedown method doesn't need db reloading as no new directory is being created there.
 partitioncounthaschanged:{[]
-    if[writedownmode~`default;:0b];
+    if[(1j~partitionsize)&writedownmode~`default;:0b];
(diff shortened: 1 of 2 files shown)


code/processes/idb.q
```diff
@@ -56,7 +56,7 @@ symfilehaschanged:{[]
 /- checks if count of partitions has changed since last reload of the IDB. Records new partition count if changed.
 /- the default writedown method doesn't need db reloading as no new directory is being created there.
 partitioncounthaschanged:{[]
-    if[writedownmode~`default;:0b];
+    if[(1j~partitionsize)&writedownmode~`default;:0b];
```
jonathonmcmurray, line 59: I think we need some comment to explain why this additional condition is here. Need to make sure this is the right logic to be using

Also partitionsize is being updated in the line below, do we need to re-order things or is this the correct order?

jonathonmcmurray: as above, can we add a comment about this check? what does it mean if partitionsize is 1 when this function is called?

jonathonmcmurray (approved):

Issue #690, opened 2025-01-24
WDB partbyenum misses null symbols


jonathonmcmurray:
When using WDB in `partbyenum` mode, we are using the integer enumeration values of symbols for the partition names. By default, null symbols will be enumerated to `0N` and therefore we end up saving these outside of any integer partition, and then they don't get saved to the HDB at the end of the day.

Bearing in mind that the IDB needs to be able to read the int partitioned db written by WDB, I don't think there is a way in kdb to have an int partition with value of `0N`.

I think the way we can fix this is to replace null symbols with some placeholder e.g. `` `NONE`` before enumerating/writing, then setting it back to null during EOD merge.

We'd also need to handle this in the `maptoint` function in the IDB, so that users are still able to query the null symbol.

PR #691, opened 2025-01-29
Fix partbyenum to enumerate against placeholding symbol

ConorSwainDI:
PR to resolve https://github.com/DataIntellectTech/TorQ/issues/690

### DIFF
--- a/code/processes/idb.q
+++ b/code/processes/idb.q
@@ -107,5 +107,5 @@ reload:.idb.intradayreload;
 
 /- helper function to support queries against the sym column
 maptoint:{[symbol]
-    sym?symbol
+    sym?`NONE^symbol
  };
--- a/code/processes/idb.q
+++ b/code/processes/idb.q
@@ -107,5 +107,5 @@ reload:.idb.intradayreload;
 
 /- helper function to support queries against the sym column
 maptoint:{[symbol]
-    sym?`NONE^symbol
+    sym?`TORQNULLSYMBOL^symbol
  };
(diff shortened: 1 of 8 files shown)


code/processes/idb.q
```diff
@@ -107,5 +107,5 @@ reload:.idb.intradayreload;
 
 /- helper function to support queries against the sym column
 maptoint:{[symbol]
-    sym?symbol
+    sym?`NONE^symbol
```
jonnypress, line 110: Could we use something different here - NONE seems like it could validly be in an instrument universe
https://finance.yahoo.com/quote/NONE-USD/
Can we do something like `TORQNULLSYMBOL or something

code/processes/wdb.q
```diff
@@ -114,16 +114,20 @@ mergemaxrows:{[tabname] mergenumrows^mergenumtab[tabname]}
 tablelist:{[] sortedlist:exec tablename from `bytes xdesc tabsizes;
     (sortedlist union tables[`.]) except ignorelist}
 
+/- function to check for null syms
+checknullsym:{[s] -1 _ `${@[x; where not ((type each x) in (10 -10h));string]} s,(::)}
```
jonnypress, line 118: What is the purpose of this function?

ConorSwainDI: This was existing logic to check whether a sym was null (old wdb code line 126). I wanted leverage it in line 124, and didn't want any duplicate code so moved it into its own function

code/processes/idb.q
ConorSwainDI: Yes I am happy to fix this to use some explicitly stated `TORQNULLSYMBOL or similar. Existing code moved null symbols to NONE

jonathonmcmurray (approved):
One naming issue, otherwise looks good

Issue #699, opened 2025-03-24
Default .replay.tmpdir to be an hsym


jonathonmcmurray:
In tickerlogreplay.q, we default `hdbdir` to be an hsym: https://github.com/DataIntellectTech/TorQ/blob/master/code/processes/tickerlogreplay.q#L109

We should also do the same for `tmpdir` (defined here: https://github.com/DataIntellectTech/TorQ/blob/master/code/processes/tickerlogreplay.q#L30)

We need to make sure this happens after any user defined value is set (e.g. via config or command line), similar to what we're doing for `hdbdir`

PR #709, opened 2025-06-17
Fix: default tempdir to be an hsym in tickerlogreplay.q

rokhayagueye98:
This PR updates tickerlogreplay.q to ensure tempdir and hdbdir to be an hsym as related to issue #699


code/processes/tickerlogreplay.q
```diff
@@ -27,7 +27,7 @@ clean:@[value;`clean;1b]                                                // clean
 sortcsv:@[value;`sortcsv;hsym first .proc.getconfigfile["sort.csv"]]    // location of  sort csv file
 compression:@[value;`compression;()];                                   // specify the compress level, empty list if no required
 partandmerge:@[value;`partandmerge;0b];                                 // setting to do a replay where the data is partitioned and then merged on disk
-tempdir:@[value;`tempdir;`:tempmergedir];                               // location to save data for partandmerge replay
+tempdir:hsym @[value;`tempdir;`:tempmergedir];                               // location to save data for partandmerge replay
```
jonathonmcmurray, line 30: I don't think we need to do this here seeing as we're doing it below - but if we are leaving it here as well, need to re-align the comment

PR #734, opened 2026-04-16
Fix #733: use PowerShell Move-Item for cross-volume directory on Windows

### LINKED ISSUE #733: `os.ren` fails intradisk move on Windows
On Windows, TorQ's built-in `.os.ren` implementation does not support moving a directory across volumes/drives.
 
This becomes a problem when WDB is configured with `default` writedown mode and the WDB location and HDB location are on different disks. In that mode, WDB needs to move the whole partition directory to HDB, but the current Windows rename/move behavior fails for that workflow.
 
## Environment
- TorQ version: [v5.2.14]
- OS: Windows Server [Windows Server 2022 Standard, 21H2]
- kdb+ version: [4.1]
- WDB writedown mode: `default`
- WDB path: on one Windows drive/volume
- HDB path: on a different Windows drive/volume
 
## Problem
TorQ already supports WDB `default` writedown mode and has had recent Windows-specific fixes around WDB/IDB end-of-day handling. However, when WDB and HDB partitions are located on different disks, the built-in `os.q` move/rename utility does not handle the required directory move on Windows.
 
The result is that `default` writedown mode is not usable in this deployment layout, even though it is a valid and important configuration. This is especially relevant because `default` writedown mode can be materially faster than `partbyenum` under higher write load.
 
## Steps To Reproduce
1. Set up TorQ on Windows.
2. Configure WDB to use `default` writedown mode.
3. Place the WDB root on one disk/volume, for example `D:`.
4. Place the HDB root on a different disk/volume, for example `E:`.
5. Run the normal WDB save/end-of-day flow that moves the written partition directory from WDB to HDB.
 
## Actual Behaviour
The directory move fails on Windows when source and destination are on different drives/volumes, at EOD.
 
## Expected Behaviour
TorQ should support this workflow on Windows by using a directory move/copy strategy that works across volumes, so WDB `default` writedown mode can complete successfully when WDB and HDB are on different disks.
 
## Suggested Fix
Update the Windows implementation of `.os.ren` in `code/common/os.q` so that directory moves across volumes are supported.
 
For example, the Windows path could detect when the source is a directory and use a Windows-compatible cross-volume directory move approach rather than relying on the current UNIX-style rename/move behavior.

khurram-saeed-malik:
Fixes DataIntellectTech/TorQ#733
## Problem
On Windows, `.os.ren` used a `move` command that fails when the source
and destination are on different drives/volumes. This breaks WDB default
writedown mode when the WDB path and HDB path reside on different disks.
## Solution
Replace the old Windows `move`-based implementation with PowerShell's
`Move-Item` cmdlet, which natively handles:
- File renames/moves
- Directory renames/moves
- Cross-volume moves (different Windows drives/disks)
The Unix path (`mv`) is unchanged.
## Files Changed
- `code/common/os.q` - updated `.os.ren`

## Contributor
Arseni Maiorov
https://github.com/mayoroff

### DIFF
--- a/code/common/os.q
+++ b/code/common/os.q
@@ -14,7 +14,6 @@ hdeldir:{[dirpath;pdir]
  .lg.o[`deldir;"deleting from  directory : ",dirpath];
  hdel each desc filelist}
 md:{if[not Fex x;system"mkdir \"",pth[x],"\""]};
-ren:{system("mv ";"move ")[NT],pth[x]," ",pth y}
 cpy:{system("cp ";"copy ")[NT],pth[x]," ",pth y}
 Vex:not 0h~type key`.@
 df:{(`$("/";"\\")[NT]sv -1_v;`$-1#v:("/";"\\")[NT]vs pth(string x;x)[10h=type x])} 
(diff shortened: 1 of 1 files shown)


jonathonmcmurray (approved):

jonathonmcmurray:
thanks @khurram-saeed-malik, merged

PR #735, opened 2026-04-20
Updated .Q.lim to give the key-values for just limit values

gordon-luo-13:
**Issue**
In the latest KDB-X release, KDB-X 5.0.20260410, `.Q.lim[]` returns a dictionary with the current and limit values - this will cause a type error.

**Suggested Fix**
Change check for limit on process connections
```$[`lim in key`.Q;$[0W=.Q.lim[][`conns][`lim];1b;0b];1b] ```


jonathonmcmurray:
thanks for raising, we have merged #738 achieving same goal

PR #740, opened 2026-05-20
Add anymaputils

bsofjan:
Add anymaputils

### DIFF
--- a/code/common/anymaputils.q
+++ b/code/common/anymaputils.q
@@ -1,0 +1,5 @@
+.anymap.writeToAnyMap:{[data;filePath] (hsym filePath) 1: data; :(::)}
+.anymap.deriveAnymapFiles:{[filePath] strPath: string[filePath]; `$(strPath;strPath,"#";strPath,"##")}
+.anymap.util.copyAnyMap:{[fromFilePath; destFilePath] fromFilePaths:.anymap.deriveAnymapFiles[fromFilePath]; system (("cp "," " sv string fromFilePaths), " ", string destFilePath)}
+.anymap.util.removeAnyMap:{[filePath] filePaths:.anymap.deriveAnymapFiles[filePath]; system ("rm "," " sv string filePaths)}
+.anymap.util.moveAnyMap:{[fromFilePath; destFilePath] fromFilePaths:.anymap.deriveAnymapFiles[fromFilePath]; system (("mv "," " sv string fromFilePaths), " ", string destFilePath)}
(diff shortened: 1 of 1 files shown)


CJA18 (commented):
This PR looks good to me - only consideration would be to switch the namespace names to .util.anymap.XXX to indicate that it is a util function (we use this naming convention on my client), but this is only if it makes sense in the context of the wider codebase (I haven't checked I will leave that to you @bsofjan)

code/common/anymaputils.q
jonathonmcmurray, line 1: lots of camelCase here, please remove

code/common/anymaputils.q
```diff
@@ -0,0 +1,5 @@
+.anymap.writeToAnyMap:{[data;filePath] (hsym filePath) 1: data; :(::)}
+.anymap.deriveAnymapFiles:{[filePath] strPath: string[filePath]; `$(strPath;strPath,"#";strPath,"##")}
```
jonathonmcmurray:
```suggestion
.anymap.deriveAnymapFiles:{[filePath] `$string[filePath],/:("";"#";"##")}
```

think something like this should work the same

code/common/anymaputils.q
```diff
@@ -0,0 +1,5 @@
+.anymap.writeToAnyMap:{[data;filePath] (hsym filePath) 1: data; :(::)}
```
jonathonmcmurray: most data writing functions in q have filepath as first param & data as second (e.g. `set`, `upsert` etc.). Seems unintuitive to have this the other way round.

code/common/anymaputils.q
```diff
@@ -0,0 +1,5 @@
...
+.anymap.deriveAnymapFiles:{[filePath] strPath: string[filePath]; `$(strPath;strPath,"#";strPath,"##")}
+.anymap.util.copyAnyMap:{[fromFilePath; destFilePath] fromFilePaths:.anymap.deriveAnymapFiles[fromFilePath]; system (("cp "," " sv string fromFilePaths), " ", string destFilePath)}
+.anymap.util.removeAnyMap:{[filePath] filePaths:.anymap.deriveAnymapFiles[filePath]; system ("rm "," " sv string filePaths)}
+.anymap.util.moveAnyMap:{[fromFilePath; destFilePath] fromFilePaths:.anymap.deriveAnymapFiles[fromFilePath]; system (("mv "," " sv string fromFilePaths), " ", string destFilePath)}
```
jonathonmcmurray: these copy/remove/move functions rely on system commands and will only work on Linux. Should use the functions in `.os` namespace of torq so that they work cross-platform.

code/common/anymaputils.q
bsofjan: Got it!

code/common/anymaputils.q
bsofjan: Agreed

code/common/anymaputils.q
bsofjan: Okay I can change this

code/common/anymaputils.q
bsofjan: Okay

bsofjan:
Addressed comments made through new commit.

bsofjan:
Hey @CJA18, thank you for your suggestions on naming the namespace! I am following the names given in the Anycache project proposal, although I am open for discussions to rename it.

bsofjan:
Test Evidence for .anymap.writetoanymap:
image

Test Evidence for .anymap.deriveanymapfiles:
image

PR #743, opened 2026-05-29
Add cacheConfig, setter process, getter script

bsofjan:
Per Phase 1.3 Setter Process and Phase 1.4 Getter Bootstrap of the Anycache Project Proposal, in this PR I am adding:

1. Setter process
2. Getter bootstrap
3. cacheconfig json

The goal of Phase 1.3 and Phase 1.4 is to put together the bare bones of a "setter" process and "getter" functionality.

### DIFF
--- a/code/common/getter.q
+++ b/code/common/getter.q
@@ -1,0 +1,3 @@
+\d .getter
+
+getCacheFromDisk:{[filePath] load hsym filePath}
--- a/code/common/getter.q
+++ b/code/common/getter.q
@@ -1,3 +1,8 @@
+// Bootstrap script that will enable an existing process to load in cache getter functionalities.
+
 \d .getter
 
-getCacheFromDisk:{[filePath] load hsym filePath}
+// Get cache from disk
+getcachefromdisk:{[filePath] load hsym filePath}
+
+\d .
--- a/code/common/getter.q
+++ b/code/common/getter.q
@@ -1,8 +1,0 @@
-// Bootstrap script that will enable an existing process to load in cache getter functionalities.
-
-\d .getter
-
-// Get cache from disk
-getcachefromdisk:{[filePath] load hsym filePath}
-
-\d .
--- a/code/processes/setter.q
+++ b/code/processes/setter.q
@@ -1,6 +1,6 @@
 // Setter process to set cache to disk
 
-\d .setter
+\d .anycache.setter
 
 // Save cache down to disk
 savecachedowntodisk:{[data;filePath] (hsym filePath) set data };
--- a/code/processes/setter.q
+++ b/code/processes/setter.q
@@ -1,7 +1,12 @@
+// Setter process to set cache to disk
+
 \d .setter
 
-saveCacheDowntoDisk:{[data;filePath] (hsym filePath) set data };
-cacheConfigLocation:.proc.getconfigfile["cacheConfig.json"];		/ - location of the cache config configuration file
-cacheConfig:.j.k raze read0 hsym first cacheConfigLocation;
+// Save cache down to disk
+savecachedowntodisk:{[data;filePath] (hsym filePath) set data };
+
+// Get location of cache config and load it in.
+cacheconfiglocation:.proc.getconfigfile["cacheconfig.json"];
+cacheconfig:.j.k raze read0 hsym first cacheconfiglocation;
 
 \d .
--- a/code/processes/setter.q
+++ b/code/processes/setter.q
@@ -1,0 +1,7 @@
+\d .setter
+
+saveCacheDowntoDisk:{[data;filePath] (hsym filePath) set data };
+cacheConfigLocation:.proc.getconfigfile["cacheConfig.json"];		/ - location of the cache config configuration file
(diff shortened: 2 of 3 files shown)


CJA18:
A few general things you need to do when committing to TorQ:
- The PR needs a more detailed description outlining the reason for it, what's being achieved, future steps to come etc
- A lot of these functions seem to be something common that has likely been done already and might have an existing util function, always check for this first as it's best practice to reuse the existing utils (if something needs to be changed then it only needs to be changed in one place)
- Follow coding standards, if you look at the rest of the codebase camelCase isn't used so follow whatever standard is used in the codebase
- For comments, in the rest of the codebase you'll see that the comments are generally block comments for each function describing what the function and each step in the function is doing, and inline comments are used in special cases (check the codebase to see when to use them)
- You committed an empty json config file, it's better to commit a file with some defaults so a reviewer can get an idea of what it will be used for

code/common/getter.q
```diff
@@ -0,0 +1,3 @@
+\d .getter
+
+getCacheFromDisk:{[filePath] load hsym filePath}
```
CJA18: escape the getter namespace at the end of this file

bsofjan: Resolved.

bsofjan:
I have reviewed the TorQ codebase and could not find an existing utility function for the setter.q and getter.q functions. There are existing save and load utilities in the TorQ codebase, however they are highly specific (e.g. loading q and k scripts, saving partitions) and not suitable for use for saving and loading with the anymap data format.

I have also addressed all the other points through my latest commit "Standardising Changes and Add Default json" and by adding more detailed description to the PR.

CJA18:
@bsofjan thanks for checking if the utility was present in TorQ and for addressing my other comments - looks good to merge

bsofjan:
@CJA18 Thank you - I think it is best to wait prior to merging as the PRs functions as checkpoints of work, and merging can be considered once the whole project comes around.

CJA18 (approved):

code/common/getter.q
```diff
@@ -0,0 +1,8 @@
+// Bootstrap script that will enable an existing process to load in cache getter functionalities.
+
+\d .getter
+
+// Get cache from disk
+getcachefromdisk:{[filePath] load hsym filePath}
```
jonathonmcmurray, line 6: filePath -> filepath

this also seems like a very thin wrapper around `load`, is it really useful to have this?

code/processes/setter.q
```diff
@@ -0,0 +1,12 @@
+// Setter process to set cache to disk
+
+\d .setter
+
+// Save cache down to disk
+savecachedowntodisk:{[data;filePath] (hsym filePath) set data };
```
jonathonmcmurray, line 6: filePath -> filepath

and similar to the getter function, very thin wrapper around `set`, is it useful?

code/processes/setter.q
```diff
@@ -0,0 +1,12 @@
...
+
+// Get location of cache config and load it in.
+cacheconfiglocation:.proc.getconfigfile["cacheconfig.json"];
+cacheconfig:.j.k raze read0 hsym first cacheconfiglocation;
```
jonathonmcmurray, line 10: I think `.proc.getconfigfile` should only return one file, do we need the `first` here?

code/processes/setter.q
```diff
@@ -0,0 +1,12 @@
+// Setter process to set cache to disk
+
+\d .setter
```
jonathonmcmurray: I think this namespace is unclear, without having additional context it isn't apparent what "setter" is or is for

code/processes/setter.q
```diff
@@ -0,0 +1,12 @@
+// Setter process to set cache to disk
```
jonathonmcmurray, line 1: should this be a standalone process? seems like caching would be something that multiple processes would do so perhaps this should be a lib rather than a process?

code/common/getter.q
bsofjan: Hello Jonathon, I understand your comments, just to give you more context, I'm following the work development phases in the Anycache proposal which I found very useful as it gives me structure to the work progress to whole Anycache system. I hope you found this helpful!

code/processes/setter.q
bsofjan: Hello Jonathon, same as above, I understand your comments, just to give you more context, I'm following the work development phases in the Anycache proposal which I found very useful as it gives me structure to the work progress to whole Anycache system. I hope you found this helpful!

code/processes/setter.q
bsofjan: I had to use `first` as `cacheconfiglocation` is a list with one element.

code/processes/setter.q
bsofjan: Alright, I can rename this to .anycache.setter!

code/processes/setter.q
bsofjan: The setter is a standalone process in the Anycache proposal so we can spin up many setter processes. Also the logic for the cachebuilder lives in the setter. The getter process is the lib where other processes can import to obtain the functionalities.

PR #749, opened 2026-06-10
removed lamda usage based on issue notes

benlord77:


cwallace22 (approved):
Looks good. getdetails is already defined on line 137 and clean removal of unnecessary lambda

PR #752, opened 2026-06-17
small changes recommended from feedback

benlord77:


code/common/api.q
```diff
@@ -26,7 +26,7 @@ list:{[ns;vartype]
 	api iasc (exec name from .api.detail)?api`name}
 
 // get all the namespaces in . form
```
cwallace22, line 28: Could also update the comment // get all non-empty namespaces in . form to align with the change

jonnypress (approved):
Loogs good

PR #759, opened 2026-08-18
Fix(windows): Make os.sleep work with redirected stdin

Jebiel:
Replace timeout.exe with non-interactive PowerShell Start-Sleep for Windows.

timeout.exe expects an interactive console and can fail when stdin is redirected, which can prevent RDB/IDB startup retries from working correctly and surface misleading os file-load errors.

Using PowerShell Start-Sleep removes the dependency on interactive stdin while preserving the existing sleep behavior.


jonnypress:
Looks good, thanks the the PR. The small caveat is that powershell will have more invocation overhead but I don't think we should be using .os.sleep for things which require fine grain control anyway (especially as it is supposed to be second granularity)

jonnypress (approved):
Looks good, thanks the the PR. The small caveat is that powershell will have more invocation overhead but I don't think we should be using .os.sleep for things which require fine grain control anyway (especially as it is supposed to be second granularity)

PR #766, opened 2026-09-18
Refactor savetables function to exit early for 0 rows

iodwyer:
savetables function doing a lot of unnecessary work when table rows = 0. Added a check to exit early.


iodwyer:
This change reduced our CPU usage a nice bit
image

code/wdb/writedown.q
```diff
@@ -26,26 +26,34 @@ tabsizes:([tablename:`symbol$()] rowcount:`long$(); bytes:`long$());       /- ke
...
-        /- empty the table
-        .lg.o[`delete;"deleting ",(string tabname)," data from in-memory table"];
-        @[`.;tabname;0#];
-        /- run a garbage collection (if enabled)
-        if[gc;.gc.run[]];
-        :1b;
-          ]; 0b};
+        
+        /- exit early if 0 row count, return 0b for no changes to table
+        if[0 = arows;:0b];
```
Papamaci444, line 33: What happens if a table didn't receive an update for the day? We still want that one to end up on the disk. We might could introduce another flag "eod" - this would be only called once a day and it definitely would write down all tables.
Rest could be same as your change.

iodwyer: Yes this requires more thought. I notice that (in my setup) the sort process calls .wdb.initmissingtables on the wdb only after sort + p attr etc is complete. This could take some time and therefore is too late; it needs to be called around EOD

